PR #46, opened 2019-03-11
Restructure C wrappers

TomAugspurger:
This converts ucp_py_ucp_fxns_wrapper.pyx to a pxd (Cython declaration file). This allows it to be imported from ucp_py.pyx, which fixes the Cython warning about compilation with multiple pyx sources.

I started to do the same conversion for the `buffer_helper.pyx` file, but that's a little tricker since it also includes Cython code. I've verified that, since we `include` that file, changes to `buffer_helper.pyx` still trigger  a cythonize & rebuild of the extensions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dask-worker-space
 __pytestcache__
 __pycache__
 *.egg-info/
+.pytest_cache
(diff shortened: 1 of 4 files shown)


quasiben:
closing as we now have new python binding merged in #180

PR #53, opened 2019-03-27
Use conda-build's Python

jakirkham:
Makes use of conda-build's Python via the [environment variable it sets]( https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process ). Default to the first `python3` if nothing else can be found.


Akshay-Venkatesh:
@jakirkham I'm wondering if having a `conda-install` rule would be better. This way I can use `make install` when not using conda but maybe use `make conda-install` in `build.sh` as part of the conda recipe?

Issue #64, opened 2019-01-29
Multiple listeners in the same process


TomAugspurger:
What's the recommendation for starting multiple listeners in the same process? I'm thinking of something like

```python
import asyncio
import ucp_py as ucp

ucp.init()


async def serve(ep):
    print('serving', ep)
    ucp.destroy_ep(ep)
    ucp.stop_listener()
    print('stopped', ep)


async def main():
    a = ucp.start_listener(serve, listener_port=13337, is_coroutine=True)
    ucp.get_endpoint(b'10.33.225.160', 13337)
    b = ucp.start_listener(serve, listener_port=13338, is_coroutine=True)
    ucp.get_endpoint(b'10.33.225.160', 13338)
    await a
    print('finished a')
    await b
    print('finished b')


if __name__ == '__main__':
    asyncio.run(main())
```

That outputs

```python
(python-3.7.2) an.taugspurger@dgx05:~/ucx-dev$ python tst.py
serving <ucp_py.ucp_py_ep object at 0x7f4f9513d510>
stopped <ucp_py.ucp_py_ep object at 0x7f4f9513d510>
serving <ucp_py.ucp_py_ep object at 0x7f4f9513d530>
stopped <ucp_py.ucp_py_ep object at 0x7f4f9513d530>
```

We're apparently hanging at `await a` (I'm not sure exactly where yet).

If I comment out the `b` related parts, so there's just a single listener, the output is as expected

```python
(python-3.7.2) an.taugspurger@dgx05:~/ucx-dev$ python tst.py
serving <ucp_py.ucp_py_ep object at 0x7fd5220b54d0>
stopped <ucp_py.ucp_py_ep object at 0x7fd5220b54d0>
finished a
```

PR #78, opened 2019-04-11
fixing bugs with multiple listeners case

Akshay-Venkatesh:
- used to be the case that ucp_context was associated with just one listener_context that was getting overwritten with contents of a new listener that was created
- enabling separate listener contexts per ucp context now (but limiting to 256 listeners)

### DIFF
--- a/tests/multiple-listener-client-test.py
+++ b/tests/multiple-listener-client-test.py
@@ -1,0 +1,24 @@
+import ucp_py as ucp
+import asyncio
+
+async def tmp():
+    ep1 = ucp.get_endpoint(b'192.168.40.20', 13337)
+    ep2 = ucp.get_endpoint(b'192.168.40.20', 13338)
+
+    await ep1.send_obj(b'hi')
+    print("past send1")
+    recv_ft1 = ep1.recv_future()
+    await recv_ft1
+    print("past recv1")
+    
+    await ep2.send_obj(b'hi')
+    recv_ft2 = ep2.recv_future()
+    await recv_ft2
+    print("past recv2")
+
+ucp.init()
+loop = asyncio.get_event_loop()
+coro = tmp()
+loop.run_until_complete(coro)
+loop.close()
+ucp.fin()
--- a/tests/multiple-listener-test.py
+++ b/tests/multiple-listener-test.py
@@ -1,0 +1,40 @@
+import asyncio
+import ucp_py as ucp
+
+def make_server(value):
+    assert isinstance(value, bytes)
+    assert len(value) == 1
+
+    async def serve(ep, lf):
+        print('serving', ep)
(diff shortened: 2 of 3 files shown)


Akshay-Venkatesh:
Aims to fix #64

tests/multiple-listener-client-test.py
```diff
@@ -0,0 +1,24 @@
+import ucp_py as ucp
+import asyncio
+
+async def tmp():
+    ep1 = ucp.get_endpoint(b'192.168.40.20', 13337)
```
quasiben: Maybe you could use `addr = ucp.get_address().encode('utf-8')` and substitue `addr` for the hardcoded ib addr ?

quasiben:
The test @Akshay-Venkatesh provided worked for me.  Thanks for doing this!  I made a small suggestion to the test as well.

My plan is now to use this branch with `distributed` and see if `LocalCluster` is now unblocked -- similar to what @TomAugspurger was demonstrating here:
https://github.com/Akshay-Venkatesh/ucx-py/issues/64#issuecomment-461928005

tests/multiple-listener-client-test.py
Akshay-Venkatesh: +1

quasiben:
I think we can merge whenever @Akshay-Venkatesh thinks it's done.  I have not been successful at launching distributed with `LocalCluster` but I think the issue is on the dask side now.

The workers never start:
> In [11]: cluster.workers
Out[11]:
[<Worker: ucx://10.33.225.165:13338, running, stored: 0, running: 0/1, ready: 0, comm: 0, waiting: 0>,
 <Worker: ucx://10.33.225.165:13339, running, stored: 0, running: 0/1, ready: 0, comm: 0, waiting: 0>]

And when destroying the cluster there are some errors:
> In [13]: cluster.close()
/home/nfs/bzaitlen/GitRepos/distributed/distributed/core.py:442: RuntimeWarning: coroutine 'UCX.close' was never awaited
  comm.close()
distributed.core - ERROR - 'NoneType' object has no attribute 'recv_future'
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/core.py", line 401, in handle_stream
    msgs = yield comm.read()
  File "/home/nfs/bzaitlen/miniconda3/envs/ucx-dev/lib/python3.7/site-packages/tornado/gen.py", line 729, in run
    value = future.result()
  File "/home/nfs/bzaitlen/GitRepos/distributed/distributed/comm/ucx.py", line 166, in read
AttributeError: 'NoneType' object has no attribute 'recv_future'

I'll look into this over the weekend and early next week

Akshay-Venkatesh:
Thanks for the update, Ben.

PR #84, opened 2019-04-22
 Change library name from `ucp_py` to `ucp`

quasiben:
fixes https://github.com/Akshay-Venkatesh/ucx-py/issues/82

cc @jakirkham note we are changing the import name for `ucxpy` -- if you are testing the import name in a conda-recipe that will need to change


quasiben:
@Akshay-Venkatesh note that I have left most the cython backend untouched.  I was going down this path but changed course after conflicts between ucp_init() in ucxp-py and ucx proper

Akshay-Venkatesh:
@quasiben Noted. Thanks for the update

PR #85, opened 2019-04-22
Change library name from `ucp_py` to `ucp`

### LINKED ISSUE #82: Drop `_py` from the library name?
It seems like it's common to import this library as ...

```python
import ucp_py as ucp
```

I wonder if, instead, we might just structure the Python library so that it is named `ucp` from within Python.  This repository would continue to be named `ucp-py`, but we would import it just as `ucp`.  This might be a bit more idiomatic from the Python side.

quasiben:
fixes #82

cc @jakirkham note we are changing the import name for ucxpy -- if you are testing the import name in a conda-recipe that will need to change

@Akshay-Venkatesh note that I have left most the cython backend untouched. I was going down this path but changed course after conflicts between ucp_init() in ucxp-py and ucx proper


mrocklin:
It looks like this is proposed and approved.  How do we typically merge things?  Is it ok to merge this?

Akshay-Venkatesh:
Back when it was just Tom and me working on this, we reviewed PRs and merge was done by the committer. I'm good with continuing with that model unless you see issues.

quasiben:
👍

Issue #80, opened 2019-04-17
How should libraries clean up?


mrocklin:
As a library user of this library, how should I clean things up?  Am I responsible for calling the global `ucp.fin` function?  What if some other library is also running that wants to use UCX?

Akshay-Venkatesh:
At the moment, cleanup is explicit. This is kind of inline with the MPI world. The main program is responsible for initiating `MPI_Init/ucp.init()`; other libraries may be used in the main program and MPI/UCX handles are passed to it; libraries wouldn't clear/initiate resources; finally, on completion of library routines the main program cleans up through `MPI_Finalize/ucp.fin()` 

Is this contrary to how python applications generally work?

mrocklin:
Well its tricky because we're making a Python library rather than a Python application.  I guess we could require users to call `ucp.init()` and `ucp.fin()` on their own, but honestly I wouldn't mind hiding this from them.  Most of them won't know what UCX is.

What are the costs of calling `ucp.init` when we import `ucp`?  Or alternatively can we make `ucp.init` idempotent so that Dask can do this (perhaps Dask can be more opinionated here than UCX is).

What happens if a user doesn't finalize `ucp` before they exit the process?  In particular, I wouldn't expect interactive users to be very diligent about doing this.

Akshay-Venkatesh:
> Well its tricky because we're making a Python library rather than a Python application. I guess we could require users to call `ucp.init()` and `ucp.fin()` on their own, but honestly I wouldn't mind hiding this from them. Most of them won't know what UCX is.
> 

I suppose we could go about calling `ucp.init()` when `ucp` module is imported. What would be a good time to cleanup resources? Is it safe to detect `ucp` module being unloaded and have `ucp.fin()` be called there?

> What are the costs of calling `ucp.init` when we import `ucp`? Or alternatively can we make `ucp.init` idempotent so that Dask can do this (perhaps Dask can be more opinionated here than UCX is).
> 

`ucp.init()` is a fairly light-weight operation. 

IIUC, we could also make `ucp.init` idempotent and maintain a _refcount_ so that we all resources are cleaned up when a series of `ucp.fin` calls have the desired effect when _refcount_ reaches 0. Am I understanding this right?

> What happens if a user doesn't finalize `ucp` before they exit the process? In particular, I wouldn't expect interactive users to be very diligent about doing this.

[...]

PR #96, opened 2019-04-26
Allow ucp init() and fin() to be called multiple times

Akshay-Venkatesh:
WIP to address #80

### DIFF
--- a/tests/test_init_fin.py
+++ b/tests/test_init_fin.py
@@ -1,0 +1,21 @@
+import ucp
+
+def test_multiple_init():
+    ucp.init()
+    ucp.init()
+    ucp.init()
+    ucp.fin()
+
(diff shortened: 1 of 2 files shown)


mrocklin:
Checking in.  What is the status on this?

Akshay-Venkatesh:
@mrocklin The changes here can be pushed to devel but this doesn't address the convenience of using `ucp` module without calling `ucp.init()`. Any thoughts on how can fill that gap?

mrocklin:
Well, this makes it safe for Dask to call `ucp.init()` at import time.  So for our users at least we don't care much.

Akshay-Venkatesh:
Shall I merge this into devel then?

mrocklin:
I'm in favor

PR #104, opened 2019-05-10
log basic resource init/fin and transfer functions

Akshay-Venkatesh:
- basic logging for now
- may extend this to capture timing information

### DIFF
--- a/ucp/_libs/ucp_py.pyx
+++ b/ucp/_libs/ucp_py.pyx
@@ -5,6 +5,9 @@ import concurrent.futures
 import asyncio
 import time
 import sys
+import logging
+import os
+import socket
 from weakref import WeakValueDictionary
 
 cdef extern from "src/ucp_py_ucp_fxns.h":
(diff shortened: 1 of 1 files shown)


ucp/_libs/ucp_py.pyx
```diff
@@ -38,6 +41,20 @@ PENDING_MESSAGES = {}  # type: Dict[Message, Future]
...
+
+def ucp_logger(fxn):
+    def wrapper(*args, **kwargs):
+        global LOGGER
+        args_repr = [repr(a) for a in args]
+        kwargs_repr = [f"{k}={v!r}" for k, v in kwargs.items()]
+        signature = ", ".join(args_repr + kwargs_repr)
+        LOGGER.debug(f"Calling {fxn.__name__}({signature})")
+        rval = fxn(*args, **kwargs)
+        LOGGER.debug(f"{fxn.__name__!r} returned {rval!r}")
```
mrocklin: FYI, this isn't free.  In Dask we ended up removing debug calls from within our inner loops because they took up non-trivial time.  This shouldn't matter unless you're above the 10k msgs/s rate (a guess), but I thought I'd mention it in case you care about that level.

Akshay-Venkatesh: @mrocklin I've made the logger a no-op if the `UCXPY_LOG_LEVEL` env var isn't set.

PR #105, opened 2019-05-16
Retry connection establishment

Akshay-Venkatesh:
- if client appears before server which listens, this PR allows the client to wait for `UCXPY_CONN_RETRY_PERIOD` seconds, and retry connection establishment for a maximum of `EP_CONN_MAX_RETRIES` attempts.
```
# client
akvenkatesh@prm-dgx-10:~/ucx-py$ UCXPY_CONN_RETRY_PERIOD=2 UCX_MEMTYPE_CACHE=n UCX_RNDV_SCHEME=put_zcopy python3 tests/send-recv-py-obj.py -o cupy -s IP -p 13337
[1558033604.520057] [prm-dgx-10:81187:1]   rdmacm_iface.c:401  UCX  ERROR received event RDMA_CM_EVENT_UNREACHABLE. status = 1. Peer: IP:13337.
connection failed. Retrying in 2 seconds
[1558033606.523174] [prm-dgx-10:81187:1]   rdmacm_iface.c:401  UCX  ERROR received event RDMA_CM_EVENT_UNREACHABLE. status = 1. Peer: IP:13337.
connection failed. Retrying in 2 seconds
[1558033608.526044] [prm-dgx-10:81187:0]   rdmacm_iface.c:401  UCX  ERROR received event RDMA_CM_EVENT_UNREACHABLE. status = 1. Peer: IP:13337.
connection failed. Retrying in 2 seconds
about to reply
talk_to_server done

# server which arrives later
akvenkatesh@prm-dgx-10:~/ucx-py$ UCX_MEMTYPE_CACHE=n UCX_RNDV_SCHEME=put_zcopy python3 tests/send-recv-py-obj.py -o cupy
about to send
about to recv
server sent:  array([48, 48..., dtype=uint8) <class 'cupy.core.core.ndarray'>
server recv:  array([48, 48..., dtype=uint8) <class 'cupy.core.core.ndarray'>
talk_to_client done
```

### DIFF
--- a/ucp/_libs/src/ucp_py_ucp_fxns.c
+++ b/ucp/_libs/src/ucp_py_ucp_fxns.c
@@ -38,6 +38,10 @@
 #define MAX_LISTEN_RETRIES 256
 #define MAX_LISTENERS 256
 
+#define EP_CONN_DEFAULT_RETRY_PERIOD 10
+#define EP_CONN_INIT   0
+#define EP_CONN_FAILED 1
+
 TAILQ_HEAD(tailhead, entry) cb_free_head, cb_used_head;
 struct entry {
     void *py_cb;                    /* pointer to python callback */
@@ -70,6 +74,7 @@ typedef struct ucp_py_ctx {
     int epoll_fd_local;
     int epoll_fd;
     int num_listeners;
+    int connection_retry_period;
     struct epoll_event ev;
 } ucp_py_ctx_t;
 
@@ -568,6 +573,13 @@ static int start_listener(ucp_worker_h ucp_worker, ucx_listener_ctx_t *context,
     return status;
 }
 
+static void ep_err_handler_cb(void *arg, ucp_ep_h ep, ucs_status_t status)
+{
+    int *connection_status = arg;
+    DEBUG_PRINT(stderr, "Inside ep_err_handler_cb %d\n", *connection_status);
+    *connection_status = EP_CONN_FAILED;
+}
+
 void *ucp_py_get_ep(char *ip, int listener_port)
 {
     ucp_ep_params_t ep_params;
@@ -577,18 +589,26 @@ void *ucp_py_get_ep(char *ip, int listener_port)
     ucp_py_internal_ep_t *internal_ep;
     struct ucx_context *request = 0;
     char tmp_str[TAG_STR_MAX_LEN];
+    ucp_err_handler_t ep_err_hanlder;
+    int connection_status = EP_CONN_INIT;
 
+    ep_err_hanlder.cb = ep_err_handler_cb;
+    ep_err_hanlder.arg = &connection_status;
+    
     internal_ep = (ucp_py_internal_ep_t *) malloc(sizeof(ucp_py_internal_ep_t));
     ep_ptr = (ucp_ep_h *) malloc(sizeof(ucp_ep_h));
     set_connect_addr(ip, &connect_addr, (uint16_t) listener_port);
     ep_params.field_mask       = UCP_EP_PARAM_FIELD_FLAGS     |
                                  UCP_EP_PARAM_FIELD_SOCK_ADDR |
+                                 UCP_EP_PARAM_FIELD_ERR_HANDLER |
                                  UCP_EP_PARAM_FIELD_ERR_HANDLING_MODE;
     ep_params.err_mode         = UCP_ERR_HANDLING_MODE_PEER;
     ep_params.flags            = UCP_EP_PARAMS_FLAGS_CLIENT_SERVER;
+    ep_params.err_handler      = ep_err_hanlder;
     ep_params.sockaddr.addr    = (struct sockaddr*)&connect_addr;
     ep_params.sockaddr.addrlen = sizeof(connect_addr);
 
+ retry:
     status = ucp_ep_create(ucp_py_ctx_head->ucp_worker, &ep_params, ep_ptr);
     if (status != UCS_OK) {
         ERROR_PRINT("failed to connect to %s (%s)\n", ip,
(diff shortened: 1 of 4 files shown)


Akshay-Venkatesh:
@mrocklin or @rjzamora I'd appreciate a review when you get a chance.

ucp/_libs/src/ucp_py_ucp_fxns.c
```diff
@@ -608,11 +631,43 @@ void *ucp_py_get_ep(char *ip, int listener_port)
...
+            ucp_request_cancel(ucp_py_ctx_head->ucp_worker, &request);
+            request_init(request);
+            ucp_request_free(request);
+            printf("connection failed. Retrying in %d seconds\n",
+                        ucp_py_ctx_head->connection_retry_period);
+            connection_status = EP_CONN_INIT;
+            sleep(ucp_py_ctx_head->connection_retry_period);
+            if (num_retries < EP_CONN_MAX_RETRIES) {
+                num_retries++;
+                goto retry;
```
mrocklin: Whoa, I'm super-curious why you've chosen to use `goto`.  I thought that that went out of style :)

ucp/_libs/src/ucp_py_ucp_fxns.c
```diff
@@ -608,11 +631,43 @@ void *ucp_py_get_ep(char *ip, int listener_port)
...
     if (UCS_PTR_IS_ERR(request)) {
         fprintf(stderr, "unable to send UCX data message\n");
+	if (EP_CONN_FAILED == connection_status) {
+            ucp_request_cancel(ucp_py_ctx_head->ucp_worker, &request);
+            request_init(request);
+            ucp_request_free(request);
+            printf("connection failed. Retrying in %d seconds\n",
+                        ucp_py_ctx_head->connection_retry_period);
+            connection_status = EP_CONN_INIT;
+            sleep(ucp_py_ctx_head->connection_retry_period);
```
mrocklin: This sleep is problematic from the perspective of an asynchronous framework (like Dask).  It's disruptive to block the main thread for more than several milliseconds.  Ideally making a connection would be asynchronous so that we could run other things in the background.

ucp/_libs/src/ucp_py_ucp_fxns.c
```diff
@@ -608,11 +631,43 @@ void *ucp_py_get_ep(char *ip, int listener_port)
...
+            if (num_retries < EP_CONN_MAX_RETRIES) {
```
mrocklin: It's common to provide a timeout in seconds.  For example I might expect something like the following:

```python
connection = await connect('127.0.0.1:8000', timeout=10)
```

ucp/_libs/src/ucp_py_ucp_fxns.c
Akshay-Venkatesh: For one, UCX makes use of it a fair bit:
```shelll
$ grep -r "goto " $HOME/ucx-github/src/ | wc -l
$ 854
```
But they're almost exclusively used for error cases unlike here. I think I'll refactor the code to use this better. Code changes coming up soon.

ucp/_libs/src/ucp_py_ucp_fxns.c
Akshay-Venkatesh: Fair point. Would it be better if I returned `unreachable endpoint` error and then the coroutine which uses `get_endpoint` function can call say `asyncio.sleep` and retry later? Essentially leaving retry duration and policy to dask or some equivalent.

ucp/_libs/src/ucp_py_ucp_fxns.c
Akshay-Venkatesh: This is probably cleaner the env variable I'm using now.

Akshay-Venkatesh: Do we want to include `max_retries` as well?

ucp/_libs/src/ucp_py_ucp_fxns.c
mrocklin: Typically you return an unfinished thing immediately that might not have full connection information and then you have some asynchronous function to wait until connection has been achieved (which usually happens in the background. 

So you might do something like this:

```python
class Endpoint:
    def __init__(self, ...):
        ... # fast setup code only
        self._connected = asyncio.ensure_future(self.connect())

    async def connect(self, timeout=None):
        if timeout is not None:
            deadline = time() + timeout
        else:
            deadline = None
        while True:
            if try_to_connect():
                return self
            else:
                await gen.sleep(0.100)
                if time() > deadline:
                    raise TimeoutError("Failed to connect to %s after %f seconds" % (self.address, timeout))

    def __await__(self):  # this is the protocol for `await my_endpoint`
        return self._connected.__await__()  # I think this would work, but please don't trust me too much.
```

I think that with something like this we'll be able to do something like the following:

```python
ep = Endpoint(address)  # this is fast

# do other things

await ep

# now we can trust that ep is connected
```

Or, more commonly

```python
ep = await Endpoint(address)
```

ucp/_libs/src/ucp_py_ucp_fxns.c
mrocklin: In applications that I've seen it's common to specify only a timeout

ucp/_libs/src/ucp_py_ucp_fxns.c
Akshay-Venkatesh: Awesome. This helps!

Akshay-Venkatesh:
@mrocklin Made some changes approximately along the lines we discussed. As it stands now, the client will keep trying every `timeout` seconds.

mrocklin:
> As it stands now, the client will keep trying every timeout seconds

Ah, sorry, the intent of timeout is "keep trying for timeout seconds, if you still haven't connected by then, then give up and raise an exception"

```python
    async def connect(self, timeout=None):
        if timeout is not None:
            deadline = time() + timeout
        else:
            deadline = None
        while True:
            if try_to_connect():
                return self
            else:
                await gen.sleep(0.100)
                if time() > deadline:
                    raise TimeoutError("Failed to connect to %s after %f seconds" % (self.address, timeout))
```

Akshay-Venkatesh:
@mrocklin Fixed the meaning of `timeout`. Any other comments?

ucp/_libs/ucp_py.pyx
```diff
@@ -729,7 +729,18 @@ def get_endpoint(peer_ip, peer_port):
     global reader_added
 
     ep = Endpoint()
-    ep.connect(peer_ip, peer_port)
+    deadline = time.time() + timeout
```
mrocklin: `timeout=None` is a common convention for "try forever".  You might consider adding support for it.

ucp/_libs/ucp_py.pyx
```diff
@@ -729,7 +729,18 @@ def get_endpoint(peer_ip, peer_port):
...
+    connection_established = False
+    while time.time() <= deadline:
+        if -1 == ep.connect(peer_ip, peer_port):
+            await asyncio.sleep(1)
```
mrocklin: How expensive is it to try to connect?  If it's cheap then we might consider decreasing the sleep time here.

ucp/_libs/ucp_py.pyx
```diff
@@ -729,7 +729,18 @@ def get_endpoint(peer_ip, peer_port):
...
+        else:
+            connection_established = True
+            break
+
+    if not connection_established:
+        raise NameError('Timeout in connection establishment attempt')
```
mrocklin: I recommend the `TimeoutError` type instead.  People know to catch those exceptions in these cases.

ucp/_libs/ucp_py.pyx
```diff
@@ -233,8 +233,8 @@ cdef class Endpoint:
     def connect(self, ip, port):
         self.ep = ucp_py_get_ep(ip, port)
         if <void *> NULL == self.ep:
-            raise NameError('Failed to connect to remote endpoint')
-        return
+            return -1
+        return 0
```
mrocklin: Python conventions would be to return True if success and False if failure (or raise an OSError).  Not a big deal though.  I think that any Cython dev will see this for C style conventions and be fine.

mrocklin:
You might consider a test for the timeout behavior

```python
@pytest.mark.asyncio
async def test_timeout():
    with pytest.raises(TimeoutError) as info:
        start = time()
        await ucp.get_endpoint('127.0.0.1', 9999, timeout=0.1)

    assert time() < start + 2
    assert time() > start + 0.1
    assert "9999" in str(info.value)
    assert "127.0.0.1" in str(info.value)
```

Akshay-Venkatesh:
@mrocklin Addressed your comments. Thanks!

PR #110, opened 2019-05-22
Hold a reference to listener futures to avoid garbage collection

Akshay-Venkatesh:
cc @mrocklin

### DIFF
--- a/ucp/_libs/ucp_py.pyx
+++ b/ucp/_libs/ucp_py.pyx
@@ -42,7 +42,7 @@ UCX_FILE_DESCRIPTOR = -1
 reader_added = 0
 UCP_INITIALIZED = False
 LOGGER = None
-listener_futures = []
+listener_futures = set()
 
 def ucp_logger(fxn):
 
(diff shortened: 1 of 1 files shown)


mrocklin:
Long term this might not be a good solution.  It might be worth adding a TODO comment somewhere to remove the leak

Akshay-Venkatesh:
> Long term this might not be a good solution. It might be worth adding a TODO comment somewhere to remove the leak

@mrocklin Does this change not plug the leak? https://github.com/Akshay-Venkatesh/ucx-py/pull/110/commits/9d6dd8552ae8330299b24307b0ae072770cb3a5e

ucp/_libs/ucp_py.pyx
```diff
@@ -42,6 +42,7 @@ UCX_FILE_DESCRIPTOR = -1
 reader_added = 0
 UCP_INITIALIZED = False
 LOGGER = None
+listener_futures = []
```
mrocklin:
```suggestion
listener_futures = set()
```

ucp/_libs/ucp_py.pyx
```diff
@@ -665,6 +666,8 @@ def start_listener(py_func, listener_port = -1, is_coroutine = False):
         if <void *> NULL != listener.listener_ptr:
             lf.listener = listener
             lf.port = port
+            assert(lf not in listener_futures)
+            listener_futures.append(lf) # hold a reference to avoid garbage collection
```
mrocklin:
```suggestion
            listener_futures.add(lf) # hold a reference to avoid garbage collection
```

ucp/_libs/ucp_py.pyx
mrocklin: Should be fine either way, but in principle this should probably support O(1) removal times

Issue #119, opened 2019-06-07
We create poorly formed memoryview objects


mrocklin:
I occasionally get errors where we seem to be receiving zero-length messages in ucx-py.  I placed a breakpoint when this happened and investigated the object.  It looks like we're forming improper memoryview objects somehow:

```python
(Pdb) pp obj
<memory at 0x7f22051d4ac8>
(Pdb) pp len(obj)
0
(Pdb) pp bytes(obj)
(b'\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
 b'\x00\x00\xdd\x00\x00\x00\x00\x00\x00\x00')
(Pdb) pp bytes(obj[:8])
b''
```

I'll take a look briefly.

mrocklin:
@TomAugspurger if you have suggestions on where this might be happening those would be welcome.

TomAugspurger:
Is this GPU or CPU memory? For CPU, the bug is likely in
https://github.com/rapidsai/ucx-py/blob/devel/ucp/_libs/ucp_py_buffer_helper.pyx#L142,
but I'm not sure exactly where.

mrocklin:
Yeah, I'm playing around in there now.  I can get it to fail in both CPU and GPU memory.

mrocklin:
This seems to be fixed in https://github.com/rapidsai/ucx-py/pull/120

PR #120, opened 2019-06-07
Use full shape in buffer protocol

mrocklin:
Previously we would only use the first dimension in a Message shape
in the __getbuffer__ method.  This would result in poorly formed memoryview
objects.

Fixes https://github.com/rapidsai/ucx-py/issues/119

### DIFF
--- a/ucp/_libs/ucp_py_buffer_helper.pyx
+++ b/ucp/_libs/ucp_py_buffer_helper.pyx
@@ -9,6 +9,7 @@ cdef extern from "common.h":
 
 import struct
 from libc.stdint cimport uintptr_t
+import numpy as np
 
 # TODO: pxd files
 
@@ -155,13 +156,13 @@ cdef class BufferRegion:
         else:
             buffer.buf = <void *>&(self.buf.buf[0])
 
-        shape2[0] = self.shape[0]
+        shape2[0] = np.prod(self.shape)
 
         buffer.format = self.format
         buffer.internal = NULL
         buffer.itemsize = self.itemsize
-        buffer.len = self.shape[0] * self.itemsize
-        buffer.ndim = 1
+        buffer.len = np.prod(self.shape) * self.itemsize
+        buffer.ndim = len(self.shape)
         buffer.obj = self
         buffer.readonly = 0  # TODO
         buffer.shape = shape2
--- a/ucp/_libs/ucp_py.pyx
+++ b/ucp/_libs/ucp_py.pyx
@@ -73,7 +73,7 @@ def ucp_logger(fxn):
     else:
         return fxn
     LOGGER.debug('done with ucxpy init')
-    
+
 
     return wrapper
 


mrocklin:
FYI @Akshay-Venkatesh this was the cause of the zero-byte messages that I was seeing.

ucp/_libs/ucp_py_buffer_helper.pyx
```diff
@@ -155,13 +156,13 @@ cdef class BufferRegion:
...
         buffer.format = self.format
         buffer.internal = NULL
         buffer.itemsize = self.itemsize
-        buffer.len = self.shape[0] * self.itemsize
-        buffer.ndim = 1
+        buffer.len = np.prod(self.shape) * self.itemsize
+        buffer.ndim = len(self.shape)
         buffer.obj = self
         buffer.readonly = 0  # TODO
         buffer.shape = shape2
```
jakirkham: Not part of the change here, but this line is a little concerning to me. It seems to be taking a pointer to memory allocated on the stack. Meaning this pointer won't be valid when we leave this function. I wouldn't be surprised if this causes segfaults somewhere else.

One alternative would be to dynamically allocate memory for `buffer.shape` to point to and fill in the content from `shape2` and then clean up this memory in `__releasebuffer__`. Another alternative would be to make `shape2` part of `self`.

Thoughts?

Akshay-Venkatesh:
@mrocklin Your changes seem to be flattening the shape. Doesn't that result in shape knowledge being lost? 

https://github.com/rapidsai/ucx-py/blob/4030eff63438ac87d58fd009656198a184914041/ucp/_libs/ucp_py_buffer_helper.pyx#L167

I likely missing the full picture here. @TomAugspurger

TomAugspurger:
FWIW, I wouldn't place high value on the code I wrote here being correct :)

jakirkham:
@Akshay-Venkatesh, is the `self.shape` variable used elsewhere? I wonder if it couldn't be turned into the desired pointer (instead of an `object` as it is now) and used with the buffer protocol here. Maybe that even simplifies things a bit. 🙂

jakirkham:
If `self.shape` is useful (like for Python users to see), maybe we can use `self._shape` instead for the same purpose.

mrocklin:
No thoughts from me.  I honestly don't know Cython that well and don't have a sense for what is handled automatically or not.  @jakirkham if you see a good obvious solution here would you be interested in making a quick PR?  

> Doesn't that result in shape knowledge being lost?

Yes.  This is less than ideal, but not terrible.  We only really care about this thing as a byte buffer, not as a full array.  We send shape and dtype information as part of a separate header.

jakirkham:
Yes, happily. One sec...

jakirkham:
Have started this with PR ( https://github.com/rapidsai/ucx-py/pull/121 ). There is a bunch of code that is assigning to `.shape`, which would need to change as a consequence. Though need input from others to understand how that should change.

mrocklin:
If no one objects I'm going to merge this in for now.  I think that this is a strict improvement.  We can still continue with #121 afterwards.  OK with you @jakirkham ?

quasiben:
I am +1 for merging

quasiben:
Wait, running tests with this PR and things are failing

quasiben:
Nevermind -- tests are passing merging

jakirkham:
Sorry I let this one slip off my radar. In general anything that gets us closer to complying with the buffer protocol is a good thing. This is certainly a step in that direction.

PR #122, opened 2019-06-12
[REVIEW] Exposing pointers for underlying ucp_worker_h and ucp_ep_h

cjnolet:
This is a quick and dirty version for now, just to test the hand-off between endpoints and worker created in cython to the C++ layer for further use.

As per @jakirkham's suggestion, I am also investigating the use of cython Extensions to help us accomplish this hand-off.


ucp/_libs/ucp_py.pyx
```diff
@@ -44,6 +49,10 @@ UCP_INITIALIZED = False
 LOGGER = None
 listener_futures = set()
 
+
+def get_ucp_worker():
+    return <size_t>get_worker()
```
Akshay-Venkatesh: Do we need to explicitly cast this as `<size_t>` ?

ucp/_libs/ucp_py.pyx
```diff
@@ -230,6 +239,9 @@ cdef class Endpoint:
     def __cinit__(self):
         return
 
+    def get_ep(self):
+        return <size_t>self.ep
```
Akshay-Venkatesh: Same question. Do we need cast to `<size_t>` ?

ucp/_libs/ucp_py.pyx
cjnolet: For this case, since `get_worker()` returns a C++ pointer and we are returning from a pure Python function, the Cython compiler complains unless we cast it to something that can be implicitly converted to a native Python type. I am still planning on looking into John's proposed solution to this.

mrocklin:
@cjnolet @Akshay-Venkatesh , what is the status here?  Should this be merged or closed?

It looks like the last comment was from Akshay, asking for clarification on a small point.

cjnolet:
Akshay and I discussed the point offline. 

This has been ready to merge. I'll resolve conflicts today.

mrocklin:
Thanks @cjnolet .  As an FYI, I'm not sure that any single person actively maintains this respository today, so for future changes you will probably need to be vocal in getting someone to merge things.

cjnolet:
This is good to go. @Akshay-Venkatesh, are you able to merge this for release?

ucp/_libs/ucp_py.pyx
```diff
@@ -17,12 +18,18 @@ cdef extern from "ucp/api/ucp.h":
...
+
 cdef extern from "src/ucp_py_ucp_fxns.h":
     cdef struct ucx_context:
         int completed
     cdef struct data_buf:
         void* buf
 
+    cdef void *get_worker()
+    cdef void *get_ep_ptr(void *ep)
+
```
Akshay-Venkatesh: @cjnolet Just one request -- can you move the above cython declarations to `ucp/_libs/ucp_py_ucp_fxns_wrapper.pyx` where all other such declarations exist?

mrocklin: Poke @cjnolet it looks like there is a small request here.  Otherwise I think that we are good to go.

cjnolet: I made the change last night but wanted to verify it end-to-end w/ cuML. It's been verified and changes pushed. Thanks, guys!

mrocklin:
Thanks @cjnolet 

@Akshay-Venkatesh , I don't understand most of this PR.  If it's ok I'd like to defer to you on merging this one.

PR #128, opened 2019-06-18
augment test_send_recv_obj with large message transfers

Akshay-Venkatesh:
Until now, `test_send_recv_obj` has only checked 2-byte transfers which stresses just the `eager` protocol. This PR adds 4MB transfers to stress `rendezvous` protocol.
cc @jakirkham

### DIFF
--- a/tests/test_send_recv_obj.py
+++ b/tests/test_send_recv_obj.py
@@ -18,8 +18,8 @@ async def echo_pair(cuda_info=None):
     try:
         yield listener, client
     finally:
-        ucp.stop_listener(listener)
         ucp.destroy_ep(client)
+        await t
         ucp.fin()
 
 
--- a/tests/test_send_recv_obj.py
+++ b/tests/test_send_recv_obj.py
@@ -4,7 +4,7 @@ import pytest
 from contextlib import asynccontextmanager
 
 import ucp
-
+msg_sizes = [2 ** i for i in range(0, 25, 4)]
 
 @asynccontextmanager
 async def echo_pair(cuda_info=None):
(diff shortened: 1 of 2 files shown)


tests/test_send_recv_obj.py
```diff
@@ -35,6 +36,21 @@
     assert result.tobytes() == msg
 
 
+@pytest.mark.asyncio
+async def test_send_recv_large_bytes():
```
mrocklin:
```suggestion
@pytest.mark.parametrize('convert', [lambda x: x, memoryview])
async def test_send_recv_large_bytes(convert):
```

tests/test_send_recv_obj.py
```diff
@@ -35,6 +36,21 @@
...
+    async with echo_pair() as (_, client):
+        x = "a"
+        x = x * 4194304
+        msg = bytes(x, encoding='utf-8')
```
mrocklin:
```suggestion
        msg = bytes(x, encoding='utf-8')
        msg = convert(msg)
```

jakirkham:
Always in favor of more tests. 🙂

Are there particular things you are looking for feedback from me on Akshay?

Akshay-Venkatesh:
Nothing specific. I'm trying to run it by someone with more python experience that's all.

tests/test_send_recv_obj.py
```diff
@@ -49,12 +56,17 @@
...
+@pytest.mark.parametrize("size", msg_sizes)
+async def test_send_recv_numpy(size):
     np = pytest.importorskip('numpy')
-    async with echo_pair() as (_, client):
-        msg = np.frombuffer(memoryview(b"hi"), dtype='u1')
+    x = "a"
+    x = x * size
+    msg = bytes(x, encoding='utf-8')
+    msg = memoryview(msg)
+    msg = np.frombuffer(msg, dtype='u1')
```
jakirkham: This should work fine. Would also work if this were the following.

```python
msg = np.asarray(msg)
```

The benefit of this expression is it is able to infer the type from the `memoryview`. Also if the data has a shape to it (like it is a 2-D matrix), then that will be preserved. `np.frombuffer` flattens everything to a 1-D array instead.

jakirkham: There's a few other lines below that could benefit from this same pattern. Not critical though.

jakirkham:
Sure. Added a small comment above. It's not critical, but it could be used to tidy things a little bit. Sorry for getting to this somewhat late.

PR #139, opened 2019-08-08
fea/add-ucx-listener-test

quasiben:
PR adds tests with message passing scheme similar to dask.  Tests use both large and small cudf and cupy objects

### DIFF
--- a/tests/test_custom_ep_send_recv.py
+++ b/tests/test_custom_ep_send_recv.py
@@ -1,0 +1,229 @@
+import asyncio
+import pytest
+
+import ucp
+import time
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "g",
+    [
+        lambda cudf: cudf.Series([1, 2, 3]),
+        lambda cudf: cudf.Series([1, 2, 3], index=[4, 5, 6]),
+        lambda cudf: cudf.Series([1, None, 3]),
+        lambda cudf: cudf.Series(range(2**13)),
+        lambda cudf: cudf.DataFrame({'a': range(2**20)}),
+        lambda cudf: cudf.DataFrame({'a': range(2**26)}),
+    ]
+)
+async def test_send_recv_cudf(event_loop, g):
+    # requires numba=0.45 (.nbytes)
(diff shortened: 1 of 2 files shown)


quasiben:
@mrocklin what do you think about the queueing system used in this test ?  My plan is to do something similar in dask

mrocklin:
> what do you think about the queueing system used in this test ?

Why is it necessary?  In theory, it's somewhat unclean to create an unchecked buffer like this.  Ideally we only pull things off of UCX when we're ready for them so that UCX (or whatever transport it uses) feels some pressure if one end starts sending things too quickly.  I don't know that this will affect us much in the common case, but it's sort of general practice not to do these things if you can avoid it.

So, why are we doing this?  What are we working around?

quasiben:
The issue in UCX seems to be that when await write/read with large messages the [Rendezvous-Protocol](https://github.com/openucx/ucx/wiki/Rendezvous-Protocol) is invoked and this can create a deadlock.   One solution to this problem is to move the read inside (make current read a `_read` function) the coroutine `serve_forever` and post the reads to a queue to be consumed by a public facing `read` function

quasiben:
Another solution to the problem is not calling `await write` but I don't fully understand why this works.  @Akshay-Venkatesh can you explain again (apologies) 

I didn't explore this very much as much of distributed relies on `await write/await read`

mrocklin:
> The issue in UCX seems to be that when await write/read with large messages the Rendezvous-Protocol is invoked and this can create a deadlock.

After reading through these comments and that document I still don't understand why UCX would deadlock here.

Akshay-Venkatesh:
> > The issue in UCX seems to be that when await write/read with large messages the Rendezvous-Protocol is invoked and this can create a deadlock.
> 
> After reading through these comments and that document I still don't understand why UCX would deadlock here.

https://github.com/rapidsai/ucx-py/issues/140#issuecomment-519596036

mrocklin:
Is it ok to close this?

quasiben:
No, i would like to keep this as a valid test -- I'll finish it up now

quasiben:
merging now

PR #144, opened 2019-08-15
[MERGE] Fixes wrong shape and stack allocation in BufferRegion

madsbk:
`len(memoryview.shape)` might not equal `memoryview.ndim`

All tests in `tests/test_buffer_region.py ` now passes on my installation

**Update:** the returned shape and strides in `__buffer__` is now allocated on the heap.

### DIFF
--- a/tests/test_buffer_region.py
+++ b/tests/test_buffer_region.py
@@ -8,7 +8,7 @@ def test_set_read():
     buffer_region = ucp.BufferRegion()
     buffer_region.populate_ptr(obj)
     res = memoryview(buffer_region)
-    assert res == obj
+    assert bytes(res) == bytes(obj)
     assert res.tobytes() == obj.tobytes()
 
     # our properties
--- a/ucp/_libs/ucp_py_buffer_helper.pyx
+++ b/ucp/_libs/ucp_py_buffer_helper.pyx
@@ -192,7 +192,8 @@ cdef class BufferRegion:
         self._populate_ptr(obj)
 
     cpdef _populate_ptr(self, format_[:] pyobj):
-        self.shape = pyobj.shape
+        # Notice, `len(memoryview.shape)` might not equal `memoryview.ndim`
+        self.shape = [pyobj.shape[i] for i in range(pyobj.ndim)]
         self._is_cuda  = 0
         # TODO: We may not have a `.format` here. Not sure how to handle.
         if hasattr(pyobj.base, 'format'):
--- a/ucp/_libs/ucp_py_buffer_helper.pyx
+++ b/ucp/_libs/ucp_py_buffer_helper.pyx
@@ -226,18 +226,18 @@ cdef class BufferRegion:
     # ------------------------------------------------------------------------
     # Manual memory management
 
-    def alloc_host(self, Py_ssize_t len):
-        self.buf = allocate_host_buffer(len)
+    def alloc_host(self, Py_ssize_t length):
+        self.buf = allocate_host_buffer(length)
         self._is_cuda = 0
         self._mem_allocated = 1
-        self.shape[0] = len
+        self.shape = (length,)
 
-    def alloc_cuda(self, len):
+    def alloc_cuda(self, length):
         cuda_check()
-        self.buf = allocate_cuda_buffer(len)
+        self.buf = allocate_cuda_buffer(length)
         self._is_cuda = 1
         self._mem_allocated = 1
-        self.shape[0] = len
+        self.shape = (length,)
 
     def free_host(self):
         free_host_buffer(self.buf)
--- a/ucp/_libs/ucp_py_buffer_helper.pyx
+++ b/ucp/_libs/ucp_py_buffer_helper.pyx
@@ -9,6 +9,7 @@ cdef extern from "common.h":
 
 import struct
 from libc.stdint cimport uintptr_t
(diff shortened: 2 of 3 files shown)


jakirkham:
> `len(memoryview.shape)` might not equal `memoryview.ndim`

Very interesting! When is that not true?

madsbk:
> > `len(memoryview.shape)` might not equal `memoryview.ndim`
> 
> Very interesting! When is that not true?

On my installation :P

tests/test_buffer_region.py
```diff
@@ -8,7 +8,7 @@ def test_set_read():
     buffer_region = ucp.BufferRegion()
     buffer_region.populate_ptr(obj)
     res = memoryview(buffer_region)
-    assert res == obj
+    assert bytes(res) == bytes(obj)
```
quasiben: 👍

quasiben:
@madsbk is it possible to add something which can explicitly test this ?  Or are the tests in buffer_region.py enough ?

jakirkham:
> > > `len(memoryview.shape)` might not equal `memoryview.ndim`
> > 
> > 
> > Very interesting! When is that not true?
> 
> On my installation :P
> I don't know why, but `pyobj.shape[i]` was `[2,0,0,0,0,0,0]` in one of the test in `tests/test_buffer_region.py`

Ah right! Yeah sometimes C code will allocating a buffer larger than the needed shape. Good catch. Sorry forgot about this.

quasiben:
These tests pass for me -- @Akshay-Venkatesh should we merge ?

ucp/_libs/ucp_py_buffer_helper.pyx
```diff
@@ -192,7 +192,8 @@ cdef class BufferRegion:
         self._populate_ptr(obj)
 
     cpdef _populate_ptr(self, format_[:] pyobj):
-        self.shape = pyobj.shape
+        # Notice, `len(memoryview.shape)` might not equal `memoryview.ndim`
+        self.shape = [pyobj.shape[i] for i in range(pyobj.ndim)]
```
kkraus14: I believe this is expected to be a tuple elsewhere as opposed to a list

Akshay-Venkatesh:
Also, can someone ELI5 what was the problem exactly and what this PR is doing?

quasiben:
@Akshay-Venkatesh I can try but I am not as familiar with buffer protocols as @jakirkham / @madsbk @kkraus14 .  On my machine (and @madsbk as well) `pyobj`, in populate_ptr looks like the following:

> [2, 0, 0, 0, 0, 0, 0, 0]

@jakirkham suggested that, " C code will allocate a buffer larger than the needed shape" . The only reference I could find about this is in the [numpy docs](https://www.numpy.org/devdocs/reference/arrays.interface.html) (look at the shape definition).  The [cython docs](https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#accepting-strings-from-python-code) recommend simply handling the 1st dimension of shape:

> self.shape = pyobj.shape[0]

For reference in a python session we see the following:

```
In [1]: memoryview(b'hi').shape                                                                                                                                                       
Out[1]: (2,)

In [2]: memoryview(b'hi').ndim                                                                                                                                                        
Out[2]: 1
```

I confirmed the large buffer size by putting a print statement in the `cpdef _populate_ptr` definition.

ucp/_libs/ucp_py_buffer_helper.pyx
Akshay-Venkatesh: I just noticed the tuple usage as well:
```python3

In [3]: np_arr = np.array([[1, 2], [3, 4]])                                                                            

In [4]: v = memoryview(np_arr)                                                                                         

In [5]: v.shape                                                                                                        
Out[5]: (2, 2)

In [6]: v2 = memoryview(b'abcdefg')                                                                                    

In [7]: v2.shape                                                                                                       
Out[7]: (7,)
```
In light of this, we should change the initialization of BufferRegion shape [here](https://github.com/rapidsai/ucx-py/blob/devel/ucp/_libs/ucp_py_buffer_helper.pyx#L114) from:
```
        self.shape = [0]
```
to
```
        self.shape = (0, )
```

quasiben:
If there are no objects can we/I merge this tomorrow ?

madsbk:
The shape is now a tuple instead of a list.
@quasiben, I think it is ready for merging now.

quasiben:
Apologies for hastily rushing this along.  I tested this PR with `test_custom_send_recv` and got some errors:

```
>   self.shape[0] = len
E   TypeError: 'tuple' object does not support item assignment

ucp/_libs/ucp_py_buffer_helper.pyx:233: TypeError
```

In various functions inside the buffer helper, ucx-py is assigning shape and this fails if shape is a tuple. 
 For example: https://github.com/rapidsai/ucx-py/blob/c70ccd0ca19363374b813e9926da72a3bba6f7b2/ucp/_libs/ucp_py_buffer_helper.pyx#L229-L240

@madsbk @Akshay-Venkatesh what are you're thoughts here ?  Should this be fixed in a follow on PR or continued to be worked in this PR ?

madsbk:
@quasiben, It should be fixed now; can you confirm?

quasiben:
Thank you @madsbk !  The tests for buffer and send_recv are passing now.  I'm going to merge and continue cleaning up tests

PR #145, opened 2019-08-15
[WIP] Two GPUs + CUDF

quasiben:
cc @mrocklin @Akshay-Venkatesh 

This PR adds a test for starting "workers" attached to separate GPUs.  This is intended to mimic a failing/halting case which we regularly see when using a similar message passing scheme with dask.

Tests currently pass for the following cases:

- cdf = cudf.DataFrame({"a": range(5), "b": range(
- cdf = cudf.DataFrame({"a": [1.0], "b": [1.0]}).head(0)
 
This is somewhat disappointing -- I was hoping at least one of the above would fail and help narrow down the issue

### DIFF
--- a/tests/test_send_recv_two_workers.py
+++ b/tests/test_send_recv_two_workers.py
@@ -1,0 +1,175 @@
+import os
+import asyncio
+import pytest
+import pickle
+import time
+
+import ucp
+import struct
+import multiprocessing
+from distributed.utils import nbytes
+
+
+async def get_ep(name, port):
+    addr = ucp.get_address().encode()
+    ep = await ucp.get_endpoint(addr, port)
+    print(name, ep)
+    return ep
+
+
+def client(env, port):
+    # wait for server to come up
+    # receive cudf object
+    # deserialize
+    # assert deserialized msg is cdf
+    # send receipt
+
+    os.environ.update(env)
+    ucp.init()
+
+    async def read():
+        await asyncio.sleep(3)
+        ep = await get_ep("client", port)
+        print("Waiting to receiving frames...")
+
+        resp = await ep.recv_future()
+        obj = ucp.get_obj_from_msg(resp)
+        nframes, = struct.unpack("Q", obj[:8])  # first eight bytes for number of frames
+        gpu_frame_msg = obj[
+            8 : 8 + nframes
+        ]  # next nframes bytes for if they're GPU frames
+        is_gpus = struct.unpack("{}?".format(nframes), gpu_frame_msg)
+
+        sized_frame_msg = obj[8 + nframes :]  # then the rest for frame sizes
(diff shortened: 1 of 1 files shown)


tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,175 @@
...
+            frames.append(frame)
+
+        header = pickle.loads(frames[0])
+        frames = frames[1:]  #
+        typ = pickle.loads(header["type"])
+        cuda_obj = typ.deserialize(header, frames)
+
+        print(cuda_obj, len(cuda_obj))
+        asyncio.sleep(5)
+        # how should we confirm data is on two GPUs ?
```
quasiben: @kkraus14 / @pentschev  at this point in the test we have sent data from GPU1 to GPU5.  Do you know of an easy way to confirm data on both GPUs ?

quasiben:
One nice thing about this style of test is that it is self contained -- no env vars need to be defined:

> py.test -s -v tests/test_send_recv_two_workers.py

tests/test_send_recv_two_workers.py
mrocklin: You might want to refer to this pynvml code that we use within distributed to get the memory use and utilization of each GPU: https://github.com/dask/distributed/blob/master/distributed/diagnostics/nvml.py

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,183 @@
...
+
+            data = cupy.arange(10)
+            frames = await distributed.comm.utils.to_frames(
+                data, serializers=("cuda", "pickle")
+            )
+        if cuda_type == "cudf":
+            import cudf
+            import numpy as np
+            asyncio.sleep(0.01)
+            # cdf = cudf.DataFrame({"a": range(5), "b": range(5)})
```
quasiben: The cudf DFs should be parameterized as part of the test

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,183 @@
...
+        print("Last Bit of Cleanup...")
+        ucp.fin()
+
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(f(port))
+
+
+def test_send_recv_cudf(event_loop):
+    base_env = {
+        "UCX_RNDV_SCHEME": "put_zcopy",
```
quasiben: We should parameterize the CUCDA_VISIBLE DEVICE for different configurations and we should also test when workers are attached to the same ib interface and difference interfaces: (ib0, ib1)

tests/test_send_recv_two_workers.py
kkraus14: You can also use `gdf['a'].data.mem.gpu_data.context.PCI_BUS_ID` which will return the unique PCI bus identifier of the GPU (will not be 0 - 7 on a DGX for example)

Issue #146, opened 2019-08-16
List requirements in setup.py


jakirkham:
Would be great to have the requirements [listed in the `setup.py`]( https://github.com/rapidsai/ucx-py/blob/c58b2ac56814264091bda89129381214a8be4eb9/setup.py#L69-L76 ). This would help when doing packaging so we are aware what should be listed as requirements and whether those have changed.

PR #147, opened 2019-08-16
Drop hard numpy requirement

jakirkham:
This reworks PR ( https://github.com/rapidsai/ucx-py/pull/120 )'s changes slightly to drop the hard NumPy requirement added by it. Fixing an issue encountered when building the ucx-py packages.

### DIFF
--- a/ucp/_libs/ucp_py_buffer_helper.pyx
+++ b/ucp/_libs/ucp_py_buffer_helper.pyx
@@ -9,7 +9,6 @@ cdef extern from "common.h":
 
 import struct
 from libc.stdint cimport uintptr_t
-import numpy as np
 
 # TODO: pxd files
 
@@ -158,7 +157,9 @@ cdef class BufferRegion:
         else:
             buffer.buf = <void *>&(self.buf.buf[0])
 
-        shape2[0] = np.prod(self.shape)
+        shape2[0] = self.shape[0]
+        for s in self.shape:
+            shape2[0] *= s
 
         buffer.format = self.format
         buffer.internal = NULL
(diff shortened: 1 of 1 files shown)


jakirkham:
Does this change still work for your use case, @mrocklin?

mrocklin:
It seems to be equivalent, but I no longer have a solid test.  I say merge at will.

jakirkham:
Thanks Matt! 😄

pentschev:
Hmmm, I'm not sure exactly why/if ucx-py really requires NumPy, but using the new ucx-py conda package from @jakirkham fails to load `ucp` because of missing NumPy:

```python
>>> import ucp
  File "<stdin>", line 1, in <module>
  File "/home/nfs/pentschev/src/ucx-py/ucp/__init__.py", line 2, in <module>
    from ._libs.ucp_py import *  # noqa
  File "ucp/_libs/ucp_py_buffer_helper.pyx", line 12, in init ucp_py
    import numpy as np
ModuleNotFoundError: No module named 'numpy'
```

TomAugspurger:
Just checking: did you rebuild your extensions?

jakirkham:
@pentschev, that seems to be referring to the `import` this patch removed. The packages also contain this change. Are you sure this isn’t coming from a local install from previous experimentation?

In any event, I agree we could use some clarity on ucx-py’s requirements. Raised issue ( https://github.com/rapidsai/ucx-py/issues/146 ) on this point. Hopefully others can fill in any details we are missing.

jakirkham:
Yeah I should regenerate the Cythonized C files here too. Though the packages used the Cython code directly. So I think we still shouldn’t be seeing this error.

pentschev:
Indeed, there was something strange with my conda install lately, after installing new packages to the environment they're not always reflected until I open a new shell session. I cleaned it up and now it works. Thanks for the help!

jakirkham:
No worries. Glad to hear things are working correctly now 🙂

PR #149, opened 2019-08-16
Generalize CUDA version in install instructions

jakirkham:


jakirkham:
cc @mrocklin

README.md
```diff
@@ -2,13 +2,14 @@
 
 ## Installing preliminary Conda packages.
 
-Some preliminary Conda packages can be installed as so.
+Some preliminary Conda packages can be installed as so. Replace `<CUDA
+version>` with either `9.2` or `10.0`.
 
 ```
-conda create -n ucx -c conda-forge -c jakirkham/label/ucx cudatoolkit=9.2 ucx-proc=*=gpu ucx ucx-py python=3.7
+conda create -n ucx -c conda-forge -c jakirkham/label/ucx cudatoolkit=<CUDA version> ucx-proc=*=gpu ucx ucx-py python=3.7
```
mrocklin: I think that conda 4.7 now auto-identifies the `cuda` virtual package.  If the `ucx` package depends on `cudatoolkit` then perhaps  we could remove that bit from the conda create line entirely, and ask instead that users ensure that they're using conda 4.7 or greater?

Of course, that might be even more complex than what's here :)

Anyway, +1 from me.  I'm glad to see that things support both.  cc @randerzander

jakirkham: That's the idea. Though I do not think the `cudatoolkit` package relies on the `cuda` virtual package yet. Have raised issue ( https://github.com/conda/conda/issues/9115 ) to follow-up on this point.

Also there is still a lot of use of Conda 4.6 in RAPIDS (particular for doing releases). This is due to issues encountered when trying to build with Conda 4.7 and conda-build 3.18. Unfortunately I don't think the issues there were diagnosed (things needed to be pinned to working versions ASAP during the 0.8 release). It would be good to figure out why that was needed and whether we can upgrade as well as generate MREs for any issues that still exist.

jakirkham: Anyways even when these things are sorted there's no harm in providing explicit instructions (explicit over implicit ;).

PR #152, opened 2019-08-19
Test cleanup

quasiben:
- validate all tests pass with `py.test -s -v tests/`
- moves not tests to benchmarks
- lint all tests


quasiben:
@Akshay-Venkatesh if you have a minute can you review ?

tests/utils.py
```diff
@@ -0,0 +1,26 @@
...
+def set_env():
+    os.environ.update(normal_env)
+
+
+@contextlib.contextmanager
+@pytest.fixture
+def ucp_init():
+    try:
+        set_env()
+        yield ucp.init()
```
Akshay-Venkatesh: @quasiben can you explain why `yield` here?

mrocklin: Maybe this, just so that we don't imply that we're actually sending anything through

```suggestion
        ucp.init()  
        yield
```

quasiben:
The double quotes change came from running [black](https://black.readthedocs.io/en/stable/) .  `Black` is a code formatter.  It's not necessary but many projects in the PyData ecosystem enforce style to some degree.  In Rapids and Dask `black` is used fairly extensively

PR #154, opened 2019-08-20
ucp cleanup

madsbk:
Depend on #152

Simplifying by removing duplicate code.

### DIFF
--- a/ucp/_libs/ucp_py.pyx
+++ b/ucp/_libs/ucp_py.pyx
@@ -53,12 +53,11 @@ def update_pending_messages():
     """Checking for finished futures in `PENDING_MESSAGES`"""
     dones = []
     for msg, fut in PENDING_MESSAGES.items():
-        assert(not fut.done())
-        completed = msg.check()
-        if completed:
+        if fut.done():
+            dones.append(msg)
+        elif msg.check():
             dones.append(msg)
             fut.set_result(msg)
-
     for msg in dones:
         PENDING_MESSAGES.pop(msg)
 
--- a/ucp/_libs/ucp_py.pyx
+++ b/ucp/_libs/ucp_py.pyx
@@ -49,6 +49,17 @@ def initialized():
     global UCX_FILE_DESCRIPTOR
     return UCX_FILE_DESCRIPTOR >= 0
 
+def update_pending_messages():
+    """Checking for finished futures in `PENDING_MESSAGES`"""
+    dones = []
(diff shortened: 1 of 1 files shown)


ucp/_libs/ucp_py.pyx
```diff
@@ -43,17 +42,30 @@ include "ucp_py_buffer_helper.pyx"
...
 
+def initialized():
+    global UCX_FILE_DESCRIPTOR
+    return UCX_FILE_DESCRIPTOR >= 0
+
+def update_pending_messages():
+    """Checking for finished futures in `PENDING_MESSAGES`"""
+    dones = []
+    for msg, fut in PENDING_MESSAGES.items():
+        assert(not fut.done())
```
quasiben: Are you always guaranteed that futures are which in `PENDING` are `not done` ?  Could a message in PENDING be done by the time the assertion check is claled ?

madsbk: Good point, fixed in f50ac3ced6e511c9952c298662e3f9a1d758acbd where done futures are passed through. Notice, calling `.set_result()`on a done future is illigal.

mrocklin: If you're concerned about state changing within this for loop then that might not be a concern.  

Recall that because we're using only one thread, race conditions like this don't occur, unless there is an `await` statement somewhere, during which any other coroutine can come in and do anything.  This is one of the nice parts about using async/await and an event loop, rather than normal threads.

quasiben: Hmm, I ran [dask-cudf-full-test.py](https://github.com/quasiben/cudf-ucx-fail-scripts/blob/master/dask-cudf-full-test.py) and was able to generate an AssertionError in this code path.  

I see what you are saying -- msg are `asyncio.Future()` and the state will only be updated when we await that message.  @madsbk / @Akshay-Venkatesh is there another await for messages besides the call to `update_pending_messages`

Akshay-Venkatesh:
cc @tomaugspurger because we had spent many days to get progress to work right

ucp/_libs/ucp_py.pyx
Akshay-Venkatesh: I don't know if there is.

For context, this part of the code needs extensive testing and not just the unit tests in tests directory. We'll need benchmarks and dask workloads also running before we can accept the change that come out of here

quasiben:
I think we can merge this -- @Akshay-Venkatesh thoughts ?

PR #155, opened 2019-08-21
explicit event loop usage

Akshay-Venkatesh:
I'm not sure exactly why there's hang when I run `asyncio.run(main())`  [here](https://github.com/rapidsai/ucx-py/compare/fix-recv-into-benchmark?expand=1#diff-c9e616aa2908e76e06ec256178d32899L154) as opposed to getting event loop and having that loop run the coroutine returned by server and client. My guess is that the event loop used [here](https://github.com/rapidsai/ucx-py/blob/devel/ucp/_libs/ucp_py.pyx#L649) and the event loop used when just calling `asyncio.run(...)` may be different.

### DIFF
--- a/benchmarks/recv-into-client.py
+++ b/benchmarks/recv-into-client.py
@@ -2,6 +2,30 @@
 Benchmark recv_into.
 
 A client server pair take turns incrementing their arrays.
+
+Sample run:
+===========
+
+server:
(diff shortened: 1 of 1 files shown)


Akshay-Venkatesh:
Thanks @madsbk

I think we can add tests that stress both paths but maybe we can keep just current changes in the PR if you think it looks good

PR #158, opened 2019-08-22
Removing `struct data_buf`, now BufferRegon has the pointer directly

madsbk:
Removed `struct data_buf`, which removes an indirection and simplify the code significantly.

I will use this PR as a stepping stone to simplify `BufferRegion` and guarantee that `BufferRegion` always is in a valid state.


quasiben:
@Akshay-Venkatesh I ran all tests with this PR and everything passes.  When you have a moment can you review ?

madsbk:
@Akshay-Venkatesh, can we get this merged?

quasiben:
Thanks @madsbk !

PR #161, opened 2019-08-22
Implements BufferRegion Constructors

madsbk:
This PR simply adds some constructors to the `BufferRegion` class. 
The is a step on the way to make `BufferRegion` only constructable through constructors.

### DIFF
--- a/tests/test_buffer_region.py
+++ b/tests/test_buffer_region.py
@@ -4,9 +4,8 @@ import ucp
 
 
 def test_set_read():
-    obj = memoryview(b"hi")
-    buffer_region = ucp.BufferRegion()
-    buffer_region.populate_ptr(obj)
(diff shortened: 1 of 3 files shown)


madsbk:
@Akshay-Venkatesh and @quasiben, this PR is also ready for merging. 
It only adds some constructors to `BufferRegion` and use them in `ucp_py.pyx`

madsbk:
> For method names like `fromBuffer` and `newBuffer`, is camel casing / mixed casing the norm in python?
> This indicates otherwise: https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments
> 
> Changes look good to me otherwise.

Right, fixed!

quasiben:
Thanks @madsbk and @Akshay-Venkatesh for the review

PR #174, opened 2019-08-30
Add ucx dask test

quasiben:
This PR adds a parameterized test version of notebook experiments we ran here:
https://github.com/quasiben/cudf-ucx-fail-scripts/blob/master/dummy-column.ipynb

cc @shwina @mrocklin


tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,60 @@
...
+import numpy as np
+import asyncio
+import dask.dataframe as dd
+from distributed import Scheduler, Worker, Client, Nanny
+from distributed.utils import log_errors
+
+#async with Nanny(s.address, protocol='ucx', nthreads=1,
+# nanny is really a worker running on a defined CUDA DEVICE
+protocol = 'ucx'
+interface = 'ib3'
```
Akshay-Venkatesh: Seems like this test shouldn't be tested outside of a `dgx-1`. If so, can you skip the test after detecting that?

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,60 @@
...
+                    env={'CUDA_VISIBLE_DEVICES': '7,6'}, interface=interface,
+                    nthreads=1) as w2:
+                async with Client(s.address, asynchronous=True) as c:
+                    with log_errors():
+
+                        def set_nb_context(x=None):
+                            import numba.cuda
+                            try:
+                                numba.cuda.current_context()
+                            except Exception:
```
Akshay-Venkatesh: I don't see CUDA_VISIBLE_DEVICES (CVD) being set for the client explicitly. Is that intended? If yes, does the test expect client to make allocations on device 0 (which will be default if not explicitly set) and have this be communicated to workers/scheduler? Also what is CVD set to in the case of scheduler?
IMO, the client, worker(s), scheduler should all have the same consistent view of CUDA capable devices -- so either set CVD for all entities or set for none.

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,60 @@
...
+                            # results = await client.gather(futures, asynchronous=True)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("cudf_obj_generator", [
+    lambda x: cudf.Series(np.arange(10000))._columns,
+    lambda x: cudf.Series(np.arange(10000))
+    ]
+)
+async def test_send_recv_cupy(event_loop, cudf_obj_generator):
```
Akshay-Venkatesh: Can you provide a brief description of the intent of this test? Also the test is named `send_recv_cupy` but it seems like the goal is to move around `cudf` objects. Maybe you meant `send_recv_cudf_series` or something.

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,185 @@
...
+        name="client", target=client, args=[env2, port]
+    )
+
+    server_process.start()
+    client_process.start()
+
+    server_process.join()
+    client_process.join()
+
+    ucp.fin()
```
Akshay-Venkatesh: `ucp.fin()` is already [here](https://github.com/rapidsai/ucx-py/pull/174/files#diff-1cca3066e91b8737cf0377fe360bbfd3R145) and [here](https://github.com/rapidsai/ucx-py/pull/174/files#diff-1cca3066e91b8737cf0377fe360bbfd3R70). Do we need this? I guess calling `fin()` after `fin()` isn't erroneous but maybe not call that because we don't reference ucp in this function at all?

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,185 @@
...
+    os.environ.update(env)
+    print("ENV SERVER: ", os.environ)
+
+    async def f(lisitener_port):
+        ucp.init()
+
+        # coroutine shows up when the client asks
+        # to connect
+        async def write(ep, li):
+            await asyncio.sleep(0.1)
```
Akshay-Venkatesh: what purpose does the sleep operation serve here?

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,185 @@
...
+            for idx, frame in enumerate(frames):
+                await ep.send_obj(frame)
+
+            print("CONFIRM RECEIPT")
+            resp = await ep.recv_future()
+            obj = ucp.get_obj_from_msg(resp)
+            assert obj == b"OK"
+            print("Shutting Down Server...")
+            ucp.stop_listener(li)
+            ucp.fin()
```
Akshay-Venkatesh: `ucp.fin()` is called inside a coroutine and again called [here](https://github.com/rapidsai/ucx-py/pull/174/files#diff-1cca3066e91b8737cf0377fe360bbfd3R154). The latter alone should suffice.

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,185 @@
...
+            frames.append(frame)
+
+        header = pickle.loads(frames[0])
+        frames = frames[1:]  #
+        typ = pickle.loads(header["type"])
+        cuda_obj = typ.deserialize(header, frames)
+
+        print(cuda_obj, len(cuda_obj), type(cuda_obj))
+        asyncio.sleep(1)
+        # how should we confirm data is on two GPUs ?
```
Akshay-Venkatesh: also, how should we confirm if final object received on client is identical to that sent by server?

tests/test_send_recv_two_workers.py
quasiben: This was designed to give a little bit of time for the client to startup.  Should we take it out ?

tests/test_send_recv_two_workers.py
quasiben: I think we cleaned up the ucp.fin calls now

tests/test-ucx-w-dask-workers.py
quasiben: resolved by ed0cce0c2c30700449aefbe1634587f34b8e7373

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,109 @@
...
+    "CUDA_VISIBLE_DEVICES": "0,1",
+}
+os.environ.update(w_env)
+
+w_env_2 = w_env.copy()
+w_env_2["CUDA_VISIBLE_DEVICES"] = "1,0"
+
+async def f(cudf_obj_generator):
+    async with Scheduler(protocol=protocol, interface=interface,
+    dashboard_address=':8789') as s:
```
mrocklin: Do we need a dashboard?  If not, maybe `dashboard_address=None` or just omit this keyword?

tests/test-ucx-w-dask-workers.py
mrocklin: The client is in the main process, so if we're using environment variables then this will have to be handled by the user calling pytest like the following:

```
CUDA_VISIBLE_DEVICES=0,1 py.test foo.py
```

Alternatively, are there ways to set this programmatically within UCX?

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,109 @@
...
+                    nthreads=1) as w2:
+                async with Client(s.address, asynchronous=True) as c:
+                    with log_errors():
+
+                        def set_nb_context(x=None):
+                            import numba.cuda
+                            try:
+                                numba.cuda.current_context()
+                            except Exception:
+                                print("FAILED EXCEPTION!")
```
mrocklin: I think that we should avoid print statements in testing code.  Lets assume that no one will be reading these logs.  If we want to check something, we should check it programmatically.

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,109 @@
...
+
+                        print(f"SETTING CUDA CONTEXT ON WORKERS: {w.worker_address} / {w2.worker_address}")
+                        out = await c.run(set_nb_context, workers=[w.worker_address, w2.worker_address])
```
mrocklin: We should be able to remove this if we use the preload script, which is more robust anyway.

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,109 @@
...
+                        print("Creating and Mapping CUDA Objects")
+                        # offset worker two for unique hash names inside of dask
+                        N = 100
+                        left = c.map(cudf_obj_generator,
+                                     range(N), workers=[w.worker_address])
+                        right = c.map(cudf_obj_generator,
+                                        range(1, N+1), workers=[w2.worker_address])
+                        print("Gather CUDA Objects")
+                        futures = c.map(lambda x, y: (x,y), left, right, priority=10)
+                        results = await c.gather(futures, asynchronous=True)
```
mrocklin:
```suggestion
                        results = await c.gather(futures)
```

You don't need to add asynchronous=True if the client has been created asynchronously.

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,109 @@
...
+async def f(cudf_obj_generator):
+    async with Scheduler(protocol=protocol, interface=interface,
+    dashboard_address=':8789') as s:
+        async with Nanny(s.address, protocol=protocol, nthreads=1,
+                memory_limit='32GB', interface=interface,
+                env=w_env,
+                ) as w:
+            async with Nanny(s.address, protocol=protocol,memory_limit='32gb',
+                    env=w_env_2, interface=interface,
+                    nthreads=1) as w2:
```
mrocklin: You might want to use the following:

```python
preload=["dask_cuda.initialize_context"]
```

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,109 @@
...
+                        for i in range(N):
+                            assert_eq(results[i][0], cudf_obj_generator(i))
+                            assert_eq(results[i][1], cudf_obj_generator(i))
+
+                        print("ALL DONE!")
+
+
+def column(x):
+    import cudf
+    import numpy as np
```
mrocklin: I recommend that we import these at the top level instead of here.  If you're guarding against them not being present, then let's use `pytest.importorskip`

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,261 @@
...
+    nlinks = pynvml.NVML_NVLINK_MAX_LINKS
+    # ngpus = pynvml.nvmlDeviceGetCount()
+    handle = pynvml.nvmlDeviceGetHandleByIndex(cuda_dev_id)
+    rx = 0
+    tx = 0
+    for i in range(nlinks):
+        transfer = pynvml.nvmlDeviceGetNvLinkUtilizationCounter(handle, i, 1)
+        rx += transfer["rx"]
+        tx += transfer["tx"]
+    return rx, tx
```
mrocklin: We might want to do something like this as a pytest fixture to ensure that we're using nvlink

```python
@pytest.fixture
def uses_nvlink():
    # setup
    yield
    # test that transfer occurred
```

We would then include this within our tests above

```python
def test_foo(uses_nvlink):
```

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.mark.skipif(mark_is_not_dgx, reason="Not a DGX")
+@pytest.mark.asyncio
+@pytest.mark.parametrize("cudf_obj_generator", [
+    column,
+    series,
+    cupy,
+    dataframe
+])
+async def test_send_recv_cuda(event_loop, cudf_obj_generator):
+    await f(cudf_obj_generator)
```
mrocklin: Why this wrapping?  Why not inline `f` here directly?

tests/test-ucx-w-dask-workers.py
quasiben: i am somewhat reluctant to use this.  It would mean bringing in dask-cuda for testing and i think we should minimize this when possible.  We could initialize with the following object as you suggested before.  

```
class CudaContext:
    def __init__(self, *args, **kwargs):
        pass
    def setup(self, worker):
        import numba.cuda
        numba.cuda.current_context()
    def teardown(self, worker):
        pass

client.register_worker_plugin(CudaContext())
``` 

what do you think ?

tests/test-ucx-w-dask-workers.py
quasiben: I am guarding against cuda context setting but perhaps this is overkill

tests/test-ucx-w-dask-workers.py
quasiben: we changed the name to `test_send_recv_cuda` . The intent of the test is to demonstrate high bandwith communication between dask workers in a controlled manner.  in this case, we are simply sending many cudf objects back and forth.  These should always pass -- at one point they did not.

tests/test-ucx-w-dask-workers.py
```diff
@@ -0,0 +1,199 @@
...
+                        results = []
+                        for i in range(npartitions):
+                            start = i * (left_n_rows // npartitions)
+                            end = (i + 1) * (left_n_rows // npartitions)
+                            worker = (w1.worker_address, w2.worker_address)[i % 2]
+                            results.append(c.submit(lambda x: x.iloc[start:end], future, workers=[worker]))
+
+                        del future
+                        for result in results:
+                            await result
```
shwina: @mrocklin @quasiben 

The latter part of this test and the region commented above it should both do roughly the same thing - correct?

Only the commented region produces a hang.

tests/test_send_recv_two_workers.py
quasiben: resolved

PR #176, opened 2019-09-04
BufferRegion: Automatic typestr / format conversion

madsbk:
`BufferRegion` now only has the `typestr` property. 
When using the buffer protocol, `BufferRegion` with seamlessly convert `typestr` to the format string expected by Python's Buffer Protocol.


madsbk:
Because of #180, this is not relevant anymore.

PR #178, opened 2019-09-05
[FEA] print unfinished messages 

madsbk:
This PR adds `unfinished_messages_info()`, which returns info about all current unfinished messages.

The idea is that if the execution has deadlocked, you can _Control-C_ out to pdb and call `ucp.unfinished_messages_info()`.

Example, if I add an extra receive in `test_send_recv_bytes()`: 
```python
@pytest.mark.asyncio
@pytest.mark.parametrize("size", msg_sizes)
async def test_send_recv_bytes(size):
    x = "a"
    x = x * size
    msg = bytes(x, encoding="utf-8")

    async with echo_pair() as (_, client):
        await client.send_obj(bytes(str(size), encoding="utf-8"))
        await client.send_obj(msg)
        resp = await client.recv_obj(len(msg))
        result = ucp.get_obj_from_msg(resp)
        await client.recv_obj(42) # extra 42 bytes receive 
    assert result.tobytes() == msg
```
It will then hang; I can _Control-C_ out to pdb and see the hanging `recv_obj`:
```
(Pdb) import ucp; print(ucp.unfinished_messages_info())
Unfinished messages:
  <Message - name: "recv_obj", cuda: False, comm_len: 256, alloc_len: -1, endpoint: "<Endpoint "client to server" on 10.33.225.165 connected to 10.33.225.165:13337>">
```

### DIFF
--- a/ucp/_libs/ucp_py.pyx
+++ b/ucp/_libs/ucp_py.pyx
@@ -68,6 +68,22 @@ def unfinished_messages_info():
         ret += "  %s\n" % msg
     return ret
 
+class MsgHangError(Exception):
+    pass
+
+def unfinished_messages_exception_trigger():
+    """Sets exception on all current unfinished messages 
+       and trigger them"""
+    update_pending_messages()
+    msg_count = len(PENDING_MESSAGES)
+    for msg, fut in PENDING_MESSAGES.items():
+        try:
+            fut.set_exception(MsgHangError(str(msg)))
+            fut.get_loop().run_until_complete(fut)
+        except MsgHangError as e:
+            print("MsgHangError: %s" %e)
+
+
 def get_ucp_worker():
     return <size_t>get_worker()
 
--- a/ucp/_libs/ucp_py.pyx
+++ b/ucp/_libs/ucp_py.pyx
@@ -699,7 +699,7 @@ def fin():
         raise RuntimeError("UCX-Py fin(): ucp_py_finalize() failed!")
 
 @ucp_logger
-async def get_endpoint(peer_ip, peer_port, timeout=None, name=None):
+async def get_endpoint(peer_ip, peer_port, timeout=None, name="client to server"):
     """Connect to a peer running at `peer_ip` and `peer_port`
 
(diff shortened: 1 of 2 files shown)


quasiben:
Thank you for taking a look at this @madsbk !  hopefully this proves useful for debugging purposes:
cc @shwina

quasiben:
Using this PR I cam across some interesting output on a hang:

>   <Message - name: "recv-future", cuda: False, comm_len: 0, alloc_len: 0, endpoint: "<Endpoint "None" on 10.33.225.165 connected to 10.33.225.165:41034>">

The name is `None`.  Other outputs are much more reasonable:

>   <Message - name: "recv-future", cuda: False, comm_len: 0, alloc_len: 0, endpoint: "<Endpoint "server to client" on 10.33.225.165 connected to unavailable>">

Does this indicate something in dask trying to close endpoint without fully shutting it down?

quasiben:
I think the answer to my question is no.  the default name is `None`

madsbk:
> I think the answer to my question is no. the default name is `None`

I have updated the default name of `get_endpoint()` to `client to server`.

ucp/_libs/ucp_py.pyx
```diff
@@ -59,6 +60,30 @@ def update_pending_messages():
     for msg in dones:
         PENDING_MESSAGES.pop(msg)
 
+def unfinished_messages_info():
+    """Returns info of all current unfinished messages"""
+    update_pending_messages()
```
Akshay-Venkatesh: @madsbk I have a concern with the usage of `update_pending_messages()` in a function which is expected to report on the state of the communication channel. `update_pending_messages()` internally calls `msg.check()` which pings the progress engine and can result in communication progress. I think this is not intended for the purpose of reporting information about unfinished messages. What are your thoughts on this?

madsbk: I think you are completly right :)
Fixed in b212df0

madsbk:
Because of #180, this is not relevant anymore.

Issue #123, opened 2019-06-14
Adding license file


jakirkham:
Would be good to add a license file. Guessing it should be the same as UCX's license. Though please correct me if not.

quasiben:
@mrocklin @Akshay-Venkatesh do you have an idea what the license should be ?

jakirkham:
If we are upstreaming this to UCX, we might also want to touch base with them to see what kind of licensing would be acceptable.

MattBBaker:
The main UCX repo is 3 clause BSD. Using that will be easiest.

mrocklin:
At the same time most RAPIDS projects use Apache 2.

pentschev:
If this repo will be long-maintained by RAPIDS, I would suggest sticking to Apache 2 license, per the reason pointed out by @mrocklin.

jakirkham:
Are we still planning on upstreaming this to UCX? If so, I’d suggest just using the UCX license file here (i.e. BSD-3-Clause) as @MattBBaker has stated. Otherwise we are merely complicating the upstreaming process for ourselves later.

mrocklin:
Sounds good. Go for it. You should probably check in with Akshay.

jakirkham:
@Akshay-Venkatesh , can you please weigh in here (when you have a chance)?

Akshay-Venkatesh:
The plan is still to upstream to UCX but those are probably aimed for post-UCX 1.7 release. Suggestion from @MattBBaker sounds reasonable to me as well.

jakirkham:
Thanks all. Have submitted PR ( https://github.com/rapidsai/ucx-py/pull/182 ) to address this. Please take a look.

PR #182, opened 2019-09-09
Include license file

jakirkham:
Copies over the OpenUCX license file and ensures it is included in various Python packages that may be produced.

Fixes https://github.com/rapidsai/ucx-py/issues/123

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,34 @@
+Copyright (c) 2014-2015      UT-Battelle, LLC. All rights reserved.
+Copyright (C) 2014-2015      Mellanox Technologies Ltd. All rights reserved.
+Copyright (C) 2014-2015      The University of Houston System. All rights reserved.
+Copyright (C) 2015           The University of Tennessee and The University 
+                             of Tennessee Research Foundation. All rights reserved.
+Copyright (C) 2016           ARM Ltd. All rights reserved.
+Copyright (c) 2016           Los Alamos National Security, LLC. All rights reserved.
+Copyright (C) 2016-2017      Advanced Micro Devices, Inc.  All rights reserved.
+Copyright (C) 2019           UChicago Argonne, LLC.  All rights reserved.
+
+Redistribution and use in source and binary forms, with or without 
+modification, are permitted provided that the following conditions 
+are met:
+
+1. Redistributions of source code must retain the above copyright 
+notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the 
+documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its 
+contributors may be used to endorse or promote products derived from 
+this software without specific prior written permission.
+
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -0,0 +1,34 @@
+Copyright (c) 2014-2015      UT-Battelle, LLC. All rights reserved.
+Copyright (C) 2014-2015      Mellanox Technologies Ltd. All rights reserved.
+Copyright (C) 2014-2015      The University of Houston System. All rights reserved.
+Copyright (C) 2015           The University of Tennessee and The University 
+                             of Tennessee Research Foundation. All rights reserved.
+Copyright (C) 2016           ARM Ltd. All rights reserved.
+Copyright (c) 2016           Los Alamos National Security, LLC. All rights reserved.
+Copyright (C) 2016-2017      Advanced Micro Devices, Inc.  All rights reserved.
+Copyright (C) 2019           UChicago Argonne, LLC.  All rights reserved.
```
jakirkham: Should NVIDIA be listed here somewhere? And if so, should this happen in [the OpenUCX license file]( https://github.com/openucx/ucx/blob/ff93b653fe48d6d4fbbfff7eb66cdee52ed8d9f8/LICENSE#L1-L9 ) as well?

pentschev: I would generally expect that in this repo only the institutions that contributed to it are listed. By checking [contributors to this repo](https://github.com/rapidsai/ucx-py/graphs/contributors), all of them are from NVIDIA, with the exception of @TomAugspurger. At some point, if this gets merged into https://github.com/openucx/ucx, then we would have NVIDIA listed there along all other institutions, but copying over all institutions that contributed to ucx mainline doesn't make much sense to me when it comes to ucx-py, which is not a fork or has no direct relationship apart from being a dependency.

jakirkham: Sure. We can change that.

My question is more should NVIDIA also be listed as a contributor in the OpenUCX repo's license file? Thoughts @Akshay-Venkatesh?

Akshay-Venkatesh: That's correct, @jakirkham 
Thanks for pointing this out. NVIDIA should be listed. I'll make a PR to add NVIDIA to the main LICENSE file here https://github.com/openucx/ucx/blob/master/LICENSE

jakirkham: Thanks Akshay! Feel free to link the PR here. 🙂

jakirkham: I've added NVIDIA here and dropped the other entities.

Am unclear on whether Anaconda should be listed here. @mrocklin @TomAugspurger, would you be able to comment on this?

TomAugspurger: I'm not sure what the policy is, but I'm happy to do whatever is easiest for the project (including "waiving" all copyright if that's even necessary / possible).

jakirkham: After asking internally was told that only NVIDIA should be listed here. So will leave as-is.

jakirkham:
@quasiben, are you comfortable merging here?

quasiben:
yes, thanks for the ping

jakirkham:
Thanks Ben!

PR #190, opened 2019-09-26
[REVIEW] Malloc check and leak fix

madsbk:
This PR fixes a memory leak of `_listener_callback_args` and checks every `malloc` for _out of memory_.


madsbk:
rerun tests

quasiben:
Looks good to me.  Should we merge ?

PR #192, opened 2019-09-26
Support Python 3.6

mrocklin:
Previously we used asyncio features that were only available in Python 3.7
This commit replaces those calls with options that are also available in 3.6

### DIFF
--- a/ucp/_libs/send_recv.pyx
+++ b/ucp/_libs/send_recv.pyx
@@ -39,7 +39,7 @@ cdef void _send_callback(void *request, ucs_status_t status):
         req.finished = True
         return
     cdef object future = <object> req.future
-    if future.get_loop().is_closed():
+    if asyncio.get_event_loop().is_closed():
         pass
     elif status == UCS_ERR_CANCELED:
         future.set_exception(UCXCanceled())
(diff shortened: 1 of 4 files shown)


ucp/_libs/send_recv.pyx
```diff
@@ -39,7 +39,7 @@ cdef void _send_callback(void *request, ucs_status_t status):
         req.finished = True
         return
     cdef object future = <object> req.future
-    if future.get_loop().is_closed():
+    if asyncio.get_event_loop().is_closed():
```
mrocklin: @madsbk is this safe?  

In particular, will we only call this function from within the thread running the event loop?

mrocklin:
Tests mostly ran, but I didn't bother to install cupy, so I didn't run the full test suite locally.

quasiben:
Thanks for this @mrocklin !

mrocklin:
Also, thanks @jakirkham for the `ucx` package.  Going from zero-to-developing was much easier this time around :)

ucp/_libs/send_recv.pyx
madsbk: Yes, it should be fine

quasiben: Thanks @madsbk

quasiben:
merging in.  Thanks again @mrocklin !

Note, I changed the branch we are merging into to `branch-0.10`

PR #203, opened 2019-10-01
Rely on setuptools' support of Cython

jakirkham:
By default setuptools knows how to handle Cython extensions and will try to build them through Cython or fallback to compiling C files (if available).


quasiben:
Thanks @jakirkham !

PR #204, opened 2019-10-01
Relax NumPy requirement

jakirkham:
Relaxes the requirement on NumPy where possible. Leverages fixed size C arrays when it makes sense. Drops an unneeded NumPy import. Also rescopes a NumPy import to the section of code where it is needed.

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -42,6 +42,7 @@ async def listener_handler(ucp_endpoint, ucp_worker, config, func):
 
     # Get the tags from the client and create a new Endpoint
     cdef uint64_t[4] tags
+    cdef uint64_t[::1] tags_mv = <uint64_t[:4:1]>(&tags[0])
     await stream_recv(ucp_endpoint, tags, sizeof(tags))
     ep = Endpoint(ucp_endpoint, ucp_worker, config, tags[0], tags[1], tags[2], tags[3])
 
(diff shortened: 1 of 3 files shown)


quasiben:
@madsbk can you give this a look ?

jakirkham:
Thanks @madsbk!

Yeah I ran into some weird issues with fixed size arrays in Cython. They seem to treat them like Python lists even though they are not.

I haven’t quite wrapped my head around it. Though agree we should look at another way to do this.

xref: https://github.com/cython/cython/issues/3162
xref: https://github.com/cython/cython/issues/3161

jakirkham:
@madsbk, I've fixed the CI failures. Could you please take another look?

ucp/_libs/core.pyx
```diff
@@ -230,18 +230,23 @@ cdef class ApplicationContext:
...
-            np.uint64(hash(uuid.uuid4())),
-            np.uint64(hash(uuid.uuid4()))
-        )
-        tags = np.array(
-            [ret._recv_tag, ret._send_tag, ret._ctrl_recv_tag, ret._ctrl_send_tag],
-            dtype="uint64"
+            tags[1],
+            tags[0],
+            tags[3],
+            tags[2]
```
jakirkham: This ordering surprised me a bit. Naively I would have taken the indices in order. However this particular ordering was needed to match the previous code here. We might benefit from using a `struct` with named entries instead. This would clarify their meaning and purpose. Have raised issue ( https://github.com/rapidsai/ucx-py/issues/210 ) on this point.

jakirkham:
Thanks Mads! 😄

PR #212, opened 2019-10-03
Buffer contiguous check

### LINKED ISSUE #211: TODO: check that data is contiguous?
We have this line in the codebase that appears to have moved around a bit. Is it still relevant? If so, how should we handle this?

https://github.com/rapidsai/ucx-py/blob/b68e727bdc27edc326cf16ef8c36debdfbd8eb3f/ucp/_libs/utils.pyx#L71

madsbk:
Fixes #211


quasiben:
Thanks @madsbk and @jakirkham for reporting

Issue #210, opened 2019-10-02
Place Endpoint's tags in a struct


jakirkham:
As [tags used by `Endpoint`]( https://github.com/rapidsai/ucx-py/blob/14e93062da4419a66ac875d28f62863fdfa06416/ucp/_libs/core.pyx#L299-L302 ) seem to always be dealt with [in batches]( https://github.com/rapidsai/ucx-py/blob/14e93062da4419a66ac875d28f62863fdfa06416/ucp/_libs/core.pyx#L44 ), it might make sense to just create a `struct` to contain them all with appropriate names. This should help clarify the code in points where [the order of indices for tags is non-intuitive]( https://github.com/rapidsai/ucx-py/blob/14e93062da4419a66ac875d28f62863fdfa06416/ucp/_libs/core.pyx#L50 ).

madsbk:
I totally agree :)

jakirkham:
Put together PR ( https://github.com/rapidsai/ucx-py/pull/215 ) to address this.

PR #215, opened 2019-10-03
Use struct tags

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/210
Fixes https://github.com/rapidsai/ucx-py/issues/223
Merges https://github.com/rapidsai/ucx-py/pull/228
Merges https://github.com/rapidsai/ucx-py/pull/230

Creates a `struct` called `Tags` to hold the different tag values when sending or receiving them. Should make it easier to follow which tag is being used where.

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -103,8 +103,8 @@ async def listener_handler(ucp_endpoint, ucp_worker, config, func):
         loop.set_exception_handler(asyncio_handle_exception)
 
     # Get the tags from the client and create a new Endpoint
-    cdef Tags[1] tags
-    cdef Tags[::1] tags_mv = <Tags[:1:1]>(&tags[0])
+    cdef Tags tags
+    cdef Tags[::1] tags_mv = <Tags[:1:1]>(&tags)
     await stream_recv(ucp_endpoint, tags_mv, tags_mv.nbytes)
     ep = Endpoint(ucp_endpoint, ucp_worker, config,
                   tags.recv_tag, tags.send_tag,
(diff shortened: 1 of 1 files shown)


jakirkham:
@madsbk, I put this together, but it still has some issue that I wasn't able to spot yet. Maybe you will be able to catch it? Anyways it causes the tests to hang when running locally. So it seems like tag order might be off somewhere.

jakirkham:
Should add this is low priority. It would be nice to do. However if there are other things needed for the release, they should be focused on first. This could happen after the release and it would be ok. 🙂

ucp/_libs/core.pyx
```diff
@@ -289,6 +293,13 @@ cdef class ApplicationContext:
         return self.config
 
 
+cdef struct Tags:
+    uint64_t send_tag
+    uint64_t recv_tag
+    uint64_t ctrl_send_tag
+    uint64_t ctrl_recv_tag
```
MattBBaker: Why are there separate send and receive tags?

jakirkham: Am mirroring [this code]( https://github.com/rapidsai/ucx-py/blob/7b739a48dc4b647edd7234424e2dd94720606c86/ucp/_libs/core.pyx#L329-L332 ).

jakirkham: Perhaps @madsbk can clarify on why he structured it that way? 🙂

quasiben:
rerun tests

jakirkham:
I think the CI error here is real. Am able to reproduce it locally as well. Suspect it is some ordering issue with the fields, but haven't quite identified where.

jakirkham:
Ok have fixed the bug and did a bit of tidying. Should be ready for another review @quasiben @madsbk 🙂

ucp/_libs/core.pyx
madsbk: @MattBBaker, I don't think that we need separate send and receive tags anymore but they are nice to have in the log messages (making it easy to follow the communication flow).
If we use the pointer address of Endpoints as IDs in the log instead, which is the same UCX does, we can combine the send/recv tags and still have a nice log.

But I think we should merge this PR and make another PR that combines the send/recv tags.

jakirkham:
Thanks @madsbk. Pushed a couple small commits to tidy a few last things. Otherwise should be good to go on my end. Please let me know if there's anything else 🙂

ucp/_libs/core.pyx
MattBBaker: It won't help with the tags, because I need those to be unified to avoid the deadlock. But I should chime in on the API issues.

madsbk: @mt-jones check https://github.com/rapidsai/ucx-py/pull/238

Issue #208, opened 2019-10-02
Stop Listener


quasiben:
@madsbk what do you think about hanging an explicit `close`/`stop_listener` function off of the Listener object.  Right now the listener only gives up the connection when it's deleted, correct ?

PR #222, opened 2019-10-04
Implementing Listener.close()

madsbk:
Fixes: #208

### DIFF
--- a/tests/test_shutdown.py
+++ b/tests/test_shutdown.py
@@ -25,8 +25,8 @@ async def test_server_shutdown():
         with pytest.raises(ucp.exceptions.UCXCanceled):
             await ep.recv(msg)
 
-    lf = ucp.create_listener(server_node)
-    await client_node(lf.port)
+    listener = ucp.create_listener(server_node)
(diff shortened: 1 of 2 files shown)


quasiben:
A wrote a small test for this:

```python
@pytest.mark.asyncio
async def test_listener_shutdown():
    """The listener shuts down"""

    async def client_node(port):
        ep = await ucp.create_endpoint(ucp.get_address(), port)
        msg = np.empty(10 ** 6)
        await asyncio.gather(ep.recv(msg))

    async def server_node(ep):
        msg = np.empty(10 ** 6)
        await ep.recv(msg)

    lf = ucp.create_listener(server_node)
    assert lf._closed == True
```

And it exposed a couple things:

1) I think, though could be wrong, `__dealloc__` is a private function not accessible from python.  So calling `close()` will fail
2) What do you think about making `_closed` `readonly` , then we could check on the Python side

madsbk:
Thanks @quasiben! 
I have added the test and `close` property as you suggested.

quasiben:
Thanks @madsbk !

PR #225, opened 2019-10-07
UCX Options Programmable 

### LINKED ISSUE #83: Programmable env vars
Often we are setting env vars like:
- `UCX_MEMTYPE_CACHE`
- `UCX_NET_DEVICES
- `UCX_RNDV_SCHEME`

Being able to programmatically set these values would make for a nicer experience.  One suggestion might be to build a Context object prior to importing ucp_libs and set/get values from that object

```python
context = ucp.Context(UCX_MEMTYPE_CACHE='n', ...)
ucp.init()
...
```

madsbk:
Fixes #83.

This makes it possible to specify UCX options programmable using `ucp.init()`:
```python
def init(options={}, env_takes_precedence=False):
    """
    Initiate UCX. Usually this is done automatically at the first API call
    but this function makes it possible to set UCX options programmable.
    Alternatively, UCX options can be specified through environment variables.

    Parameters
    ----------
    options: dict, optional
        UCX options send to the underlaying UCX library
    env_takes_precedence: bool, optional
        Whether environment variables takes precedence over the `options`
        specified here.
    """
```

cc @pentschev, @mrocklin

### DIFF
--- a/ucp/public_api.py
+++ b/ucp/public_api.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
+import os
 from ._libs import core
 from ._libs.core import Endpoint  # TODO: define a public Endpoint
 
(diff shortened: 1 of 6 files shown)


MattBBaker:
If I am reading this code right, it looks like if you call ucp.get_config() it will initialize the context? It seems it would be better if ucp.get_config() could be called without the implicit initialization of the context, so that a user could look through the config results without having to call reset().

ucp/public_api.py
```diff
@@ -24,6 +25,34 @@ def _get_ctx():
 # the functions here.
 
 
+def init(options={}, env_takes_precedence=False):
```
pentschev: Can we also be allowed to pass the `UCXPY_IFNAME` here? It seems that this isn't possible now.

pentschev: That would also need to appear in the `get_config()` then.

pentschev: Nevermind, @quasiben helped me realize `UCXPY_IFNAME` isn't necessary to use with Dask.

madsbk:
> If I am reading this code right, it looks like if you call ucp.get_config() it will initialize the context? It seems it would be better if ucp.get_config() could be called without the implicit initialization of the context, so that a user could look through the config results without having to call reset().

Good point:
```python
def get_config():
    """
    Returns all UCX configuration options as a dict.
    If UCX is uninitiated, the options returned are the
    options used if UCX were to be initialized now.
    Notice, this function doesn't initiate UCX.

    Returns
    -------
    dict
        The current UCX configuration options
    """
```

pentschev:
Mostly yes, I'm just trying to finish the dask-cuda adaptations necessary to ensure everything works, but so far it seems that it does.

quasiben:
Any reason not to merge ?

pentschev:
No, I will merge as is, since my tests were successful, if we need other changes, we can address later.

Thanks @madsbk for the PR and @quasiben and @MattBBaker for reviews.

Issue #223, opened 2019-10-04
Cython warning: "may be used uninitialized"


madsbk:
When compiling `branch-0.10` I see this Cython warning:
```
ucp/_libs/core.c: In function ‘__pyx_gb_3ucp_5_libs_4core_18ApplicationContext_8generator1’:
ucp/_libs/core.c:7470:32: warning: ‘__pyx_t_14’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     __pyx_cur_scope->__pyx_t_0 = __pyx_t_14;
```
It refers to: https://github.com/rapidsai/ucx-py/blob/eba2cd68cd6da90558cd532b32b334659ff66c4d/ucp/_libs/core.pyx#L250
It still gives the warning when I change all arguments to something like:
```python
await stream_send(42, bytearray(), 42)
```
However, if I remove the `await` the warning goes away.

Any thoughts?

quasiben:
The warning points to this bit of generated c code:
```c
  if (likely(__pyx_r)) {
    __pyx_cur_scope->__pyx_t_0 = __pyx_t_14;
    __pyx_cur_scope->__pyx_t_1 = __pyx_t_17;
    __pyx_cur_scope->__pyx_t_2 = __pyx_t_19;
    __pyx_cur_scope->__pyx_t_3 = __pyx_t_21;
    __pyx_cur_scope->__pyx_t_4 = __pyx_t_23;
```

I went digging through some old cython issues and found that there have been issues with conditionals: https://github.com/cython/cython/issues/2274 -- but I don't that as being the issue here

jakirkham:
FWIW (not that I completely understand why) this warning goes away for me with PR ( https://github.com/rapidsai/ucx-py/pull/215 ).

jakirkham:
Yeah it seems to have something to do with the `for`-loop. If I unroll the `for`-loop, the warning also goes away. ( https://github.com/rapidsai/ucx-py/pull/230 )

Maybe some issue with how Cython is generating code there as Ben has suggested.

jakirkham:
Came up with a somewhat similar (though still not quite the same) warning and raised upstream as issue ( https://github.com/cython/cython/issues/3172 ).

jakirkham:
Now that PR ( https://github.com/rapidsai/ucx-py/pull/215 ) is working and already fixes some other issues (in addition to this one), maybe we should just go ahead with that.

jakirkham:
[...]

PR #230, opened 2019-10-08
Unroll `for`-loop used to assign UUIDs to tags

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/223

Not sure if this is better, but it does get rid of the warning.

cc @madsbk


jakirkham:
Superseded by PR ( https://github.com/rapidsai/ucx-py/pull/215 ).

PR #233, opened 2019-10-09
[REVIEW] Fix/nvlink test

quasiben:
PR adds nvlink test (two processes and two independent GPUs)

```
UCXPY_IFNAME=enp1s0f0 UCX_MEMTYPE_CACHE=n UCX_TLS=rc,tcp,sockcm,cuda_copy,cuda_ipc UCX_SOCKADDR_TLS_PRIORITY=sockcm py.test -s -v "tests/test_send_recv_two_workers.py::test_send_recv_cudf[dataframe]" 
```

fixes https://github.com/rapidsai/ucx-py/pull/233


quasiben:
Currently running with:
> UCXPY_IFNAME=enp1s0f0 CUDA_VISIBLE_DEVICES=0,1 UCX_MEMTYPE_CACHE=n UCX_TLS=tcp,cuda_copy,cuda_ipc,sockcm UCX_SOCKADDR_TLS_PRIORITY=sockcm py.test -s -v "tests/test_send_recv_two_workers.py::test_send_recv_cudf[dataframe]"

quasiben:
@madsbk can you look this over ?  The goal of this test is to show that nvlink is working correctly.  Currently, this is show no NVLINK usage with the following command:

> UCXPY_IFNAME=enp1s0f0 CUDA_VISIBLE_DEVICES=0,1 UCX_MEMTYPE_CACHE=n UCX_TLS=tcp,cuda_copy,cuda_ipc,sockcm UCX_SOCKADDR_TLS_PRIORITY=sockcm py.test -s -v "tests/test_send_recv_two_workers.py::test_send_recv_cudf"

Note that the PR has code for using pynvml.  I'm currently working with ops to figure out why nvml calls required sudo

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,274 @@
...
+            recv_msg = msg.tobytes()
+            assert recv_msg == close_msg
+            print("Shutting Down Server...")
+            await ep.signal_shutdown()
+            ep.close()
+            lf.close()
+
+        lf = ucp.create_listener(write, port=listener_port)
+        try:
+            while not lf.closed:
```
quasiben: @madsbk this while loop seems wrong.  in the old bindings  lf had a coroutine accessor attached to it and I could hand that back to the loop.  should I be doing something differently here ?

madsbk: Maybe save `lf` globally then it will not go out of scope?

quasiben:
@madsbk this is now passing for me.  I reinstalled UCX from scratch and that resolved the issue.

tests/test_send_recv_two_workers.py
```diff
@@ -0,0 +1,296 @@
...
+import pytest
+import rmm
+import ucp
+from utils import more_than_two_gpus
+
+cmd = "nvidia-smi nvlink --setcontrol 0bz"  # Get output in bytes
+subprocess.check_call(cmd, shell=True)
+
+pynvml = pytest.importorskip("pynvml", reason="PYNVML not installed")
+ITERATIONS = 100
```
quasiben: @pentschev do you want to see this parameterized -- is 100 enough for all objects ?

pentschev:
> @pentschev this is now passing for me. I reinstalled UCX from scratch and that resolved the issue.

Glad to hear that!

tests/test_send_recv_two_workers.py
pentschev: I guess parameterizing it would be better, but it's not absolutely necessary now I'd say. I tried running this for quite a while and this test seems to work, even increasing to 1000 iterations.

tests/send.py
```diff
@@ -0,0 +1,185 @@
+import asyncio
```
quasiben: @pentschev this is the send.py we were playing with earlier today

quasiben:
@pentschev I've added send.py/recv.py to this PR.  Confusingly, things have chnaged and I now see errors after 94 iterations:

```
Bytes Used: 1156186112 94
[1571877360.667247] [dgx15:49526]          cudamem.c:114  UCX  WARN cuMemGetAddressRange(devPtr=0x7f0658000000) failed
python: symbol lookup error: /home/nfs/bzaitlen/miniconda3/envs/cudf_dev10.1/lib/ucx/libucm_cuda.so.0: undefined symbol: ucs_fatal_error_format
==49526== Profiling application: python tests/recv.py
==49526== Profiling result:

```

quasiben:
@madsbk if you execute send.py and recv.py in separate processes you can see similar errors to those in https://github.com/rapidsai/ucx-py/issues/284

tests/recv.py
```diff
@@ -0,0 +1,219 @@
...
+
+pynvml = pytest.importorskip("pynvml", reason="PYNVML not installed")
+ITERATIONS = 1000
+
+
+def cuda_array(size):
+    import cupy as cp
+
+    return cp.empty(size, dtype=cp.uint8)
+    # return rmm.device_array(size, dtype=np.uint8)
```
quasiben: @madsbk in distributed this function normally return rmm/numba.device_array .  When sending/receiving cupy array is this safe ?

madsbk:
One problem is _cyclic references_, which this PR should fix: https://github.com/dask/distributed/pull/3171

jakirkham:
It looks like there is some commented code that is still here. Should it be dropped?

quasiben:
For now I'd like to keep the commented code as we are switching back and forth between cupy and cudf/rmm/numba

benchmarks/recv-into-client.py
```diff
@@ -153,7 +153,8 @@ def serve(port, n_bytes, n_iter, recv, np, verbose, increment):
     expected = np.ones(n_bytes, dtype="u1")
     #            0 or n_iter
     expected *= int(increment) * n_iter
-    np.testing.assert_array_equal(msg, expected)
+    # breakpoint()
+    # np.testing.assert_array_equal(msg, expected)
```
jakirkham: Well, just as an example, I'm guessing code like this was left over from earlier debugging. Should it be dropped? Or is there something I'm missing here?

quasiben: Ah!  Yes, I should clean that up!

tests/debug-tests/recv.py
```diff
@@ -0,0 +1,223 @@
...
+ITERATIONS = 1000
+
+
+def cuda_array(size):
+    # import cupy as cp
+
+    # return numba.cuda.device_array((size,), dtype=np.uint8)
+
+    # return cp.empty(size, dtype=cp.uint8)
+    return rmm.device_array(size, dtype=np.uint8)
```
jakirkham: I wonder if we could use a parameterized fixture to try all 3.

ref: https://docs.pytest.org/en/stable/fixture.html#parametrizing-fixtures

tests/debug-tests/recv.py
```diff
@@ -0,0 +1,223 @@
...
+    # assert deserialized msg is cdf
+    # send receipt
+
+    os.environ.update(env)
+    # ucp.init(
+    #     options={
+    #         "TLS": "tcp,cuda_copy,cuda_ipc,sockcm",
+    #         "SOCKADDR_TLS_PRIORITY": "sockcm",
+    #     }
+    # )
```
jakirkham: Is this still needed?

benchmarks/recv-into-client.py
jakirkham: There are a few other things like this that might be worth checking on real quick. If some of them need to be there, happy to revisit.

tests/debug-tests/recv.py
quasiben: The debug tests of send/recv are the equivalent of test_send_recv_two_workers.py but separated into two processes.  This was needed during debugging to help isolate where cudaFree/cudaMalloc were causing problems.  these test are useful for debugging will probably not run in CI.  Over time, I suspect we can remove them

tests/debug-tests/recv.py
quasiben: cleaned up

benchmarks/recv-into-client.py
quasiben: cleaned up.  Thanks for checking this

PR #238, opened 2019-10-10
Combining communication tags

madsbk:
This PR combines the communication tags into `msg_tag` and `ctrl_tag`, which fixes a problem when a Endpoint connects to itself: https://github.com/rapidsai/ucx-py/pull/215#discussion_r333047909

Also, the logging now uses the address of the underlying `ucp_ep_h` as the Endpoint ID, which is the same as UCX uses thus making it easier to follow the interplay between UCX and UCX-Py


jakirkham:
@MattBBaker, could you please take a look?

MattBBaker:
Looks good. :+1: I'll port my code to this one.

Issue #241, opened 2019-10-11
Include non-GPU conda installation instructions in README


mrocklin:
@jakirkham , are we ready for this?

jakirkham:
Sure. Instead of `ucx-proc=*=gpu` one might write `ucx-proc=*=<proc type>` and explain `<proc type>` is either `cpu` or `gpu`.

mrocklin:
Is it too early to include this in the README?

jakirkham:
I haven't been actively using these packages myself. So I can't really comment on how useable they are or are not. That would be a question for someone else I think. Alternatively if you have a use case for them, maybe you could try them out on that use case and see how well they behave. In either case the feedback would be welcome. 🙂

mrocklin:
Sure.  I'll try things out locally and then report back.

jakirkham:
Thanks for doing that 🙂

pentschev:
I tried the packages, for a while and I didn't notice any difference from the ones I compiled from source, so I think they're fairly safe. Thanks @jakirkham for them! :)

jakirkham:
Were those the CPU ones, the GPU ones, or both?

pentschev:
GPU, I didn't test the CPU at all so far.

mrocklin:
Hrm, I'm getting this

```
mrocklin@carbon:~$ conda create -n ucx -c conda-forge -c jakirkham/label/ucx  ucx-proc=*=cpu ucx ucx-py python=3.7 dask ipython
Collecting package metadata (current_repodata.json): done
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: | 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed                                                                                                             
                                                                                                                   
UnsatisfiableError: The following specifications were found to be incompatible with each other:                    



Package ucx-proc conflicts for:
ucx-proc[build=cpu]
```

mrocklin:
My guess is that while there are cpu-only ucx packages, the ucx-py packages in @jakirkham 's channel somehow need the GPU variants.

quasiben:
@mrocklin can you trry with the channel `jakirkham/label/ucx-new` (notice the `new`)

mrocklin:
That does work.  Thanks @quasiben .  

[...]

PR #249, opened 2019-10-14
Update README

### LINKED ISSUE #248: [DOC] Readme Suggestions
I was trying to build `ucx` using dask and ran into these issues: 

### 1. **Incorrect Dask URL:**
The below url for dask is incorrect i think as it should point to dask instead of rapidsai:

https://github.com/rapidsai/ucx-py/blob/branch-0.10/README.md#dask
```
    git clone git@github.com:rapidsai/dask.git
```

### 2. **http instead of git@**
Also, i am wondering if it will be better to have a `http` urls instead of `git@url` in docs as users wont always have `ssh` keys setup across containers. 

### 3. Dask-cuda conda seems to be pulling previous non ucx-version
Below seems to be pulling previous version of dask-cuda, we may want to add the installation command with with all the dependencies.

Instead of below:
```
conda install -c rapidsai dask-cuda
```

May be something like:
```
conda install -c rapidsai-nightly/label/cuda10.0 -c rapidsai -c nvidia -c numba -c conda-forge dask-cuda
```

### 4. Script to check if ucx is working for a non source installation

This issue may be similar to document UCX usage with dask issue but it will be nice to have test/script i can run which checks if ucx was set up correctly. 

https://github.com/rapidsai/ucx-py/issues/229

quasiben:
PR resolves #248 

@jakirkham  do you have any comments while I'm mucking around in the README ?

cc @VibhuJawa

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,21 +34,21 @@ These three libraries provide a powerful combination of HPC message passing tool
 
 ### dask ###
 
-    git clone git@github.com:rapidsai/dask.git
+    git clone https://github.com/dask/dask.git
     cd dask
     pip install -e .
     cd ..
 
 ### dask distributed ###
 
-    git clone git@github.com:dask/distributed.git
+    git clone https://github.com:dask/distributed.git
     cd distributed
     pip install -e .
     cd ..
 
 ### dask-cuda ###
 
-    conda install -c rapidsai dask-cuda
+    conda install -c rapidsai-nightly dask-cuda
 
 ### conda-forge Dependencies ###
 
@@ -58,15 +58,11 @@ These three libraries provide a powerful combination of HPC message passing tool
 
     git clone https://github.com/openucx/ucx
     cd ucx
-    git remote add Akshay-Venkatesh git@github.com:Akshay-Venkatesh/ucx.git
-    git remote update Akshay-Venkatesh
-    git checkout ucx-cuda
     ./autogen.sh
     mkdir build
     cd build
     ../configure --prefix=$CONDA_PREFIX --enable-debug --with-cuda=$CUDA_HOME --enable-mt --disable-cma CPPFLAGS="-I//$CUDA_HOME/include"
     make -j install
-    cd ../..
 
 ### ucx-py ###
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ These three libraries provide a powerful combination of HPC message passing tool
     ./autogen.sh
     mkdir build
     cd build
-    ../configure --prefix=$CONDA_PREFIX --enable-debug --with-cuda=$CUDA_HOME --enable-mt --disable-cma CPPFLAGS="-I//$CUDA_HOME/include"
+    ../configure --prefix=$CONDA_PREFIX --enable-debug --with-cuda=$CUDA_HOME --enable-mt CPPFLAGS="-I//$CUDA_HOME/include"
     make -j install
 
 ### ucx-py ###


README.md
```diff
@@ -34,21 +34,21 @@ These three libraries provide a powerful combination of HPC message passing tool
...
-    git clone git@github.com:dask/distributed.git
+    git clone https://github.com:dask/distributed.git
     cd distributed
     pip install -e .
     cd ..
 
 ### dask-cuda ###
 
-    conda install -c rapidsai dask-cuda
+    conda install -c rapidsai-nightly dask-cuda
```
VibhuJawa: I think we would have to add other channels here too as this still seems to pulling a version from `19-09-24`
i.e it pulls below version:
```rapidsai-nightly/label/cuda10.0/linux-64::dask-cuda-0.10.0a190924-py37_15```

While below pulls the latest version:

```bash
conda install -c rapidsai-nightly/label/cuda10.0 -c rapidsai -c nvidia -c numba -c conda-forge dask-cuda
```

jakirkham: I don't think we want to use the CUDA labeled channels. Those are getting phased out. Also it locks in a specific CUDA Version, which the user may not want.

My guess is if this was needed there is another issue occurring here like packages are missing or something similar. Could you please raise a separate issue about this? Possibly on dask-cuda?

VibhuJawa: Yup, 
I forgot to remove the `cuda` version. 
Below works too. 

```
 conda install -c rapidsai-nightly -c rapidsai -c nvidia -c numba -c conda-forge dask-cuda --dry
```
Sure, I will raise the issue on  `dask-cuda` .

jakirkham:
Would add the [`compilers`]( https://github.com/conda-forge/compilers-feedstock ) package to the conda-forge dependencies for building from source.

jakirkham:
Also there is another git URL in the ucx-py section that we might want to fix.

README.md
```diff
@@ -58,15 +58,11 @@ These three libraries provide a powerful combination of HPC message passing tool
 
     git clone https://github.com/openucx/ucx
     cd ucx
-    git remote add Akshay-Venkatesh git@github.com:Akshay-Venkatesh/ucx.git
-    git remote update Akshay-Venkatesh
-    git checkout ucx-cuda
     ./autogen.sh
     mkdir build
     cd build
     ../configure --prefix=$CONDA_PREFIX --enable-debug --with-cuda=$CUDA_HOME --enable-mt --disable-cma CPPFLAGS="-I//$CUDA_HOME/include"
```
jakirkham: I think we can drop `--disable-cma` now as that was fixed in `master`. ( https://github.com/openucx/ucx/pull/3424 )

jakirkham:
We may want to [update the `conda install` command for non-development installs]( https://github.com/rapidsai/ucx-py/issues/241#issuecomment-541175247 ), but I'm also happy to handle this independently.

README.md
VibhuJawa: Raised: https://github.com/rapidsai/dask-cuda/issues/159

jakirkham: Thanks @VibhuJawa! 😄

jakirkham:
Have placed my other Conda related suggestions in PR ( https://github.com/rapidsai/ucx-py/pull/253 ). Please take a look and let me know what you think 🙂

README.md
```diff
@@ -58,15 +58,11 @@ These three libraries provide a powerful combination of HPC message passing tool
 
     git clone https://github.com/openucx/ucx
     cd ucx
-    git remote add Akshay-Venkatesh git@github.com:Akshay-Venkatesh/ucx.git
-    git remote update Akshay-Venkatesh
-    git checkout ucx-cuda
```
jakirkham: Updating CI to match in PR ( https://github.com/rapidsai/ucx-py/pull/268 ).

PR #254, opened 2019-10-15
fix/benchmark

quasiben:


pentschev:
A little late, but I'm good with it too. It's been a useful benchmark already for me.

Issue #258, opened 2019-10-15
Drop `--with-prof` option to `build_ext`


jakirkham:
The `--with-prof` option (or `UCX_PY_WITH_PROF` environment variable) define the `UCX_PY_PROF` constant. In the `legacy` branch, this was used to select portions of code in [`ucp/_libs/src/ucp_py_ucp_fxns.h`]( https://github.com/rapidsai/ucx-py/blob/c94a7400cf0df0fdf7a9e48e8a09c8231c194385/ucp/_libs/src/ucp_py_ucp_fxns.h ) and [`ucp/_libs/src/ucp_py_ucp_fxns.c`]( https://github.com/rapidsai/ucx-py/blob/c94a7400cf0df0fdf7a9e48e8a09c8231c194385/ucp/_libs/src/ucp_py_ucp_fxns.c ) for inclusion during compilation. However those files have been dropped in the rewrite. As a result the `UCX_PY_PROF` constant has no effect on compilation. Therefore it makes sense to clean this `--with-prof` flag out as it doesn't do anything any more.

jakirkham:
Addressing with PR ( https://github.com/rapidsai/ucx-py/pull/260 ).

PR #260, opened 2019-10-15
Drop the now irrelevant `--with-prof` flag

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/258

This flag no longer has any effect on the compilation process of ucx-py after the rewrite. So go ahead and drop this flag.


jakirkham:
Handled as part of PR ( https://github.com/rapidsai/ucx-py/pull/263 ).

jakirkham:
Thanks :)

PR #270, opened 2019-10-16
Python getters to the underlying UCP objects

### LINKED ISSUE #117: Provide getters for pointers to underlying ucp objects
It would be very powerful if ucx-py could enable users to initialize and use endpoints in Python, while also being able to make use of the initialized comms underlying the UCP API.

For instance, once endpoints are initialized, it would be useful to be able to extract pointers to objects necessary for sending and receiving messages (eg `ucp_ep_h`, `ucp_context_h`, `ucp_worker_h`, `ucp_listener_h`).

I'll note that I'm still diving into ucx, so I'm still unsure of the actual objects that would need to be exposed. 

The idea is that we want to be able to use Dask to initialize UCP endpoints and send the initialized objects down to C++, but the cuML comms layer would not be aware that any initialization was done in Python or in Dask.

I am hopeful that the cuml comms API could just support the ucp API, but release deadlines might cause this priority to be lowered. At the present time, I do not believe we need to support messages being sent from Python->C++ or vice versa.

### LINKED ISSUE #130: Build Cython extension type for passing UCP endpoint and worker information through Cython at runtime
I wonder if it might make sense to create a wrapper object that allows Python users to pass around the relevant information, but hide the details of pointers and whatever else from Python and only expose this to Cython/C/C++. This is similar to how `memoryview`s behave for instance. If this is a reasonable approach, it might make sense to build a very small [Cython extension type]( https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html ). Thoughts?

_Originally posted by @jakirkham in https://github.com/rapidsai/ucx-py/issues/117#issuecomment-499970672_

madsbk:
Fix #117 and fix #130, including tests

### DIFF
--- a/tests/test_ucx_getters.py
+++ b/tests/test_ucx_getters.py
@@ -1,0 +1,29 @@
+import pytest
+import ucp
+
+
+@pytest.mark.asyncio
+async def test_get_ucp_worker():
+    worker = ucp.get_ucp_worker()
+    assert isinstance(worker, int)
+
+    def server(ep):
+        assert ep.get_ucp_worker() == worker
+
+    lt = ucp.create_listener(server)
+    ep = await ucp.create_endpoint(ucp.get_address(), lt.port)
+    assert ep.get_ucp_worker() == worker
+
+
+@pytest.mark.asyncio
+async def test_get_endpoint():
+    def server(ep):
+        ucp_ep = ep.get_ucp_endpoint()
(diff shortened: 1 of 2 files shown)


quasiben:
@cjnolet does this support the cuML use case you were after ?

jakirkham:
Thanks @madsbk!

@cjnolet, when you get back, we can follow-up on any other changes that might be needed. 🙂

PR #275, opened 2019-10-16
Drop `ucx-setup.sh`

jakirkham:
This appears to be an old script from the early days of developing ucx-py. As it hasn't been kept up-to-date and we are not using it currently, go ahead and drop it.


quasiben:
Thanks @jakirkham

Issue #245, opened 2019-10-12
Why don't we use UCX_TLS=all


mrocklin:
Hypothetically we should be able to set `UCX_TLS=all` and UCX should choose all available transports.  However in practice we don't do this, we set a specific set manually.  I believe that this came out of need, simply setting `all` wasn't giving us the performance that we expected.  It would be useful to start exploring why this is, and see if we can identify issues upstream that would help to address this long-term.

cc @pentschev @Akshay-Venkatesh

pentschev:
I agree, this would be the simplest. IMO more important than `UCX_TLS`, if we didn't need to set the `UCX_NET_DEVICES` variable, things would then be much easier, given that may change depending on system's topology.

MattBBaker:
I can't speak to the earliest days of ucx-py, but I believe part of the motivation was that the TCP TL wasn't working well, so the UCX_TLS was being manually set to work around early TCP UCX bugs. All of that should be fixed now. I'm not sure if UCX_NET_DEVICES is needed either. In previous UCX apps I've never needed to manually set that env variable either. 

What happens if none of the variables are set on the latest release?

jakirkham:
What about `UCX_SOCKADDR_TLS_PRIORITY` or `UCXPY_IFNAME`? Are these still needed? Or can we dispose of them as well?

MattBBaker:
`UCX_SOCKADDR_TLS_PRIORITY` Likely isn't, but I'm not sure how `UCXPY_IFNAME` is meant to be used.

quasiben:
`UCXPY_IFNAME` is an internal env var specific to ucx-py mostly used for our testing CI

jakirkham:
Well thus far I've been unable to run our test suite when setting `UCX_TLS="all"`. Instead it hangs when trying to run the tests in `tests/test_multiple_nodes.py` (none of which complete). This is identical to the behavior I see when leaving `UCX_TLS` unset (guessing that is the same as setting it to `"all"`). Have been using commit ( https://github.com/rapidsai/ucx-py/commit/c3e7a457c1f7990536af07f7a3f917aadd55e082 ) to test.

jakirkham:
[...]

PR #279, opened 2019-10-16
Ensure "all" is a valid TLS value to enable CUDA

pentschev:
This allows setting `UCX_TLS=all` to support CUDA, instead of passing all options individually.

cc @quasiben

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -440,8 +440,8 @@ class _Endpoint:
         self._recv_count = 0
         self._closed = False
         self.pending_msg_list = []
-        # UCX supports CUDA if "cuda" is part of the TLS
-        self._cuda_support = "cuda" in config['TLS']
+        # UCX supports CUDA if "cuda" is part of the TLS or TLS is "all"
+        self._cuda_support = "cuda" in config['TLS'] or config['TLS'] == "all"
 
     @property
     def uid(self):


quasiben:
PR should close
https://github.com/rapidsai/ucx-py/pull/279

jakirkham:
Was looking at this code yesterday and wondering if that was what was causing issues in ( https://github.com/rapidsai/ucx-py/issues/245 ). Thanks for fixing this Peter!

pentschev:
Note that we found a side effect to the use of `UCX_TLS=all`, as described in https://github.com/openucx/ucx/issues/4273#issuecomment-542904715.

jakirkham:
Thanks for the link. Was not aware of that issue. Will take a look.

jakirkham:
Was curious if this would fix [a hang I saw in our test suite with `UCX_TLS=all`]( https://github.com/rapidsai/ucx-py/issues/245#issuecomment-542386313 ). Unfortunately it seem that hang is still there. Not sure if this is a ucx-py issue or UCX issue.

Should add this is not a blocker of this PR AFAIAC. Though it does mean issue ( https://github.com/rapidsai/ucx-py/issues/245 ) should stay open.

quasiben:
@jakirkham I am seeing the same hang.  I think we should merge this but keep https://github.com/rapidsai/ucx-py/issues/245 open as you've said

PR #280, opened 2019-10-17
Non blocking UCX progress mode

madsbk:
This PR implements a `guarantee_msg_order` argument when creating Endpoints, which UCX tag messaging [doesn't guarantees](https://github.com/dask/distributed/pull/3135#issuecomment-545480041).

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -248,6 +248,20 @@ cdef void _listener_callback(ucp_ep_h ep, void *args):
     )
 
 
+async def _non_blocking_mode(weakref_ctx):
+    """This help function maintains a UCX progress loop.
+    Notice, it only keeps a weak reference to `ApplicationContext`, which makes it
+    possible to call `ucp.reset()` even when this loop is running.
+    """
+    while True:
+        ctx = weakref_ctx()
+        if ctx is None:
+            return
+        ctx.progress()
+        del ctx
+        await asyncio.sleep(0)
+
+
 cdef class _Listener:
     """This represents the private part of Listener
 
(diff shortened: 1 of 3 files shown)


pentschev:
Sorry for the naive question, but I'm still mostly clueless about ucx-py's internals, would any changes be necessary for https://github.com/rapidsai/ucx-py/pull/254 to take advantage of this PR?

madsbk:
> Sorry for the naive question, but I'm still mostly clueless about ucx-py's internals, would any changes be necessary for #254 to take advantage of this PR?

Yes, you need to set:
```python
ucp.init(blocking_progress_mode=False)
```

pentschev:
Thanks @madsbk. Just wanted to report it here that I see no difference in bandwidth running https://github.com/rapidsai/ucx-py/pull/254 with or without these changes, not sure whether we really expected changes though.

quasiben:
rerun test

quasiben:
rerun tests

ucp/_libs/core.pyx
```diff
@@ -241,28 +242,41 @@ cdef class ApplicationContext:
...
         cdef ucp_worker_params_t worker_params
         cdef ucs_status_t status
-        self.all_epoll_binded_to_event_loop = set()
+        self.event_loops_binded_for_progress = set()
+        self.progress_tasks = []
         self.config = {}
         self.initiated = False
 
+        if 'UCXPY_NON_BLOCK_MODE' in os.environ:
+            print("UCXPY_NON_BLOCK_MODE")
```
pentschev: Can we try to keep naming more consistent? Everywhere else it's referred to as "blocking", I would expect this to be `UCXPY_NON_BLOCKING_MODE` as well.

madsbk: Good point :)

quasiben:
ok to merge ?

madsbk:
Yes

PR #289, opened 2019-10-21
Fixing UCX leaks and safe UCX shutdown

madsbk:
This PR fixes many resource leaks and makes `ucp.reset()` raise an exception if the `ApplicationContext` is still referenced by `Endpoint` and `Listener` objects. If that is the case, we cannot safely shutdown UCX.

### DIFF
--- a/tests/test_reset.py
+++ b/tests/test_reset.py
@@ -1,0 +1,65 @@
+import pytest
+import ucp
+
+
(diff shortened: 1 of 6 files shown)


quasiben:
Woohoo!  Thanks @madsbk and @jakirkham for tracking this down

PR #293, opened 2019-10-24
Guarantee Message Order

madsbk:
This PR implements a `guarantee_msg_order` argument when creating Endpoints, which UCX tag messaging [doesn't guarantees](https://github.com/dask/distributed/pull/3135#issuecomment-545480041).


quasiben:
Not sure if this PR supercedes PR #280 but both are approved

PR #294, opened 2019-10-24
Add container NVLink section and lint file

jacobtomlinson:
I've added a section to the README about `NVLink and Docker/Kubernetes` to explain how to use NVLink when using containers to isolate/deploy processes.

I've also corrected markdown linting issues and spelling/grammar generally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,103 +1,140 @@
 # Python Bindings for UCX
 
-# Installing preliminary Conda packages
+## Installing preliminary Conda packages
 
-Some preliminary Conda packages can be installed as so. Replace `<CUDA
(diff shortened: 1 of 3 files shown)


quasiben:
This is really great!  Thank you for doing this @jacobtomlinson !  Would you mind adding this to the docs as well https://ucx-py.readthedocs.io/en/latest/ ?  If you're busy, no problem -- we can take care of this later

jacobtomlinson:
Happy to! Do you think it should be there instead or as well as?

quasiben:
I would say instead -- over time we might want to reduce the README in favor of the docs

quasiben:
Thanks @jacobtomlinson

PR #295, opened 2019-10-26
[DRAFT] Add Examples

quasiben:
cc @jakirkham

### DIFF
--- a/examples/cudf-example.py
+++ b/examples/cudf-example.py
@@ -1,0 +1,45 @@
+import time
+
+from dask_cuda import DGX
+from dask_cuda.initialize import initialize
+from distributed import Client
+
+import cudf
+import dask_cudf
+from tornado.ioloop import IOLoop
+
+enable_tcp_over_ucx = True
+enable_infiniband = False
+enable_nvlink = False
+
+
+async def run():
+    initialize(
+        create_cuda_context=True,
+        enable_tcp_over_ucx=enable_tcp_over_ucx,
+        enable_infiniband=enable_infiniband,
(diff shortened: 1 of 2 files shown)


pentschev:
LGTM.

jakirkham:
Would it be possible to run these examples in CI?

pentschev:
Today CI is limited to a single GPU. Even if possible, I wonder about the extent of usefulness of this for now.

jakirkham:
The example? It's useful for people starting out with this package. It can be a nice verification that things got installed/configured correctly. Or a nice jumping off point for more complex code that uses this pattern.

The main concern I have with excluding this example from CI testing is that it is likely to go stale given the rapid rate of change in the library.

quasiben:
We could run this as part of CI but as @pentschev said we are currently limited to one GPU.   @jakirkham is that something you could take a look at ?  Possibly ping on this issue https://github.com/rapidsai/ops/issues/504 ?

pentschev:
My comment may have been a bit misleading, what I mean is that currently it's probably possible to run the example on CI, but given it's limited to a single GPU, I don't think that would be too useful for now (before we have multi-GPU support).

jakirkham:
Certainly it would be beneficial to have the ideal environment to test on, but there is even value in just running the code in a non-ideal environment. It will help us catch functions that have been changed or dropped and forces us to handle these changes in the example somehow.

jakirkham:
Raised the CI testing point as issue ( https://github.com/rapidsai/ucx-py/issues/308 ).

jakirkham:
Thanks @quasiben!

Issue #299, opened 2019-10-28
Claiming PyPI name


jakirkham:
It would be good to claim the PyPI name to avoid other people grabbing it and sticking something up there in place of this module. That said, [the `ucp` name is already taken on PyPI]( https://pypi.org/project/ucp/ ). So it may be worth renaming our package to `ucx-py` to get a name no one has taken. We may additionally want to rename our modules to avoid conflicts with the `ucp` package (should users install both).

mrocklin:
How about ucx?

mrocklin:
I realized that I could check that easily.  UCX is free if we want it.

mrocklin:
There are reasons to prefer ucp, I understand, just throwing it out there.

jakirkham:
> How about ucx?

This becomes problematic when packaging in conda-forge as that name is already used for the C library. So we still need to name it `ucx-python`, `python-ucx`, or similar.

jakirkham:
Should add while this doesn't seem like a big deal it has caused a lot of headache with `graphviz` (just as an example). 🙂

jakirkham:
Plus we are already using ucx-py in the repo name and the Conda packages. Probably easier to just lean into it. 😉

quasiben:
How do you feel about scikit-learn/sklearn ?  That would be another model for what ucx-py/ucp is, no ?

jakirkham:
Yeah I guess the question is do we think people we use our package and this other `ucp` at the same time. If so, we would want to avoid them clobbering each other. If not, then yeah no need to complicate things. 🙂

madsbk:
I vote for `ucx` or `ucxpy`. 
Personally, I don't like `ucp` -- I know that the `p` stands for _protocol_, which makes sense if you are familiar with the whole UCX software stack but for the average user it is just confusing.

But no matter what name we decide, we should use the same name in this repo, `distributed`, `cudf`, and `cuml`. (e.g. it is called `ucx` in `distributed`)

pentschev:
I agree that `ucx` may be strange because of the UCX C library, that may be one day a conda package or something of that sort. I dislike `ucp` just like @madsbk. IMHO something like `ucxpy` would be a reasonable compromise on both sides.

quasiben:
I like ucxpy as it make discussing this effort a bit easier even though it may not be technically as correct as `ucp`.  I also don't expect there to be a pypi release but @jakirkham noted people can do malicious things on pypi by name squatting a package.

[...]

PR #307, opened 2019-10-31
Rename package to ucxpy

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/299

This renames the package to `ucxpy` like the PyPI name that we claimed.


quasiben:
cc @cjnolet to keep you and cuML folks aware

PR #309, opened 2019-11-01
Cleanup the closing of endpoint

madsbk:
Instead of `close()` and `async signal_shutdown()`, we now have:

```python
    def abort(self):
        """Close the communication immediately and abruptly.
        Useful in destructors or generators' ``finally`` blocks.

        Notice, this functions doesn't signal the connected peer to close.
        To do that, use `Endpoint.close()`
        """

    async def close(self):
        """Close the endpoint cleanly.
        This will attempt to flush outgoing buffers before actually
        closing the underlying UCX endpoint.
        """
```

### DIFF
--- a/benchmarks/recv-into-client.py
+++ b/benchmarks/recv-into-client.py
@@ -118,8 +118,7 @@ def serve(port, n_bytes, n_iter, recv, np, verbose, increment):
                 print("\n")
                 print(df)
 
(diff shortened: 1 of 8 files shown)


quasiben:
Very cool!  Seems like a much cleaner API

quasiben:
Thanks for updating docs as well!

PR #312, opened 2019-11-04
Clean Benchmarks

quasiben:
This PR moves the non-working/older benchmarks into a separate directory .  I haven't looked in detail at all the benchmarks but though we should keep them around for a little bit longer.  Perhaps as we start benchmarking in earnest we can remove both `old_test` and `old_benchmarks`

Thoughts ?


pentschev:
Sounds good to me. If we think they may not be too useful in the future, we could just create a new branch or a tag to find them if needed and delete them from the main branch.

quasiben:
@jakirkham when you have a minute can you review ?

jakirkham:
Yeah I agree with Peter. Why not just delete them if they are not relevant? We can always recover them with Git if we were wrong. 🙂

jakirkham:
rerun tests

jakirkham:
Thanks Ben! 😄

Issue #290, opened 2019-10-21
Segmentation fault when running simple cuDF example


jakirkham:
I tried running the following simple example with cuDF. However was unable to proceed as it encountered a segmentation fault. Figured I'd raise here initially ( given its similarity in construction to the example in issue, https://github.com/rapidsai/ucx-py/issues/284 ). Though it is possible this issue traces back to something else.

<details>
<summary>Example:</summary>

```python
import time

import cudf
import cupy

import dask_cudf

from distributed import Client
from dask_cuda import DGX
from dask_cuda.initialize import initialize
from tornado.ioloop import IOLoop


enable_tcp_over_ucx=True
enable_infiniband=False
enable_nvlink=False


async def run():
    initialize(create_cuda_context=True,
               enable_tcp_over_ucx=enable_tcp_over_ucx,
               enable_infiniband=enable_infiniband,
               enable_nvlink=enable_nvlink)

    async with DGX(interface="enp1s0f0",
                   protocol="ucx",
                   enable_tcp_over_ucx=enable_tcp_over_ucx,
                   enable_infiniband=enable_infiniband,
                   enable_nvlink=enable_nvlink,
                   asynchronous=True) as dgx:
        async with Client(dgx, asynchronous=True) as client:
            d = dask_cudf.from_cudf(
                cudf.DataFrame([("a", range(10))]),
                npartitions=2
            )
            r = d.sum()

            for i in range(100):
                print("Running iteration:", i)
                start = time.time()
                res = await client.compute(r)
                print("Time for iteration", i, ":", time.time() - start)


if __name__ == "__main__":
    loop = IOLoop.current()
    loop.run_sync(run)
```

</details>

<details>
<summary>Backtrace:</summary>

```c
Running iteration: 0
[dgx15:39190:0:39190] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x8)
==== backtrace ====
    0  /home/nfs/jkirkham/miniconda/envs/ucx/lib/libucs.so.0(+0x22601) [0x7f04949ba601]
    1  /home/nfs/jkirkham/miniconda/envs/ucx/lib/libucs.so.0(+0x227d2) [0x7f04949ba7d2]
    2  /lib/x86_64-linux-gnu/libpthread.so.0(+0x12890) [0x7f04c60cd890]
[...]

PR #314, opened 2019-11-05
fix hanging tests

quasiben:
This PR fixes a hang between `test_send_recv.py` and `test_send_recv_two_workers.py`.  By using `spawn` in multiprocessing things now work, however, there is an annoying warning in the two worker test:

```
Error in sys.excepthook:

Original exception was:
```

Not sure why this is occurring but everything passes.  cc @rjzamora


quasiben:
While fixing the hang I also cleaned up some other tests

tests/debug-tests/recv.py
```diff
@@ -119,16 +134,16 @@ def client(env, port, func):
...
+    # import cupy as cp
 
-    if isinstance(rx_cuda_obj, cp.ndarray):
-        cp.testing.assert_allclose(rx_cuda_obj, pure_cuda_obj)
-    else:
-        assert_eq(rx_cuda_obj, pure_cuda_obj)
+    # if isinstance(rx_cuda_obj, cp.ndarray):
+    #     cp.testing.assert_allclose(rx_cuda_obj, pure_cuda_obj)
+    # else:
+    #     assert_eq(rx_cuda_obj, pure_cuda_obj)
```
jakirkham: What's the reason for commenting this code?

quasiben:
in the middle of debugging a memory leak

quasiben:
This PR is ready to go. I'd like to keep working on send/recv  debug-tests in a separate PR if that's ok.  Things are functional for the debug tests but I would like to continue refining elsewhere

tests/debug-tests/recv.py
```diff
@@ -68,35 +48,14 @@ def client(env, port, func):
...
-                            frames.append(b"")
-
-            msg = await from_frames(frames)
+            frames, msg = await recv(ep)
+
+            # msg = msg["data"]
+            # # msg = msg.T
+            # msg = {"data": to_serialize(msg)}
+            # frames = await to_frames(msg, serializers=("cuda", "dask", "pickle"))
+            # Send meta data
```
jakirkham: Is this still needed?

tests/debug-tests/recv.py
jakirkham: Any thoughts @quasiben? 🙂

tests/test_send_recv_two_workers.py
```diff
@@ -17,7 +17,7 @@
 # subprocess.check_call(cmd, shell=True)
 
 pynvml = pytest.importorskip("pynvml", reason="PYNVML not installed")
-ITERATIONS = 100
+ITERATIONS = 1
```
jakirkham: Should we also do `from utils import ITERATIONS` here?

tests/debug-tests/utils.py
```diff
@@ -0,0 +1,87 @@
+import argparse
+
+from distributed.comm.utils import from_frames
+from distributed.utils import nbytes, parse_bytes
+
+import numpy as np
+import ucp
+
+ITERATIONS = 50
```
jakirkham: Should this also be `from utils import ITERATIONS`?

tests/debug-tests/recv.py
quasiben: ooops.... thanks!

tests/test_send_recv_two_workers.py
```diff
@@ -43,7 +43,10 @@ def client(env, port, func):
     # assert deserialized msg is cdf
     # send receipt
 
+    ucp.reset()
     os.environ.update(env)
+    ucp.init()
```
jakirkham: Does using `reset()` and `init()` now work? Had [some issues]( https://github.com/rapidsai/ucx-py/issues/290#issuecomment-544996019 ) with this previously.

tests/test_send_recv_two_workers.py
```diff
@@ -126,8 +129,9 @@ def server(env, port, func):
     # create listener receiver
     # write cudf object
     # confirm message is sent correctly
-
+    ucp.reset()
     os.environ.update(env)
+    ucp.init()
```
jakirkham: Same question as above.

tests/test_send_recv_two_workers.py
quasiben: for now, no.  I want to keep those test separated since test_send_recv_two_workers.py will eventually run in CI.  I'm also going to add CLI args to the debug test

jakirkham:
Thanks for the cleanup Ben! Had a few questions above. 🙂

tests/debug-tests/utils.py
quasiben: same as above

tests/test_send_recv_two_workers.py
quasiben: yes, they should be working now.  I think this is only necessary when running all tests together -- pytest, i think does something strange here

quasiben:
@jakirkham can you give this another look ?

tests/test_send_recv_two_workers.py
jakirkham: Ok cool. Honestly wasn't sure. So thanks for the clarification. 🙂

jakirkham:
Thanks @quasiben! 😄

quasiben:
Thanks @jakirkham !

PR #319, opened 2019-11-11
Doc Update - Add NumPy Example

quasiben:
cc @jakirkham

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -5,8 +5,19 @@ Quickstart
 Setup
 -----
 
+With GPU support:
+
+::
+
     conda create -n ucx -c conda-forge -c conda-forge/label/rc_ucx \
-    cudatoolkit=<CUDA version> ucx-proc=*=<processor type> ucx ucx-py python=3.7
+      cudatoolkit=<CUDA version> ucx-proc=*=gpu ucx ucx-py python=3.7
+
+Without GPU support:
+
+::
+
+    conda create -n ucx -c conda-forge -c conda-forge/label/rc_ucx \
+      ucx-proc=*=cpu ucx ucx-py python=3.7
 
 For a more detailed guide on installation options please refer to the :doc:`install` page.
 
@@ -19,6 +30,7 @@ Process 1 - Server
 .. code-block:: python
 
     import asyncio
+    import time
     import ucp
     import numpy as np
 
@@ -28,10 +40,10 @@ Process 1 - Server
 
     async def send(ep):
         # recv buffer
-        arr = np.empty(n_bytes, dtype=np.uint8)
+        arr = np.empty(n_bytes, dtype='u1')
         await ep.recv(arr)
-        assert np.count_nonzero(arr) == 0
-        print("Received NumPy array")
+        assert np.count_nonzero(arr) == np.array(0, dtype=np.int64)
+        print("Received CuPy array")
 
         # increment array and send back
         arr += 1
@@ -41,11 +53,15 @@ Process 1 - Server
         await ep.close()
         lf.close()
 
-    lf = ucp.create_listener(send, port)
-    while not lf.closed():
-        await asyncio.sleep(0.1)
+    async def main():
+        global lf
+        lf = ucp.create_listener(send, port)
 
+        while not lf.closed():
+            await asyncio.sleep(0.1)
 
+    if __name__ == '__main__':
+        asyncio.run(main())
 
 Process 2 - Client
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


quasiben:
@pentschev if you have a minute can you review ?

docs/source/quickstart.rst
```diff
@@ -10,9 +10,75 @@ Setup
...
+~~~~~~~~~
+
+.. code-block:: python
+
+    import asyncio
+    import ucp
+    import numpy as np
+
+    n_bytes = 2**30
+    host = ucp.get_address(ifname='enp0s25')
```
pentschev: Maybe clarify that this should be the local ethernet interface? It's unlikely that this will be the same for others, changing this to `eth0` probably makes more sense, as that is a more common name.

docs/source/quickstart.rst
```diff
@@ -10,9 +10,75 @@ Setup
 
 For a more detailed guide on installation options please refer to the :doc:`install` page.
 
+Send/Recv NumPy Arrays
+---------------------
+
+Process 1
```
pentschev:
```suggestion
Process 1 - Server
```

docs/source/quickstart.rst
```diff
@@ -10,9 +10,75 @@ Setup
...
+        await ep.close()
+        lf.close()
+
+    lf = ucp.create_listener(send, port)
+    while not lf.closed():
+        await asyncio.sleep(0.1)
+
+
+
+Process 2
```
pentschev:
```suggestion
Process 2 - Client
```

docs/source/quickstart.rst
```diff
@@ -10,9 +10,75 @@ Setup
 
 For a more detailed guide on installation options please refer to the :doc:`install` page.
 
+Send/Recv NumPy Arrays
+---------------------
+
+Process 1
+~~~~~~~~~
```
pentschev:
```suggestion
~~~~~~~~~~~~~~~~~~
```

docs/source/quickstart.rst
```diff
@@ -10,9 +10,75 @@ Setup
...
+        lf.close()
+
+    lf = ucp.create_listener(send, port)
+    while not lf.closed():
+        await asyncio.sleep(0.1)
+
+
+
+Process 2
+~~~~~~~~~
```
pentschev:
```suggestion
~~~~~~~~~~~~~~~~~~
```

docs/source/quickstart.rst
```diff
@@ -10,9 +10,75 @@ Setup
...
+
+    n_bytes = 2**30
+    host = ucp.get_address(ifname='enp0s25')
+    port = 13337
+
+    async def send(ep):
+        # recv buffer
+        arr = np.empty(n_bytes, dtype=np.uint8)
+        await ep.recv(arr)
+        print("Received NumPy array")
```
pentschev: Maybe assert the array here?

docs/source/quickstart.rst
```diff
@@ -10,9 +10,75 @@ Setup
...
+    resp = np.empty_like(msg)
+    await ep.recv(resp, msg_size)  # receive the echo
+    np.testing.assert_array_equal(msg + 1, resp)
+
+
 Send/Recv CuPy Arrays
 ---------------------
 
+.. note::
+    If you are passing CuPy arrays between GPUs, ensure you have correctly set ``UCX_TLS`` with ``cuda_ipc``. See the :doc:`configuration` for more details
```
pentschev: What happens if you don't?

quasiben: Good point.  I changed the language to reflect that the user *wants* to use nvlink

quasiben:
@pentschev any additional thoughts ?

docs/source/quickstart.rst
```diff
@@ -23,13 +90,14 @@ Process 1
     import cupy as cp
 
     n_bytes = 2**30
-    host = ucp.get_address(ifname='enp1s0f0')
+    host = ucp.get_address(ifname='enp1s0f0')  # ethernet device name
```
jakirkham: Why is it `eth0` for NumPy and `enp1s0f0` for CuPy? Could they use the same `ifname` or is there a reason for them to be different?

docs/source/quickstart.rst
```diff
@@ -10,11 +10,78 @@ Setup
...
+
+    import asyncio
+    import ucp
+    import numpy as np
+
+    port = 13337
+    n_bytes = 2**30
+    host = ucp.get_address(ifname='eth0')  # ethernet device name
+    ep = await ucp.create_endpoint(host, port)
+    msg = np.zeros(n_bytes, dtype='u1') # create some data to send
```
pentschev: Can we be consistent? Here `dtype='u1'`, in the first process it's `dtype=np.uint8`, I would suggest changing one of them to match the other.

pentschev:
@quasiben probably a stupid question, but have you tested both processes and checked that they both run and exit successfully? That can be an issue if someone is attaching tools like `nvprof`.

quasiben:
@pentschev things should be consistent now -- i also changed the style of the example so they could be consumed as files more easily.  And yes, they exit cleanly

PR #322, opened 2019-11-12
Split installation documentation to CPU/GPU sections

mrocklin:
It's complex regardless, but this way we stop people from
installing cudatoolkit, and give them a copy-pasteable solution in
the CPU-only case.


quasiben:
thanks for doing this @mrocklin !  I'm going to push some changes to get CI moving along

PR #325, opened 2019-11-12
Document Cython requirement

jakirkham:
This adds our Cython requirement to our README. Hopefully helps out users installing this from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ cd ..
 #### conda-forge Dependencies
 
 ```bash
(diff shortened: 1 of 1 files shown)


quasiben:
What do you think of adding that env creation line to docs page.  Doing this will also trigger a build and will not require us to have an admin merge changes in

jakirkham:
I thought about it, but would like to avoid duplicating the same content. Should hopefully avoid some of it going stale and contradicting other content.

That said, I don't have any strong thoughts on where this content should live, but perhaps we can save that discussion for another issue. 🙂

jakirkham:
Thanks Ray! 😄

PR #328, opened 2019-11-15
Ci check dask-distributed

madsbk:


quasiben:
rerun tests

quasiben:
@madsbk is this something you are still working on ?

madsbk:
I'm not sure if this is a good idea. What do you think, do we want to test dask/ucx in our CI?

quasiben:
I am not opposed to testing dask-ucx in this CI -- dask is currently one of the primary users of UCX.  However, there are now ucx/ucx-py conda packages which would make testing feasible for dask. 
 @mrocklin do you have an opinion here ?

In general there is a fair amount of overlap between dask/ucx/dask-cuda and it's not clear to me where CI testing should happen.  However, in these earlier stages I would vote for more duplicative testing

pentschev:
I have been thinking about similar situations for dask-cuda, but I haven't reached a conclusion yet. There are things from dask core and distributed I would like to test (e.g. CUDA-related code). It seems that this kind of issue is now spanning multiple projects, perhaps we should add such tests into dask-cuda instead? Alternatively, I have been thinking of maybe another repo to test only the things that projects can't test by themselves.

mrocklin:
I think that testing the relevant components of downstream packages (as this does) seems sensible to me.  Having ucx-py test `distributed/comm/tests/test_ucx.py` sounds good, expecially because the Dask CI generally *doesn't* test that code because of the UCX dependency.  Hopefully once UCX-Py stabilizes we'll be able to do more testing downstream in Dask, and this relationship can flip.

quasiben:
@mads I think the CI issue will go away if you upgrade the cudf/dask-cudf version to 0.11:

https://github.com/rapidsai/ucx-py/blob/5b764d194e7c510e92d48fadb921e2f85273054e/ci/gpu/build.sh#L50

0.11/0.12 has a critical fix for the RMM

quasiben:
Happy to push the fix if you are busy

madsbk:
@quasiben please do :)

quasiben:
Not sure what happened but i was restricted from pushing to your branch -- maybe branch restrictions were added: https://help.github.com/en/github/administering-a-repository/enabling-branch-restrictions

  For now I created https://github.com/rapidsai/ucx-py/pull/341 to move this along without bothering you further

PR #332, opened 2019-11-20
Set UCX_CUDA_IPC_CACHE=n upon ucp import

pentschev:
Set `UCX_CUDA_IPC_CACHE=n` upon ucp import to prevent memory leaks when NVLink is enabled (see https://github.com/openucx/ucx/issues/4410).

Using `ucp.init` to do this is currently not possible as it only supports UCP configurations and `UCX_CUDA_IPC_CACHE` is part of UCT, which is considerably more complex to query/set, as it requires traversing all MDs and interfaces.

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -20,6 +20,13 @@ if os.environ.get("UCX_MEMTYPE_CACHE", "") != "n":
     logger.debug("Setting env UCX_MEMTYPE_CACHE=n, which is required by UCX")
     os.environ["UCX_MEMTYPE_CACHE"] = "n"
 
+if os.environ.get("UCX_CUDA_IPC_CACHE", "") != "n":
+    # See <https://github.com/openucx/ucx/issues/4410>
+    logger.debug(
+        "Setting env UCX_CUDA_IPC_CACHE=n, which is required to avoid NVLink memory leaks"
+    )
+    os.environ["UCX_CUDA_IPC_CACHE"] = "n"
+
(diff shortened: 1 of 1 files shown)


quasiben:
Thanks @pentschev

quasiben:
@madsbk are you ok with this change ?

ucp/__init__.py
```diff
@@ -20,6 +20,14 @@
     logger.debug("Setting env UCX_MEMTYPE_CACHE=n, which is required by UCX")
     os.environ["UCX_MEMTYPE_CACHE"] = "n"
 
+if os.environ.get("UCX_CUDA_IPC_CACHE", "") != "n":
```
madsbk: I think we should make it _overwriteable_, such that we only sets `UCX_CUDA_IPC_CACHE` if it is not already set.
E.g. 
```python
if "UCX_CUDA_IPC_CACHE" not in os.environ:
    ...
``` 
Similarly, let's also make `UCX_MEMTYPE_CACHE`_overwriteable_:
```python
if "UCX_MEMTYPE_CACHE" not in os.environ:
    ...
```
It is very useful to be able to compare the performance impact of the cache options.

quasiben: Sounds good.  I'll take care of this -- Peter is on PTO

quasiben:
rerun tests

quasiben:
Thanks @madsbk !

pentschev:
Makes sense, thanks for reviews and updating the PR while I was away.

PR #333, opened 2019-11-20
add join benchmark

quasiben:
This PR adds a benchmark tests designed to be executed with pytest on a DGX1

> py.test -v "benchmarks/dask-join.py::test_join"

Sample output

```
Join benchmark
NVLINK: True
-------------------
n_bytes  | 713.89 MB

===================
91.74 MB / s
===================


Join benchmark
NVLINK: False
-------------------
n_bytes  | 262.96 MB

===================
26.29 MB / s
===================
```

cc @mrocklin

### DIFF
--- a/benchmarks/dask-join.py
+++ b/benchmarks/dask-join.py
@@ -1,6 +1,7 @@
 import asyncio
 import time
 
+import dask
 import dask.dataframe as dd
 from dask_cuda import DGX
 from dask_cuda.initialize import initialize
@@ -8,6 +9,7 @@ from distributed import Client
 from distributed.utils import format_bytes
 
 import cudf
+import cupy
 import pytest
 
 enable_tcp_over_ucx = True
@@ -15,7 +17,7 @@ enable_infiniband = False
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("enable_nvlink", [True, False])
+@pytest.mark.parametrize("enable_nvlink", [True])
 async def test_join(enable_nvlink):
     initialize(
         create_cuda_context=True,
@@ -33,26 +35,35 @@ async def test_join(enable_nvlink):
         asynchronous=True,
     ) as dgx:
         async with Client(dgx, asynchronous=True) as client:
-            await client.run(cudf.set_allocator, "default", pool=True,
-                    initial_pool_size=int(2.4e10))
+            await client.run(
+                cudf.set_allocator, "default", pool=True, initial_pool_size=int(24e9)
+            )
 
             n_rows = 1_000_000_000
             n_partitions = 100
             n_keys = 500_000_000
 
             def make_partition(n_rows, n_keys, name):
-                return cudf.DataFrame({
-                    name: cupy.random.random(n_rows),
-                    "id": cupy.random.randint(0, n_keys, size=n_rows),
-                })
+                return cudf.DataFrame(
+                    {
+                        name: cupy.random.random(n_rows),
+                        "id": cupy.random.randint(0, n_keys, size=n_rows),
+                    }
+                )
 
             left = dd.from_delayed(
-                [dask.delayed(make_partition)(n_rows // n_partitions, n_keys, "x") for _ in range(n_partitions)],
-                meta=make_partition(1, n_keys, "x")
+                [
+                    dask.delayed(make_partition)(n_rows // n_partitions, n_keys, "x")
+                    for _ in range(n_partitions)
+                ],
+                meta=make_partition(1, n_keys, "x"),
             )
             right = dd.from_delayed(
-                [dask.delayed(make_partition)(n_rows // n_partitions, n_keys, "y") for _ in range(n_partitions)],
-                meta=make_partition(1, n_keys, "y")
+                [
+                    dask.delayed(make_partition)(n_rows // n_partitions, n_keys, "y")
+                    for _ in range(n_partitions)
+                ],
+                meta=make_partition(1, n_keys, "y"),
             )
 
             left, right = dask.persist(left, right)
@@ -60,20 +71,14 @@ async def test_join(enable_nvlink):
             await right
 
             start = time.time()
-            out = await gleft.merge(gright, on=["id"]).persist()
+            out = await left.merge(right, on=["id"]).persist()
             stop = time.time()
 
             duration = stop - start
 
-            # bandwidth_workers = dgx.scheduler.bandwidth_workers
-            # bandwidth_workers_total = sum([d for w, d in bandwidth_workers.items()])
             bandwidth = dgx.scheduler.bandwidth
+            print("NVLink:", enable_nvlink, "Rows:", n_rows, "Bandwidth:", format_bytes(bandwidth))
 
-            with open("benchmarks.txt", "a+") as f:
-                f.write("Join benchmark\n")
-                f.write(f"NVLINK: {enable_nvlink}\n")
-                f.write("-------------------\n")
-                f.write(f"Aggregate Bandwidth  | {format_bytes(bandwidth)}\n")
-                f.write("\n===================\n")
-                f.write(f"{format_bytes(total / took)} / s\n")
-                f.write("===================\n\n\n")
+            _ = await client.profile(server=True, filename="join-communication.html")
+
+
--- a/benchmarks/dask-join.py
+++ b/benchmarks/dask-join.py
@@ -1,0 +1,91 @@
+import asyncio
+import time
+from time import perf_counter as clock
+
+import dask.array as da
+import dask.dataframe as dd
+from dask_cuda import DGX
+from dask_cuda.initialize import initialize
+from distributed import Client
+from distributed.utils import format_bytes
+from distributed.utils_test import captured_logger
+
+import cudf
+import dask_cudf
+import pytest
+
+enable_tcp_over_ucx = True
+enable_infiniband = False
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("enable_nvlink", [True, False])
+async def test_join(enable_nvlink):
+    initialize(
+        create_cuda_context=True,
+        enable_tcp_over_ucx=enable_tcp_over_ucx,
+        enable_infiniband=enable_infiniband,
+        enable_nvlink=enable_nvlink,
+    )
+
+    async with DGX(
+        interface="enp1s0f0",
+        protocol="ucx",
+        enable_tcp_over_ucx=enable_tcp_over_ucx,
+        enable_infiniband=enable_infiniband,
+        enable_nvlink=enable_nvlink,
+        asynchronous=True,
+    ) as dgx:
+        async with Client(dgx, asynchronous=True) as client:
+
+            n_rows = 1_000_000_000
+            n_keys = 5_000_000
+
+            left = dd.concat(
+                [
+                    da.random.random(n_rows).to_dask_dataframe(columns="x"),
+                    da.random.randint(0, n_keys, size=n_rows).to_dask_dataframe(
+                        columns="id"
+                    ),
+                ],
+                axis=1,
+            ).persist()
+            n_rows = 10_000_000
+
+            right = dd.concat(
+                [
+                    da.random.random(n_rows).to_dask_dataframe(columns="y"),
+                    da.random.randint(0, n_keys, size=n_rows).to_dask_dataframe(
+                        columns="id"
+                    ),
+                ],
+                axis=1,
+            ).persist()
+
+            gleft = left.map_partitions(cudf.from_pandas)
+            gright = right.map_partitions(cudf.from_pandas)
+            gleft = await gleft.persist()
+            gright = await gright.persist()  # persist data in device memory
+
+            start = clock()
+            out = gleft.merge(gright, on=["id"])  # this is lazy
+            out = await out.persist()
+            stop = clock()
+
+            took = stop - start
+
+            async def f(dask_scheduler):
+                return dask_scheduler.bandwidth_workers
+
+            data = await client.run_on_scheduler(f)
+            total = sum([d for w, d in data.items()])
+            total_str = format_bytes(total)
+
+            with open("benchmarks.txt", "a+") as f:
+                f.write("Join benchmark\n")
+                f.write(f"NVLINK: {enable_nvlink}\n")
+                f.write("-------------------\n")
+                f.write(f"n_bytes  | {format_bytes(total)}\n")
+                f.write("\n===================\n")
+                f.write(f"{format_bytes(total / took)} / s\n")
+                f.write("===================\n\n\n")
--- a/benchmarks/dask-join.py
+++ b/benchmarks/dask-join.py
@@ -1,6 +1,50 @@
-import asyncio
-import time
+"""
+Conda Environment
+-----------------
 
+# environment.yml
+name: ucx-benchmark
+channels:
+  - conda-forge
+  - conda-forge/label/rc_ucx
+  - rapidsai-nightly
+  - nvidia
+dependencies:
+  - cudatoolkit
+  - ucx-proc=*=gpu
+  - ucx
+  - ucx-py
+  - dask-cudf=0.11
+  - dask-cuda=0.11
+  - pytest
+  - pytest-asyncio
+  - python=3.7
+  - numba=0.46
+  - pip
+  - pip:
+    - git+https://github.com/dask/distributed@0b68318112b13d70a9cdd741e5db00da2ec6a8f5
+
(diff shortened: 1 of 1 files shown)


benchmarks/dask-join.py
```diff
@@ -0,0 +1,91 @@
...
+            out = await out.persist()
+            stop = clock()
+
+            took = stop - start
+
+            async def f(dask_scheduler):
+                return dask_scheduler.bandwidth_workers
+
+            data = await client.run_on_scheduler(f)
+            total = sum([d for w, d in data.items()])
```
quasiben: @mrocklin is this the total amount of transferred data ?

quasiben:
I'm confused as to why with `NVLINK=True` the benchmark records `713.89 MB` and with `NVLINK=False` the benchmark records `262.96 MB`

benchmarks/dask-join.py
```diff
@@ -0,0 +1,91 @@
...
+
+            start = clock()
+            out = gleft.merge(gright, on=["id"])  # this is lazy
+            out = await out.persist()
+            stop = clock()
+
+            took = stop - start
+
+            async def f(dask_scheduler):
+                return dask_scheduler.bandwidth_workers
```
mrocklin: You can access the scheduler directly with `cluster.scheduler`

benchmarks/dask-join.py
mrocklin: No, this is only bandwidth.  If you want transfers you'll have to look at the workers.  The state is here: 

https://github.com/dask/distributed/blob/35551998d7350cd5ae6a5c24970d8437fd8d521d/distributed/worker.py#L1925-L1938

We can probably start accumulating some metrics around that in the scheduler though.  I plan to focus a bit on this sort of thing next week.

mrocklin: The `cluster.scheduler.bandwidth` number might also be interesting.  This aggregates the bandwidth over all workers over all time.

quasiben:
Thanks for the review @mrocklin.  I updated the calls to access `bandwidth` -- perhaps I should wait until next week to continue with this work

madsbk:
> I'm confused as to why with `NVLINK=True` the benchmark records `713.89 MB` and with `NVLINK=False` the benchmark records `262.96 MB`

I think it is because you are joining on random values. My guess is that the communication amount changes at each run?

quasiben:
@madsbk that's a good point.  Perhaps setting a seed should guarantee the same random number for each run

mrocklin:
@quasiben I'm intersted in running this.  Any suggestions on setting up an environment?

quasiben:
try the following:

```
conda create -n ucx-benchmark -c conda-forge -c conda-forge/label/rc_ucx -c rapidsai-nightly -c nvidia   \
  cudatoolkit=<CUDA version> ucx-proc=*=gpu ucx ucx-py rapids=0.11 python=3.7 
```

mrocklin:
Do I need any development branches or environment variables to get things to an ideal state?

quasiben:
This *should* be fine for ucx/ucx-py you will want latest of dask/dask-cuda/distributed.  If you are playing with cupy you will need master.  No environment variables needed as dask-cuda should be handling all of this now

mrocklin:
OK, I also had to add pytest-asyncio, and ucx-py master (due to the recent rmm.DeviceBuffer change in distributed.  Things kind of work now, but they seem to stall out halfway through the merge, I think on the `shuffle-join` tasks.  Are you sure that you don't have some other environment variables set somewhere?

mrocklin:
Hrm, it looks like one of the workers is stalled out churning on something

image

quasiben:
I think this error is related to the recent distributed PR:
https://github.com/dask/distributed/pull/3261

```
distributed.utils - ERROR - int() argument must be a string, a bytes-like object or a number, not 'cython_function_or_method'
  File "/datasets/bzaitlen/GitRepos/distributed/distributed/utils.py", line 663, in log_errors
    yield
  File "/datasets/bzaitlen/GitRepos/distributed/distributed/comm/ucx.py", line 188, in read
    await self.ep.recv(frame)
  File "/datasets/bzaitlen/GitRepos/ucx-py/ucp/public_api.py", line 282, in recv
    await self._ep.recv(buffer, nbytes=nbytes)
  File "ucp/_libs/core.pyx", line 673, in recv
    nbytes = get_buffer_nbytes(buffer, check_min_size=nbytes,
  File "ucp/_libs/utils.pyx", line 75, in ucp._libs.utils.get_buffer_nbytes
    shape = [int(s) for s in iface['shape']]
TypeError: int() argument must be a string, a bytes-like object or a number, not 'cython_function_or_method'
distributed.worker - ERROR - int() argument must be a string, a bytes-like object or a number, not 'cython_function_or_method'
```

We can either fix this issue or revert the commit into distributed

quasiben:
xref https://github.com/rapidsai/rmm/issues/183 for fixing this issue

mrocklin:
Even when I walk back from the commit in dask/distributed, things still stall for me :(

jakirkham:
I think the code in that distributed PR is ok. It’s using the keyword argument, which appears to be the problem in the RMM issue.

As to the issue mentioned in the OP over there, my guess is PR ( https://github.com/rapidsai/rmm/pull/182 ) fixes the issue. We knew this was coming around the corner. We just hadn’t formed a consensus on the solution yet. That has now happened and a fix is in. Things should be fine once a new nightly is built. Though let us know if not. 🙂

quasiben:
@mrocklin rmm is fixed.  again, sorry for the issues here.  If you remove the directory and start again things should work now.  I tested with the following:

> conda create -n ucx-benchmark -c conda-forge -c conda-forge/label/rc_ucx -c rapidsai-nightly -c nvidia     cudatoolkit=10.1 ucx-proc=*=gpu ucx ucx-py rapids=0.11 pytest pytest-asyncio python=3.7

Then installed dask/distributed from master and tests passed:
> py.test dask-join.py -s -v

*Note: this creates a file: benchmarks.txt*

mrocklin:
I'm running the following, and things still stall for me:

```python
import asyncio
import time
from time import perf_counter as clock

import dask
import dask.array as da
import dask.dataframe as dd
from dask_cuda import DGX
from dask_cuda.initialize import initialize
from distributed import Client, wait
from distributed.utils import format_bytes

import cupy
import cudf
import dask_cudf
import pytest

enable_tcp_over_ucx = True
enable_infiniband = False
enable_nvlink = False

initialize(
    create_cuda_context=True,
    enable_tcp_over_ucx=enable_tcp_over_ucx,
    enable_infiniband=enable_infiniband,
    enable_nvlink=enable_nvlink,
)

cluster = DGX(
    interface="enp1s0f0",
    protocol="ucx",
    enable_tcp_over_ucx=enable_tcp_over_ucx,
    enable_infiniband=enable_infiniband,
    enable_nvlink=enable_nvlink,
    silence_logs=False,
    dashboard_address=":8789"
)

client = Client(cluster)
client

n_rows = 50_000_000
n_partitions = 100
n_keys = 10_000_000

def make_partition(n_rows, n_keys, name):
    return cudf.DataFrame({
        name: cupy.random.random(n_rows),
        "id": cupy.random.randint(0, n_keys, size=n_rows),
    })

left = dd.from_delayed(
    [dask.delayed(make_partition)(n_rows // n_partitions, n_keys, "x") for _ in range(n_partitions)],
    meta=make_partition(1, n_keys, "x")
)
right = dd.from_delayed(
    [dask.delayed(make_partition)(n_rows // n_partitions, n_keys, "y") for _ in range(n_partitions)],
    meta=make_partition(1, n_keys, "y")
)

left, right = dask.persist(left, right)


out = left.merge(right, on=["id"])
out = out.persist()
_ = wait(out)
```

pentschev:
I can confirm the same behavior locally with @mrocklin's code. However, for me the hang is actually caused due to a crash:

```python-traceback
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/comm/core.py", line 222, in connect
    _raise(error)
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/comm/core.py", line 205, in _raise
    raise IOError(msg)
OSError: Timed out trying to connect to 'ucx://10.33.227.161:39829' after 10 s: connect() didn't finish in time

During handling of the above exception, another exception occurred:

  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/worker.py", line 1908, in gather_dep
    self.rpc, deps, worker, who=self.address
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/worker.py", line 3126, in get_data_from_worker
    comm = await rpc.connect(worker)
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/core.py", line 880, in connect
    connection_args=self.connection_args,
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/comm/core.py", line 231, in connect
    _raise(error)
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/comm/core.py", line 205, in _raise
    raise IOError(msg)
OSError: Timed out trying to connect to 'ucx://10.33.227.161:39829' after 10 s: Timed out trying to connect to 'ucx://10.33.227.161:39829' after 10 s: connect() didn't finish in time
distributed.worker - INFO - Can't find dependencies for key ('shuffle-join-shuffle-group-19dfb72dc73ef3ea1c9e1b3fc2a6b10c', 2, (0, 7))
distributed.worker - ERROR - Worker stream died during communication: ucx://10.33.227.161:52101
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/comm/core.py", line 222, in connect
    _raise(error)
  File "/datasets/pentschev/miniconda3/envs/ucx-benchmark/lib/python3.7/site-packages/distributed/comm/core.py", line 205, in _raise
    raise IOError(msg)
OSError: Timed out trying to connect to 'ucx://10.33.227.161:52101' after 10 s: connect() didn't finish in time

...
```

quasiben:
I have a high degree of confidence that the issue is due to numba versions.  I think, rapids (the metapackge) is not compatible with numba=0.46.  instead, the following should work:

> conda create -n ucx-benchmark-2 -c conda-forge -c conda-forge/label/rc_ucx -c rapidsai-nightly -c nvidia cudatoolkit=10.1 ucx-proc=*=gpu ucx ucx-py dask-cudf=0.11 pytest pytest-asyncio python=3.7 numba=0.46

mrocklin:
OK, I've pushed up a modified version that has somewhat cleaner data creation.  I've also included a line to collect a profile of the workers' administrative threads.  This requires dask/distributed master today.

At the top of the file I've included a conda environment file that should work (as of right now, it's not tightly pinned) and a Dask config file, which should help a bit with the profiling.

mrocklin:
@quasiben and I went through this, and we learned a few things:

1.  Oddly, turning on infiniband made things very slow
2.  RMM was helpful in performance, decreasing total runtime about a factor of two
3.  There is an odd bottleneck in asyncio, I'll take a look at this.  But also cc'ing @madsbk , who I suspect may be interseted.
4.  We spend a large amount of time deserializing cudf index objects.  cc @kkraus14 

There is a profile of the workers' administrative threads (which handle serialization and communication) here: 

https://matthewrocklin.com/raw-host/dgx-cudf-ucx-join-comm-profile.html

There is a conda environment and dask config file in the top of the benchmark file for reproducibility.

mrocklin:
OK, I wasn't able to resolve the `is_closed` issue, but I did learn some things.

-  It is definitely being called frequently by something (judging by the activity from a print statement)
-  ucx-py does call it somewhat frequently here, but commenting out that code didn't seem to improve performance.

<details>

```diff
diff --git a/ucp/_libs/send_recv.pyx b/ucp/_libs/send_recv.pyx
index 576a64f..3ed7652 100644
--- a/ucp/_libs/send_recv.pyx
+++ b/ucp/_libs/send_recv.pyx
@@ -60,9 +60,9 @@ cdef void _send_callback(void *request, ucs_status_t status):
         return
     cdef object future = <object> req.future
     cdef object log_str = <object> req.log_str
-    if asyncio.get_event_loop().is_closed():
-        pass
-    elif status == UCS_ERR_CANCELED:
+    # if asyncio.get_event_loop().is_closed():
+    #     pass
+    if status == UCS_ERR_CANCELED:
         future.set_exception(UCXCanceled())
     elif status != UCS_OK:
         msg = "Error sending%s " %(" \"%s\":" % log_str if log_str else ":")
@@ -100,9 +100,9 @@ cdef void _tag_recv_callback(void *request, ucs_status_t status,
     cdef object future = <object> req.future
     cdef object log_str = <object> req.log_str
     msg = "Error receiving%s " %(" \"%s\":" % log_str if log_str else ":")
-    if asyncio.get_event_loop().is_closed():
-        pass
-    elif status == UCS_ERR_CANCELED:
+    # if asyncio.get_event_loop().is_closed():
+    #     pass
+    if status == UCS_ERR_CANCELED:
         future.set_exception(UCXCanceled())
     elif status != UCS_OK:
         msg += (<object> ucs_status_string(status)).decode("utf-8")
@@ -158,9 +158,9 @@ cdef void _stream_recv_callback(void *request, ucs_status_t status,
     cdef object future = <object> req.future
     cdef object log_str = <object> req.log_str
     msg = "Error receiving %s" %(" \"%s\":" % log_str if log_str else ":")
-    if asyncio.get_event_loop().is_closed():
-        pass
-    elif status == UCS_ERR_CANCELED:
+    # if asyncio.get_event_loop().is_closed():
+    #     pass
+    if status == UCS_ERR_CANCELED:
         future.set_exception(UCXCanceled())
     elif status != UCS_OK:
         msg += (<object> ucs_status_string(status)).decode("utf-8")
```

</details>

We're not getting much of a trace about where this is coming from from the profile.  It *looks* like someone is calling this as a new coroutine (although I don't know why this would happen).  Any help investigating this would be welcome.

kkraus14:
> We spend a large amount of time deserializing cudf index objects. cc @kkraus14

Do you have a profile by any chance? I'll take a peek at the serialization / deserialization code but this should be very similar to sending any other object like Series / DataFrame.

mrocklin:
Yes, if you're comfortable with a Dask profilie, then you want to look here, in particular on the right side.

https://matthewrocklin.com/raw-host/dgx-cudf-ucx-join-comm-profile.html

kkraus14:
> Yes, if you're comfortable with a Dask profilie, then you want to look here, in particular on the right side.
> 
> https://matthewrocklin.com/raw-host/dgx-cudf-ucx-join-comm-profile.html

https://github.com/rapidsai/cudf/blob/branch-0.11/python/cudf/cudf/core/index.py#L995

That line is doing a deep copy where we should be doing a shallow copy and is the reason for the slowdown.

kkraus14:
@quasiben @madsbk https://github.com/rapidsai/cudf/blob/branch-0.11/python/cudf/cudf/core/buffer.py#L112 is another source of slowdown as it looks like we're currently doing a device to device copy for every buffer we receive from distributed/ucx-py, is this still needed or is ucx-py using RMM where we can share the memory nicely now?

quasiben:
@kkraus14 ucx-py does allocate any buffers at the moment.  Instead, it operates on buffers handed to it.  In dask we generate an RMM buffer and hand it to UCX to populate with received data:

https://github.com/dask/distributed/blob/0b68318112b13d70a9cdd741e5db00da2ec6a8f5/distributed/comm/ucx.py#L46-L52

So cudf deserialization *should* always be operating on an RMM buffer and the copy shouldn't be necessary.

kkraus14:
So frames will be a list of `rmm.DeviceBuffer` objects once we're all said and done? We should be able to consume those directly and reference count them to control lifetime nicely once the C++ and Python refactors are done 😄.

jakirkham:
That sounds right. Does that work for you?

kkraus14:
> That sounds right. Does that work for you?

Yup. With the Python refactor we'll be able to construct `Buffer` objects from those zero-copy and then construct `Column` objects from those zero-copy.

jakirkham:
Fantastic! 😀

madsbk:
@quasiben do you want to merge this or should we refer to https://github.com/rapidsai/dask-cuda/blob/branch-0.11/dask_cuda/benchmarks/local_cudf_merge.py?

quasiben:
@madsbk I think you're right.  The work here has now been encapsulated `local_cudf_merge.py` .  @mrocklin thank you for the help here!

PR #342, opened 2019-11-29
Add TopologicalDistance class to identify closest devices to a GPU

pentschev:
This will help applications such as dask-cuda identify the closest InfiniBand device to a GPU.

Note that I only added one test and that requires either to be running on a DGX-1 or DGX-2. Since these changes are very much hardware dependent, I don't know if we could have a test to run on generic hardware. Ideas for a test like that are welcome.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -56,7 +56,7 @@ conda install "cudatoolkit=$CUDA_REL" \
 conda install -c conda-forge "async_generator" "automake" "libtool" \
                               "cmake" "automake" "autoconf" "cython" \
                               "pytest" "pkg-config" "pytest-asyncio" \
-                              "pynvml"
+                              "pynvml" "libhwloc"
 
 # install ucx from conda-forge rc channel
 # conda install -c conda-forge/label/rc_ucx "ucx-proc=*=gpu" "ucx"
--- a/conda/environments/builddocs_py37.yml
+++ b/conda/environments/builddocs_py37.yml
@@ -15,6 +15,7 @@ dependencies:
 - recommonmark
 - pandoc=<2.0.0
 - pip
+- libhwloc
 - ucx
 - ucx-py
 - pip:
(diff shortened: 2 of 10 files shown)


jakirkham:
Would it be possible to detect the proximity of other things like drives?

pentschev:
> Would it be possible to detect the proximity of other things like drives?

It is now. :)

jakirkham:
Interesting. What would one need to do to find the closest drive?

pentschev:
> Interesting. What would one need to do to find the closest drive?

https://github.com/rapidsai/ucx-py/pull/342/files#diff-000ee5071a36008fa30a492232e3c274R134-R135

This is how you can find the distance of GPU to block devices. In that case, both have a distance of `4`, so from a performance perspective, they're the same. The value of `4` is the number of hops from the common ancestor of both devices (GPU and HDD in that case) to the GPU.

jakirkham:
I wonder if we can use this to constrain spilling or other IO operations to the drive closest to the device.

pentschev:
I don't see why not. This would need additional work on identifying where each device is mount and if they're not in RAID, in which case we would not be able to specify the data path exactly.

jakirkham:
> I don't see why not. This would need additional work on identifying where each device is mount and if they're not in RAID, in which case we would not be able to specify the data path exactly.

Should we open an issue? I expect this will be needed in the medium term.

pentschev:
Feel free to open an issue if you think that would be helpful. I'm not entirely sure there will be a very reliable solution, given that it depends on the machine's configuration, how permissions are handled, etc., but we can discuss that in the issue.

quasiben:
Apologies -- my conda env blew up.  Looks great!  Tested and passes

mike-wendt:
FWIW @quasiben @jakirkham this broke the feedstock due to a new dependecy `libhwloc` I'll add it to [rapdisai/ucx-split-feedstock](https://github.com/rapidsai/ucx-split-feedstock/tree/enh-gpuci) on branch `enh-gpuci` to see if this fixes the build [failure](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/conda/job/ucx-py-build/51/console)

quasiben:
Thanks @mike-wendt for checking in -- we have been chatting about this on slack

PR #345, opened 2019-12-04
Setting UCX_SOCKADDR_TLS_PRIORITY=all as default

madsbk:
As far as I can see, we always have to set `UCX_SOCKADDR_TLS_PRIORITY=sockcm` in order to run on multiple nodes. And since there isn't any drawback?, let's set it as default.

Notice, I am setting the option as an environment variable. It is possible to set it as a UCX option but I think this is more readable.

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -28,6 +28,12 @@ if "UCX_CUDA_IPC_CACHE" not in os.environ:
     )
     os.environ["UCX_CUDA_IPC_CACHE"] = "n"
 
+if "UCX_SOCKADDR_TLS_PRIORITY" not in os.environ:
+    logger.debug(
+        "Setting env UCX_SOCKADDR_TLS_PRIORITY=sockcm, which is required to connect multiple nodes"
+    )
+    os.environ["UCX_SOCKADDR_TLS_PRIORITY"] = "sockcm"
+
 # Set the root logger before importing modules that use it
 _level_enum = logging.getLevelName(os.getenv("UCXPY_LOG_LEVEL", "WARNING"))
 logging.basicConfig(level=_level_enum, format="%(levelname)s %(message)s")
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -30,10 +30,10 @@ if "UCX_CUDA_IPC_CACHE" not in os.environ:
 
 if "UCX_SOCKADDR_TLS_PRIORITY" not in os.environ:
     logger.debug(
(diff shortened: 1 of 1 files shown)


madsbk:
rerun tests

pentschev:
I thought the default `UCX_SOCKADDR_TLS_PRIORITY=all` would cover this. For me it seems to be working with that.

pentschev:
I have no strong opinions on whether we should set `UCX_SOCKADDR_TLS_PRIORITY=sockcm`, but I just wanted to report that I just tested TCP over UCX, NVLink and InfiniBand without this and it works, even when I remove that from https://github.com/rapidsai/dask-cuda/blob/branch-0.11/dask_cuda/initialize.py#L67.

madsbk:
Interesting, I never tried setting `UCX_SOCKADDR_TLS_PRIORITY=all` but it also works for me. 
However, if I set nothing it defaults to `UCX_SOCKADDR_TLS_PRIORITY=rdmacm,*`, which doesn't work for me. I get the following error:
```
[1575460425.147620] [dgx12:1028 :0]   rdmacm_iface.c:409  UCX  ERROR received event RDMA_CM_EVENT_UNREACHABLE. status = 2. Peer: 127.0.0.1:35809.
[1575460426.138844] [dgx12:1028 :0]   rdmacm_iface.c:409  UCX  ERROR received event RDMA_CM_EVENT_UNREACHABLE. status = 2. Peer: 127.0.0.1:35809.
[1575460426.894397] [dgx12:1175 :0]      wireup_ep.c:486  UCX  ERROR no supported sockaddr auxiliary transports found for mlx5_0:1
[1575460426.911305] [dgx12:1173 :0]   rdmacm_iface.c:409  UCX  ERROR received event RDMA_CM_EVENT_UNREACHABLE. status = 2. Peer: 127.0.0.1:35809.
[1575460427.140295] [dgx12:1028 :0]   rdmacm_iface.c:409  UCX  ERROR received event RDMA_CM_EVENT_UNREACHABLE. status = 2. Peer: 127.0.0.1:35809.
```

I am change it to `UCX_SOCKADDR_TLS_PRIORITY=all`

pentschev:
Wait a second, I just checked and the default is also `UCX_SOCKADDR_TLS_PRIORITY=rdmacm,*` for me. I could have sworn I had seen the deafult was `UCX_SOCKADDR_TLS_PRIORITY=all`, maybe I'm mistaken or the default has changed. Either way, I now have a strong opinion we should indeed have this PR in. :)

pentschev:
rerun tests

pentschev:
Thanks @madsbk !

PR #347, opened 2019-12-04
add conda env in sourc install instructions

quasiben:
cc @beckernick

PR #351, opened 2019-12-05
Move install instructions out of README

jakirkham:
As the install instructions are mostly in the docs anyways, this fills in some missing details in the docs and removes them from the README in the spirit of DRY. Should keep things easier to maintain. Discussion on this would be very welcome.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -56,7 +56,8 @@ RAPIDS Dependencies
 
 ::
 
-    conda install -c rapidsai-nightly -c nvidia -c conda-forge dask-cuda rmm
+    conda install -n ucx -c rapidsai-nightly -c nvidia -c conda-forge \
+        dask-cuda rmm
 
 UCX
 ~~~
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -54,7 +54,7 @@ Test Dependencies
 
 ::
 
-    conda install -n ucx -c rapidsai-nightly -c nvidia -c conda-forge \
+    conda install -n ucx -c rapidsai -c nvidia -c conda-forge \
         pytest pytest-asyncio \
         cupy numba>=0.46 rmm \
         distributed
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -6,7 +6,9 @@ Prerequisites
 
 UCX depends on the following system libraries being present: ``libibverbs``,
 ``librdmacm``, and ``libnuma`` (``numactl`` on Enterprise Linux).  Please
-install these with your Linux system's package manager.
+install these with your Linux system's package manager. When building from
+source you will also need the ``*-dev`` (``*-devel`` on Enterprise Linux)
+packages as well.
 
 Conda
 -----
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -77,7 +77,7 @@ UCX
     ../contrib/configure-devel --prefix=$CONDA_PREFIX --with-cuda=$CUDA_HOME --enable-mt CPPFLAGS="-I/$CUDA_HOME/include"
     make -j install
 
-UCX-PY
+UCX-Py
 ~~~~~~
 
 ::
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -44,7 +44,7 @@ conda-forge Dependencies
 
 ::
 
-    conda create -n ucx-foo -c conda-forge \
+    conda create -n ucx -c conda-forge \
         automake make cmake libtool pkg-config \
         libhwloc \
         python=3.7 setuptools cython cupy distributed numba>=0.46 \
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -43,6 +43,13 @@ conda-forge Dependencies
 
     conda install -c conda-forge automake make cmake libtool pkg-config pytest-asyncio cython cupy distributed setuptools
 
+dask-cuda
+~~~~~~~~~
+
+::
+
+    conda install -c rapidsai-nightly -c nvidia -c conda-forge dask-cuda
+
 UCX
 ~~~
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -38,7 +38,8 @@ The following instructions assume you'll be using ucx-py on a CUDA enabled syste
 
 
 
-1) Install UCX
+UCX
+~~~
 
 ::
 
@@ -53,7 +54,8 @@ The following instructions assume you'll be using ucx-py on a CUDA enabled syste
     ../contrib/configure-devel --prefix=$CONDA_PREFIX --with-cuda=$CUDA_HOME --enable-mt CPPFLAGS="-I/$CUDA_HOME/include"
     make -j install
 
-2) Install UCX-PY
+UCX-PY
+~~~~~~
 
 ::
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -50,12 +50,12 @@ conda-forge Dependencies
         pytest pytest-asyncio \
         ipython
 
-dask-cuda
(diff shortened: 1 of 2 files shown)


docs/source/install.rst
```diff
@@ -1,9 +1,21 @@
 Install
 =======
 
+Prerequisites
+-------------
+
+UCX depends on the following system libraries being present: ``libibverbs``,
```
cjnolet: With the topological distance feature that was added to ucx-py, I noticed I also needed to install the `libhwloc` package. Is that already included in one of these packages?

jakirkham: Yeah @quasiben is doing that in PR ( https://github.com/rapidsai/ucx-py/pull/347 ). Will rebase after that lands.

jakirkham: Have merged in Ben's branch. `libhwloc` is a dependency of the `ucx-py` package. So should get pulled in automatically when installing that. We also have it listed in the build from source section below.

docs/source/install.rst
```diff
@@ -1,9 +1,22 @@
...
+install these with your Linux system's package manager. When building from
+source you will also need the ``*-dev`` (``*-devel`` on Enterprise Linux)
+packages as well.
+
 Conda
 -----
 
+Some preliminary Conda packages can be installed as so. Replace
+``<CUDA version>`` with either ``9.2``, ``10.0``, or ``10.1``. These are
+available both on ``rapidsai`` and ``rapidsai-nightly``.
```
jakirkham: @randerzander, does this address your concern?

docs/source/install.rst
```diff
@@ -25,28 +38,34 @@ Source
...
-    UCX depends on the following system libraries being present: ``libibverbs``, ``librdmacm``, ``librdmacm``,
-    and ``libnuma`` (numactl on Enterprise Linux).  Please install these with your Linux system's package manager.
-    Additionally, we provide an example conda environment which installs necessary dependencies for building and testing ucx/ucx-py
 
+conda-forge Dependencies
+~~~~~~~~~~~~~~~~~~~~~~~~
 
+::
 
+    conda create -n ucx-foo -c conda-forge \
```
quasiben: conda create -n ucx -c conda-forge

docs/source/install.rst
```diff
@@ -25,28 +38,34 @@ Source
...
+        automake make cmake libtool pkg-config \
+        libhwloc \
+        python=3.7 setuptools cython cupy distributed numba>=0.46 \
+        pytest pytest-asyncio \
+        ipython
 
-1) Create Conda Env
+RAPIDS Dependencies
```
quasiben: What do you think of not including RAPIDs dependencies in the install instructions.  We could mention this is the quickstart or examples but RAPIDS isn't necessary for installing ucx-py

docs/source/install.rst
quasiben: This was my fault -- sorry

docs/source/install.rst
jakirkham: Well [RMM is used in our tests]( https://github.com/rapidsai/ucx-py/blob/e15bb7ce858bc3eb33fc29c3250d2e10e3b1402d/tests/test_send_recv_two_workers.py#L12 ). How should we track our test dependencies?

Though this does raise a good point. There are things like `cmake` and `ipython` that are not strictly dependencies either. Should they be dropped?

docs/source/install.rst
jakirkham: np. Should be fixed now :)

docs/source/install.rst
pentschev: I would say we drop everything that isn't really a hard dependency. If tests require additional libraries like RMM, then we can either say that or skip such tests when libraries aren't available.

docs/source/install.rst
```diff
@@ -56,10 +77,12 @@ The following instructions assume you'll be using ucx-py on a CUDA enabled syste
     ../contrib/configure-devel --prefix=$CONDA_PREFIX --with-cuda=$CUDA_HOME --enable-mt CPPFLAGS="-I/$CUDA_HOME/include"
     make -j install
 
-3) Install UCX-PY
+UCX-PY
```
pentschev: This is just a question (or nitpicking, depending how you look at it), but what's the right way of writing this name? I've been using `UCX-Py`, is `UCX-PY` the right one?

jakirkham: Fair question. Not sure. Happy to change to `UCX-Py` if that makes more sense.

docs/source/install.rst
jakirkham: How about we break out the test dependencies? This way users and packagers can still see what is needed to run the tests.  🙂

As to the tests, I'll save that for another PR. Seems out-of-scope for a doc update 😉

pentschev: > How about we break out the test dependencies? This way users and packagers can still see what is needed to run the tests. 🙂

Seems good to me.

> As to skipping the tests, I'll save that for another PR. Seems out-of-scope for a doc update 😉

Absolutely, I didn't mean to suggest doing it here and now, just as the general solution that would be nice.

docs/source/install.rst
pentschev: I don't mind it either way, just asking because I'd like to know for the long-term. The reason I've been writing `UCX-Py` is to projects like NumPy and CuPy (and yes, these are the correct ways to write them).

docs/source/install.rst
```diff
@@ -1,9 +1,22 @@
 Install
 =======
 
+Prerequisites
+-------------
+
+UCX depends on the following system libraries being present: ``libibverbs``,
+``librdmacm``, and ``libnuma`` (``numactl`` on Enterprise Linux).  Please
```
pentschev: I think we're still missing `libibcm` here, or does one of the other libraries include that?

docs/source/install.rst
jakirkham: Yeah I went ahead with your suggestion. Agree it seems reasonable 👍

docs/source/install.rst
jakirkham: Let's do that in another pass. We don't have builds with it yet.

docs/source/install.rst
jakirkham: Ok, I've given a go at this. PTAL 🙂

Agreed. It would be a good thing to do. It seems some tests use `pytest.importorskip`, but it's a bit inconsistent. Raised issue ( https://github.com/rapidsai/ucx-py/issues/352 ) to track this.

quasiben: I like the move to test dependencies.  Thanks @jakirkham

docs/source/install.rst
pentschev: If we don't have builds with it yet, does that mean we won't support InfiniBand for now? That is absolutely necessary for InfiniBand and if someone tries to `enable_infiniband=True` in dask-cuda, that will fail.

jakirkham: I wouldn't say "won't". It's that we "don't" have this functionality currently. We can of course update the doc when that changes.

Understood, but I think this is a different discussion.

quasiben: When the CDT is merged in we'll add the `libibcm` to the docs

quasiben: Hopefully, this happens either tomorrow or early next week

pentschev: Ok, cool. I just wanted to make sure we wouldn't miss this by accident. Thanks for clarifying.

jakirkham: Documenting in PR ( https://github.com/rapidsai/ucx-py/pull/360 ). Seems like we should have the `libibcm` dependency resolved soonish.

pentschev: Awesome! Thanks @jakirkham .

PR #356, opened 2019-12-06
[Critical] Setting UCX_SOCKADDR_TLS_PRIORITY=sockcm

madsbk:
It turns out that `UCX_SOCKADDR_TLS_PRIORITY=all` doesn't work in the general case. We need to set `UCX_SOCKADDR_TLS_PRIORITY=sockcm`.

Since [`LocalCUDACluster`](https://github.com/rapidsai/dask-cuda/blob/97f485d1f449b9d6a45d79a86971d604d458cfa2/dask_cuda/local_cuda_cluster.py#L64) doesn't sets it by default and `DGX` is deprecated, it will fail for a new user.

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -30,10 +30,10 @@ if "UCX_CUDA_IPC_CACHE" not in os.environ:
 
 if "UCX_SOCKADDR_TLS_PRIORITY" not in os.environ:
     logger.debug(
-        "Setting env UCX_SOCKADDR_TLS_PRIORITY=all, "
+        "Setting env UCX_SOCKADDR_TLS_PRIORITY=sockcm, "
         "which is required to connect multiple nodes"
     )
-    os.environ["UCX_SOCKADDR_TLS_PRIORITY"] = "all"
+    os.environ["UCX_SOCKADDR_TLS_PRIORITY"] = "sockcm"
 
 # Set the root logger before importing modules that use it
 _level_enum = logging.getLevelName(os.getenv("UCXPY_LOG_LEVEL", "WARNING"))


pentschev:
Actually, `LocalCUDACluster` does set that after https://github.com/rapidsai/dask-cuda/pull/205. But I think this may be important for people who may use UCX-Py directly (without dask-cuda).

Could we please get this in 0.11 still @raydouglass ?

madsbk:
But only if `enable_tcp_over_ucx or enable_infiniband or enable_nvlink`
right?

pentschev:
Right, but users have to enable at least one of them. I'm not sure if we should make `enable_tcp_over_ucx=True` always if `protocol="ucx"`, do you think we should?

madsbk:
> Right, but users have to enable at least one of them. I'm not sure if we should make `enable_tcp_over_ucx=True` always if `protocol="ucx"`, do you think we should?

No, I think that we should be able to use `LocalCUDACluster` and manually specify all options through env.

pentschev:
Yeah, this is what I think now too. Perhaps in the long run we should consider making `enable_tcp_over_ucx=True` as default, as of today, I'm not sure if dask-cuda is usable with `protocol="ucx"` but `enable_tcp_over_ucx=False`.

madsbk:
In the [cudf benchmark](https://github.com/rapidsai/dask-cuda/blob/97f485d1f449b9d6a45d79a86971d604d458cfa2/dask_cuda/benchmarks/local_cudf_merge.py#L129) we have:
```python

    cluster = LocalCUDACluster(
        protocol=args.protocol, n_workers=args.n_workers, CUDA_VISIBLE_DEVICES=args.devs
    )
    client = Client(cluster)
```
which works fine for me?

pentschev:
Ah, alright. That's my mistake. If you don't set anything UCX's defaults will be used, such as `UCX_TLS=all` which should work in most cases indeed. I've seen some cases where not setting `enable_tcp_over_ucx=True` doesn't seem to work well for me, I'll have to investigate that further. Anyway, none of that should hold this PR back.

raydouglass:
rerun tests

PR #368, opened 2019-12-11
Use __dealloc__ for TopologicalDistance Cython code

### LINKED ISSUE #365: Avoiding `__del__` in `TopologicalDistance`
Noticed that we are using [`__del__` in the `TopologicalDistance`]( https://github.com/rapidsai/ucx-py/blob/eba981f3e486d4d7305eba5e54b13c11c499e303/ucp/_libs/topological_distance.pyx#L28-L30 ). However this isn't supported for extension types (snippet from the docs below).

> **Note:** There is no `__del__()` method for extension types.

ref: https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#finalization-method-dealloc

pentschev:
Fix #365

### DIFF
--- a/ucp/_libs/topological_distance.pyx
+++ b/ucp/_libs/topological_distance.pyx
@@ -25,7 +25,7 @@ cdef class TopologicalDistance:
         """
         self.topo = <hwloc_topology_t> initialize_topology()
 
-    def __del__(self):
+    def __dealloc__(self):
         cdef hwloc_topology_t topo = <hwloc_topology_t> self.topo
         hwloc_topology_destroy(topo)
 


pentschev:
@jakirkham should address #365, let me know if there are any concerns.

jakirkham:
Looks good. Do we know what the CI failure is?

pentschev:
rerun tests

pentschev:
> Looks good. Do we know what the CI failure is?

Not sure, but I've seen similar errors to this in dask-cuda recently. I think they're related to a timeout, perhaps it's too short for that test.

pentschev:
There have been some errors that were only reproducible when running all tests, choosing such tests individually would not cause the tests to fail.

pentschev:
Has CI been failing recently or are those errors just showing up because I'm lucky? Both runs failed with different failures and both unrelated to this PR.

quasiben:
I don't know why this is failing either.  I've tried with an updated env and the tests pass locally.  i'll rerun again and see what happens

quasiben:
rerun tests

jakirkham:
Thanks @pentschev and @quasiben! 😄

PR #371, opened 2019-12-16
Create and run local network

madsbk:
This PR implements `utils.run_on_local_network()`, which spawns, connects, and runs a number of workers.

```python

def run_on_local_network(
    n_workers, worker_func, worker_args=None, server_address=None, ucx_options_list=None
):
    """
    Creates a local UCX network of `n_workers` that runs `worker_func`

    Parameters
    ----------
    n_workers : int
        Number of workers (nodes) in the network.
    worker_func: callable (can be a coroutine)
        Function that each worker execute. 
        Must have signature: `worker(rank, eps, args)` where
            - rank is the worker id
            - eps is a dict of ranks to ucx endpoints
            - args given here as `worker_args`
    worker_args: object
        The argument to pass to `worker_func`.
    server_address: str
        Server address for the workers. If None, get_address() is used.
    ucx_options_list: list of dict
        Options to pass to UCX when initializing workers, one for each worker.

    Returns
    -------
    dev_names : str
        Names of the closest net devices
    """
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,9 +12,7 @@ import versioneer
 from setuptools import setup
 from setuptools.extension import Extension
 
-include_dirs = [os.path.dirname(get_python_inc())] + [
-    "/home/mkristensen/anaconda3/envs/rap-nightly/include"
-]
+include_dirs = [os.path.dirname(get_python_inc())]
 library_dirs = [get_config_var("LIBDIR")]
 libraries = ["ucp", "uct", "ucm", "ucs", "hwloc"]
 extra_compile_args = ["-std=c99"]
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,9 @@ import versioneer
 from setuptools import setup
 from setuptools.extension import Extension
 
-include_dirs = [os.path.dirname(get_python_inc())]
+include_dirs = [os.path.dirname(get_python_inc())] + [
+    "/home/mkristensen/anaconda3/envs/rap-nightly/include"
+]
 library_dirs = [get_config_var("LIBDIR")]
 libraries = ["ucp", "uct", "ucm", "ucs", "hwloc"]
 extra_compile_args = ["-std=c99"]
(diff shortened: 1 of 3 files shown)


ucp/utils.py
```diff
@@ -39,3 +45,133 @@ def get_address(ifname=None):
             s.fileno(), 0x8915, struct.pack("256s", ifname[:15])  # SIOCGIFADDR
         )[20:24]
     )
+
+
+def get_closest_net_devices(gpu_dev):
```
quasiben: This is not related to this PR but do you think we could/should start using this over `UCXPY_IFNAME` ?

setup.py
```diff
@@ -12,7 +12,9 @@
 from setuptools import setup
 from setuptools.extension import Extension
 
-include_dirs = [os.path.dirname(get_python_inc())]
+include_dirs = [os.path.dirname(get_python_inc())] + [
```
quasiben: Is this leftover from debugging sessions ?

quasiben:
This looks good to me -- left a few small comments.  I imagine this will be particularly useful for MPI style testing of ucx-py, correct ?

setup.py
jakirkham: Assuming you are meaning the line below. Yeah this should already be the same as `os.path.dirname(get_python_inc())` unless I'm missing something.

madsbk: Yes, this was included by mistake :)

ucp/utils.py
madsbk: Yes, I was hoping we could begin using it. 
But it properly best if it is possible to overwrite it using `UCXPY_IFNAME`.

madsbk:
rerun test

madsbk:
rerun tests

ucp/utils.py
quasiben: Makes sense

PR #374, opened 2019-12-17
add glossary page and update more TLS config options

quasiben:
PR updates TLS options and adds a glossary page for the many acronyms in the UCX/networking ecosystem

Thanks to @hiroyuki-sato for the supplying the [glossary](https://github.com/hiroyuki-sato/openucx-docs/)

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -69,11 +69,28 @@ Values:
 
 Transport Methods (Simplified):
 
-- ``rc`` -> InfiniBand (ibv_post_send, ibv_post_recv, ibv_poll_cq)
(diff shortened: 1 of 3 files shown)


quasiben:
Also, shamelessly import the TLS config options from UCX
https://github.com/openucx/ucx/blob/e6cffc822710ed032e93e8f46acd389dd590455c/src/ucp/core/ucp_context.c#L75-L94

pentschev:
This is a really good idea, thanks for putting this together @quasiben !

pentschev:
Everything looks good, thanks again @quasiben !

PR #375, opened 2019-12-18
cudf merge benchmark

madsbk:
This PR implements a cudf merge benchmark similar to the benchmark in [dask-cuda](https://github.com/rapidsai/dask-cuda/blob/branch-0.12/dask_cuda/benchmarks/local_cudf_merge.py) but this implementation uses [`run_on_local_network()`](https://github.com/rapidsai/ucx-py/blob/53f7170446566f3a1a79a35bc4e6ffa2d7f683f0/ucp/utils.py#L122) instead of Dask.

```
$ python cudf_merge.py --help
usage: cudf-merge.py [-h] [--chunks-per-dev N] [-d LIST] [-s ip] [-c N]
                     [--frac-match FRAC] [--net-devices LIST]
                     [--profile FILENAME] [--cuda-profile]
                     [--rmm-init-pool-size BYTES]

optional arguments:
  -h, --help            show this help message and exit
  --chunks-per-dev N    Number of chunks per device
  -d LIST, --devs LIST  GPU devices to use (default "0").
  -s ip, --server-address ip
                        Server address (default `ucp.get_address()`).
  -c N, --chunk-size N
  --frac-match FRAC     Fraction of rows that matches (default 0.3)
  --net-devices LIST    List of net devices to use, one for each device or
                        "auto"
  --profile FILENAME    Write profile for each worker to `filename.RANK`
  --cuda-profile        Enable CUDA profiling, use with `nvprof --profile-
                        child-processes --profile-from-start off`
  --rmm-init-pool-size BYTES
                        Initial RMM pool size (default 1/2 total GPU memory)
```

#### Example on a DGX-1 machine
```
$ python  cudf_merge.py -c 100000000 --devs=0,1,2,3,4,5,6,7 --net-devices=auto
cudf merge benchmark
----------------------------
device(s)      | [0, 1, 2, 3, 4, 5, 6, 7]
chunks-per-dev | 1
chunk-size     | 100.00 MB
frac-match     | 0.3
============================
Wall-clock     | 983.40 ms
Bandwidth      | 3.70 GB/s
Throughput     | 26.07 GB/s
```
Notice, _Bandwidth_ is the total aggregated communication bandwidth and _Throughput_ is the processing rate of the dataframe data.

Issue #370, opened 2019-12-12
Small transfers with NVLink are slower than TCP


pentschev:
I first observed this with Dask, in general any transfers smaller than 8MB-10MB with NVLink are slower than with TCP over UCX. Here are some comparisons of TCP over UCX and NVLink with UCX-Py (using `ucp/benchmarks/recv-into-client.py`, CuPy objects, no Dask involved):

TCP:
* 1MB: 373MB/s
* 5MB: 364MB/s
* 10MB: 342MB/s
* 100MB: 400MB/s
* 1000MB: 406MB/s
* 10000MB: Too long to run, but certainly not going to defer much from previous results

NVLink:
* 1MB: 160MB/s
* 5MB: 326MB/s
* 10MB: 795MB/s
* 100MB: 6.7GB/s
* 1000MB: 50GB/s
* 10000MB: 122 GB/s (suspicious, doesn't look accurate)

The purpose of this issue is just to track what we can do to improve it, but I don't believe this is not really possible, most likely the overhead of NVLink makes it inefficient for such small transfers. There were some discussions during the UCX hackathon around this issue, and I believe the solution will be something like falling back to a different transport (e.g., TCP) for small transfers.

cc @randerzander @quasiben @jakirkham

mrocklin:
This slide from @Akshay-Venkatesh has different numbers.  I'm curious what the difference is here.

<img width="1589" alt="Screen Shot 2019-12-12 at 4 44 11 PM" src="https://user-images.githubusercontent.com/306380/70757979-b4dee980-1cfe-11ea-8a6a-3d11fc98f266.png">

pentschev:
I did some more testing with `ucx_perftest` as following:

```
# server
CUDA_VISIBLE_DEVICES=0 UCX_TLS=tcp,sockcm,cuda_copy UCX_SOCKADDR_TLS_PRIORITY=sockcm ucx_perftest -t tag_bw -m cuda -s SIZE -n 10 -p 8999

# client
CUDA_VISIBLE_DEVICES=1 UCX_TLS=tcp,sockcm,cuda_copy UCX_SOCKADDR_TLS_PRORITY=sockcm ucx_perftest HOSTNAME -t tag_bw -m cuda -s SIZE -n 10 -p 8999
```

The above is for CUDA+TCP, for NVLink the only change was appending `cuda_ipc` to `UCX_TLS`.

[...]

PR #378, opened 2019-12-19
Buffed send and receive

madsbk:
This PR make send and receive use a temporary buffer when communicating GPU memory and `UCXPY_COMM_BUFFER=y` is defined. 
Each buffer is associated a single endpoint thus all communication can use the same UCX memory registration. 

**Warning**, for now temporary buffers are cached without any thoughts on memory usage!

Depend on #375.

#### Run on 8 GPUs on a DGX-1
Running the [cudf-merge benchmark](https://github.com/rapidsai/ucx-py/pull/375).

| Setup | Wall-clock | Bandwidth | Throughput | Speedup |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| No Buffer | 1.03 s  | 3.97  GB/s  | 24.97 GB/s  |1 |
| Buffer  | 742.74 ms | 5.34 GB/s |  34.47  GB/s  | 1.39  |
| Buffer + IPC_CACHE | 520.30 ms | 8.95 GB/s | **49.47 GB/s** | 1.98  |

<details>
 <summary>Details</summary>

```
$ python  cudf-merge.py -c 100000000 --devs=0,1,2,3,4,5,6,7 --net-devices=auto
cudf merge benchmark
----------------------------
device(s)      | [0, 1, 2, 3, 4, 5, 6, 7]
chunks-per-dev | 1
chunk-size     | 100.00 MB
frac-match     | 0.3
============================
Wall-clock     | 1.03 s
Bandwidth      | 3.97 GB/s
Throughput     | 24.97 GB/s
```
```
$ UCXPY_COMM_BUFFER=y python  cudf-merge.py -c 100000000 --devs=0,1,2,3,4,5,6,7 --net-devices=auto
cudf merge benchmark
----------------------------
device(s)      | [0, 1, 2, 3, 4, 5, 6, 7]
chunks-per-dev | 1
chunk-size     | 100.00 MB
frac-match     | 0.3
============================
Wall-clock     | 742.74 ms
Bandwidth      | 5.34 GB/s
Throughput     | 34.47 GB/s
```
```
$ UCXPY_COMM_BUFFER=y UCX_CUDA_IPC_CACHE=y python  cudf-merge.py -c 100000000 --devs=0,1,2,3,4,5,6,7 --net-devices=auto
cudf merge benchmark
----------------------------
device(s)      | [0, 1, 2, 3, 4, 5, 6, 7]
chunks-per-dev | 1
chunk-size     | 100.00 MB
frac-match     | 0.3
============================
Wall-clock     | 520.30 ms
Bandwidth      | 8.95 GB/s
Throughput     | 49.20 GB/s
```
</details>

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -23,6 +23,12 @@ from .send_recv import tag_send, tag_recv, stream_send, stream_recv
 from .utils import get_buffer_nbytes
 
 
+if 'UCXPY_COMM_BUFFER' in os.environ:
+    from collections import defaultdict
+    from rmm._lib.device_buffer import DeviceBuffer
+    import cupy
+
+
 cdef assert_ucs_status(ucs_status_t status, msg_context=None):
     if status != UCS_OK:
         msg = "[%s] " % msg_context if msg_context is not None else ""
@@ -572,6 +578,11 @@ class _Endpoint:
         self._cuda_support = "cuda" in ctx.config['TLS'] or ctx.config['TLS'] == "all"
         self._close_after_n_recv = None
 
+        if 'UCXPY_COMM_BUFFER' in os.environ:
+            self._cuda_alloc_cache = defaultdict(list)
+        else:
+            self._cuda_alloc_cache = None
+
     @property
     def uid(self):
         return self._ucp_endpoint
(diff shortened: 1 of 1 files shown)


Akshay-Venkatesh:
@madsbk 
Shouldn't you check if `buffer` passed to `tag_send` already belongs to RMM pool? If yes, it has a single registration associated with it anyway and you can avoid the copy overhead to temporary buffer.

jakirkham:
Thanks @madsbk. Very happy to see this implemented and to see it is already helping performance.

Akshay-Venkatesh:
@madsbk Quite possible that solving OOM issues isn't the goal of the PR at all but I'll bring this up anyway.

About the single registration -- I'm not sure how cache.pop() and cache.append() is implemented internally so correct me if I'm mistaken. When the very first attempt is made to send and len of cache is 0, a new DeviceBuffer of size equal to transfer size (nbytes) is requested from RMM. This device buffer could belong to underlying cudaMalloced allocation `alloc_x`. Also the expectation here is that `alloc_x` probably much greater than `nbytes`. This means that all subsequent requests of size `<= alloc_x` can use the same registration as long as `nbytes <= alloc_x` or as long as aggregate message size of concurrent requests <= `alloc_x`. 

When this limit is exceeded either due to concurrency or message size, then I expected len(cache) to return 0 and this results in a new DeviceBuffer(nbytes) call. I believe this may still result in RMM needing a new cudaMalloced allocation `alloc_y` if it's own internal resources are exhausted. For `alloc_y` there would be a new CUDA IPC registration/memory handle associated (which is fine because it's still better than having multiple active handles when `buffer` doesn't belong to RMM). 

My concern is if RMM does cudaFree to cut down memory usage when it notices memory isn't being used as much. The current bug in UCX (https://github.com/rapidsai/ucx-py/issues/370) means that this cudaFree may not actually translate to memory being freed on the GPU because another peer is mapping that memory and it hasn't called close on the memory handle associated with it.  If this is the case, unfortunately using temporary buffers still doesn't solve the OOM issue in https://github.com/rapidsai/ucx-py/issues/370 

@kkraus14 @quasiben @pentschev thoughts on this?

madsbk:
> @madsbk
> Shouldn't you check if `buffer` passed to `tag_send` already belongs to RMM pool? If yes, it has a single registration associated with it anyway and you can avoid the copy overhead to temporary buffer.

Is this true if the buffer was previously registered for communication with another GPU? That will result in a new CUDA IPC registration right?

Akshay-Venkatesh:
There's just one unique handle for every cudaMalloced region actually.

jakirkham:
Well there are a few things to take into account.

1. RMM can be run will pool mode on/off
2. RMM may have multiple pools (due to growth)

Given these, I'm not sure if simply assuming RMM is around is sufficient.

quasiben:
@Akshay-Venkatesh I believe what you have outlined is good summary and correct.  New allocations will be triggered when the buffer is greater than size of the current transfer buffer _and_ UCX will still hold a mapping to the previous buffer where `cudaFree` calls on the exporter will not remove the memory.  

I don't think this work was done to resolve the OOM issues we were seeing but it is a compelling technique to potentially assist in that effort (similar to what you described HPC folks doing -- reserve scratch space for communication buffers).

madsbk:
Now that RMM and UCX is begging to play nice, we don't need this anymore.

PR #379, opened 2019-12-20
use LocalCUDACluster instead of DGX object in examples

quasiben:


pentschev:
LGTM, thanks @quasiben !

PR #382, opened 2020-01-14
Provide utility function for C string to unicode conversion

jakirkham:
Adds a simple utility function to make it easier to facilitate conversion of C strings (`char*`) to Python strings (`unicode`). As a bonus leverages the Python buffer protocol to avoid unneeded copies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,10 +30,7 @@ ext_modules = [
     ),
     Extension(
         "ucp._libs.send_recv",
-        sources=[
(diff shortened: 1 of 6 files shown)


jakirkham:
Strange. I'm unable to reproduce these issues when running the tests locally. Is there something up with CI that I'm not aware of?

jakirkham:
Replacing with PR ( https://github.com/rapidsai/ucx-py/pull/392 ), which is much simpler.

PR #383, opened 2020-01-14
WIP: Mark C level functions as `nogil`

jakirkham:
These functions should be safe to use without the Python GIL as they are not creating or destroying Python objects. So mark them as such.

Note: This does **not** actually release the GIL. We still have to do that ourselves. However this allows us to use these in code sections where we have already released the GIL. We can also continue to use them when we have the GIL as well.

### DIFF
--- a/ucp/_libs/core_dep.pxd
+++ b/ucp/_libs/core_dep.pxd
@@ -16,7 +16,7 @@ cdef extern from "Python.h":
     Py_buffer* PyMemoryView_GET_BUFFER(PyObject *mview)
 
 
-cdef extern from "src/c_util.h":
+cdef extern from "src/c_util.h" nogil:
     ctypedef struct ucp_listener_params_t:
         pass
 
@@ -40,7 +40,7 @@ cdef extern from "src/c_util.h":
     void c_util_get_ucp_ep_params_free(ucp_ep_params_t *param)
 
 
-cdef extern from "ucp/api/ucp.h":
+cdef extern from "ucp/api/ucp.h" nogil:
     ctypedef struct ucp_context_h:
         pass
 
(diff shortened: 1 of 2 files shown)


quasiben:
Thansk @jakirkham .  Looks like CI is failing with legitimate errors:

> ucp/_libs/send_recv.pyx:90:51: Cannot assign type 'void (void *, ucs_status_t)' to 'ucp_send_callback_t'

jakirkham:
Yeah that should be fixed by PR ( https://github.com/rapidsai/ucx-py/pull/390 ).

jakirkham:
Even though this passes. We should hold off on merging for a bit as there are some warning. These highlight some additional changes we need.

From [the log]( https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/1054/pipeline/ ) we can see the following warnings:

```
warning: ucp/_libs/send_recv.pyx:85:40: Casting a GIL-requiring function into a nogil function circumvents GIL validation
warning: ucp/_libs/send_recv.pyx:132:8: Casting a GIL-requiring function into a nogil function circumvents GIL validation
warning: ucp/_libs/send_recv.pyx:148:40: Casting a GIL-requiring function into a nogil function circumvents GIL validation
warning: ucp/_libs/send_recv.pyx:195:8: Casting a GIL-requiring function into a nogil function circumvents GIL validation
warning: ucp/_libs/core.pyx:353:12: Casting a GIL-requiring function into a nogil function circumvents GIL validation
warning: ucp/_libs/core.pyx:424:12: Casting a GIL-requiring function into a nogil function circumvents GIL validation
```

This is happening because the C functions potentially may not hold the GIL when calling our Cython callbacks. We can fix this by defining our Cython callbacks to also not hold the GIL, but acquire it internally when it is needed.

Will make some changes to this affect in a second.

jakirkham:
To fix the warnings, it makes to type our callbacks as `nogil`. However as we still need the `gil` within the callbacks, we need to acquire the GIL internally. This means we should also release the GIL before calling into C functions. Given all of this, it makes more sense to start a new PR that covers these as well. Have added PR ( https://github.com/rapidsai/ucx-py/pull/391 ).

PR #384, opened 2020-01-14
Simplify PyMemoryView_GET_BUFFER call

jakirkham:
As Cython supplies `PyMemoryView_GET_BUFFER` in its API as of 0.29.14+, simplify the code by using their API. Make a few more simplifications to `_data_from_memoryview` so that it can be simply inlined.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -54,7 +54,7 @@ conda install "cudatoolkit=$CUDA_REL" \
 
 # needed for asynccontextmanager in py36
 conda install -c conda-forge "async_generator" "automake" "libtool" \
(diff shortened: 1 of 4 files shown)


pentschev:
> As Cython supplies PyMemoryView_GET_BUFFER in its API as of 0.29.14+

Perhaps not absolutely necessary, but should we update the version requirement in https://github.com/rapidsai/ucx-py/blob/branch-0.12/ci/gpu/build.sh#L57 and https://github.com/rapidsai/ucx-py/blob/branch-0.12/docs/source/install.rst#build-dependencies ?

jakirkham:
Thanks Peter! Added in the latest commit 🙂

jakirkham:
Thanks all! 😄

PR #385, opened 2020-01-14
Use `uintptr_t` to coerce to/from Python `int`s & C pointers

jakirkham:
Follow the convention in other RAPIDS Cython code of using `uintptr_t` as an intermediate when coercing a Python `int` to a C pointer or vice versa.


jakirkham:
Not at all. Thanks for doing that Peter! 😄

jakirkham:
Thanks all for the reviews! 😄

PR #387, opened 2020-01-14
Make `ucp_ep` a `struct` and `ucp_ep_h` a pointer

jakirkham:
As UCX defines `ucp_ep` as a `struct` and `ucp_ep_h` as a pointer to a `ucp_ep` object, correct the definition in `core_dep.pxd` to accordingly.

ref: https://github.com/openucx/ucx/blob/v1.6.1/src/ucp/api/ucp_def.h#L83


jakirkham:
As there are a few more of these to fix, replacing with PR ( https://github.com/rapidsai/ucx-py/pull/388 ).

PR #390, opened 2020-01-14
Cast callback function pointers to expected type

jakirkham:
Ensure Cython recognizes these function pointers match the expected callback type by explicitly casting them to that type. This becomes more important when trying to release the GIL in these code sections. As Cython seems to not recognize these are function pointers and not Python objects.

### DIFF
--- a/ucp/_libs/send_recv.pyx
+++ b/ucp/_libs/send_recv.pyx
@@ -82,13 +82,13 @@ def tag_send(uintptr_t ucp_ep, buffer, size_t nbytes,
     cdef ucp_ep_h ep = <ucp_ep_h>ucp_ep
     cdef void *data = <void*><uintptr_t>(get_buffer_data(buffer,
                                          check_writable=False))
-    cdef ucp_send_callback_t cb = <ucp_send_callback_t>_send_callback
+    cdef ucp_send_callback_t _send_cb = <ucp_send_callback_t>_send_callback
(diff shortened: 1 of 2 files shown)


ucp/_libs/send_recv.pyx
```diff
@@ -127,25 +128,29 @@ def tag_recv(uintptr_t ucp_worker, buffer, size_t nbytes,
...
+        <ucp_tag_recv_callback_t>_tag_recv_callback
+    )
     cdef ucs_status_ptr_t status = ucp_tag_recv_nb(worker,
                                                    data,
                                                    nbytes,
                                                    ucp_dt_make_contig(1),
                                                    tag,
                                                    -1,
-                                                   _tag_recv_callback)
+                                                   cb)
```
quasiben: This is extremely nit picky but i found the name `_tag_recv_callback` to be useful when reading the code.  What do you think about renaming the variable `_tag_recv_cb` and the same for the other callback funcs ?

jakirkham: Sure. That seems fine.

FWIW the only reason this variable is here is we could not fit the cast in the line length limit. 😉

jakirkham: Addressed in the latest commit. Thanks!

quasiben: Thanks @jakirkham

jakirkham:
Thanks @quasiben! 😄

Issue #251, opened 2019-10-14
Releasing the GIL


jakirkham:
Currently we don't release the GIL in the code. Perhaps this is something we should? If so, it would be good to identify where we can release the GIL. If this is not something we should do, it would be good to understand why.

jakirkham:
@madsbk, do you have any thoughts on this?

madsbk:
I think we should be able to release the GIL at most UCX API calls but they are all non-blocking so I don't think it will have a significant impact on the performance.

mrocklin:
Agreed.  I would wait on this until profiling showed that it was important.

jakirkham:
I'm not as concerned about performance speedup as handling major issues like hangs at this stage. The latter hurts usability so is worth thinking about IMHO.

mrocklin:
I agree that hangs are bad.  Are you saying that holding onto the GIL is
causing UCX to hang?

jakirkham:
Wanted to check and see if there would be any problem pursuing this path should it be needed. Am hearing there shouldn’t be which is good.

mrocklin:
In my mind it might be helpful, but really we should focus on getting a
benchmark up and running quickly so that we can make these kinds of
determinations.  I think that once someone is actually playing with
dask-cudf joins then we'll know a ton more about what work needs to be
done.  I'm personally less excited about cleanup work until we have a
benchmark to help give direction and priorities.

pentschev:
We have since made several improvements to GIL release, plus UCXX should mostly cover this correctly. Therefore, it shouldn't be relevant anymore, closing.

PR #391, opened 2020-01-14
WIP: Release GIL around C functions

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/251

Mark C functions as `nogil`. Adjust the Cython C-level callbacks to not require the GIL (though acquire it internally as needed). Make sure to release the GIL before calling UCX C functions.


pentschev:
This PR looks correct to me, but I don't have much experience on working with the GIL, so I'll leave the review for this to someone with more experience, perhaps @quasiben or @madsbk .

ucp/_libs/core.pyx
```diff
@@ -252,19 +252,20 @@ async def listener_handler(ucp_endpoint, ctx, ucp_worker, func, guarantee_msg_or
...
-    cdef object ctx = <object> a.py_ctx
-    cdef object func = <object> a.py_func
-    asyncio.ensure_future(
-        listener_handler(
-            int(<uintptr_t><void*>ep),
-            ctx,
-            int(<uintptr_t><void*>a.ucp_worker),
-            func,
-            a.guarantee_msg_order
+    with gil:
```
quasiben: This needs to be with the gil ?

quasiben: Oh , any call back needs to be with the gil, correct ?

ucp/_libs/send_recv.pyx
```diff
@@ -82,43 +83,47 @@ def tag_send(uintptr_t ucp_ep, buffer, size_t nbytes,
...
-    cdef ucp_send_callback_t _send_cb = <ucp_send_callback_t>_send_callback
-    cdef ucs_status_ptr_t status = ucp_tag_send_nb(ep,
-                                                   data,
-                                                   nbytes,
-                                                   ucp_dt_make_contig(1),
-                                                   tag,
-                                                   _send_cb)
+    cdef ucp_send_callback_t _send_cb
+    cdef ucs_status_ptr_t status
+    with nogil:
```
quasiben: Should this be with the gil since it's a callback executed by C ?

ucp/_libs/core.pyx
jakirkham: That's right.

ucp/_libs/send_recv.pyx
jakirkham: Our Cython callbacks are acquiring the GIL before they call into Python. So we need to be sure to release the GIL beforehand.

jakirkham:
FWIW this still needs a bit more work. There are other places we could potentially benefit from releasing the GIL.

quasiben:
@jakirkham is there a good test for testing GIL release ?

jakirkham:
We can call [`PyGILState_Check`]( https://docs.python.org/3/c-api/init.html#c.PyGILState_Check ) to verify it is released. Though I may not be totally understanding. What did you have in mind?

quasiben:
i didn't know about `PyGILState_Check` that seems like a good check.  Is it easy to add a test with that ?  I was considering more complicated tasks around threading...

jakirkham:
I suppose we could add some asserts in the code. Is that what you have in mind?

quasiben:
Yeah, I at not very familiar with adding in GIL release so I am mostly speaking out of ignorance.  I checked the Pandas source and I don't think they test GIL release though they do have calls to `PyGILState_Ensure` in the code.  So maybe an assert would be good.

quasiben:
There is a merge conflict.  Once this is resolved we should be good to merge

jakirkham:
> There is a merge conflict

Yeah saw that. Will fix in a bit.

> Once this is resolved we should be good to merge

Yeah we should probably hold off merging until we are releasing the GIL in more places. Will take a look at this tomorrow.

pentschev:
Closing, xref https://github.com/rapidsai/ucx-py/issues/251#issuecomment-3245463217.

Issue #349, opened 2019-12-05
Update doc environment file to use rapidsai channel


jakirkham:
Would be good to update [the doc environment file]( https://github.com/rapidsai/ucx-py/blob/486852efbf42b2004f42dc40d07f583301e38db9/conda/environments/builddocs_py37.yml#L3 ) to use the `rapidsai` channel once things are more finalized.

quasiben:
Agreed.  For the time being I have been using my channel:
https://github.com/quasiben/ucx-py/blob/fix-rtd/conda/environments/builddocs_py37.yml#L3

this was done to over come the `libnuma.so.1` issue with UCX

jakirkham:
Submitted PR ( https://github.com/rapidsai/ucx-py/pull/393 ) to update the channel ordering.

PR #393, opened 2020-01-15
Update doc environment file

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/349

### DIFF
--- a/conda/environments/builddocs_py37.yml
+++ b/conda/environments/builddocs_py37.yml
@@ -1,6 +1,5 @@
 name: ucx_dev
 channels:
-- conda-forge/label/rc_ucx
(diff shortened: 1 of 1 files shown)


quasiben:
Glad to see the specific label being remove.  Currently RTD is still using my fork for docs:

https://github.com/quasiben/ucx-py/blob/fix-rtd/conda/environments/builddocs_py37.yml#L3

This was done to over come the libnuma.so.1 issue with UCX.  UCX is built with libnuma but that library is not available on the RTD build machine and RTD folks are generally not motivated to add it

quasiben:
Thanks @raydouglass and @jakirkham

PR #395, opened 2020-01-15
Use RMM's DeviceBuffer (in cudf-merge and elsewhere)

jakirkham:
Use RMM `DeviceBuffer` for allocations.

### DIFF
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -16,8 +16,6 @@ from dask.utils import format_bytes, format_time, parse_bytes
 
 import cudf
 import cupy
-import numba
-import numba.cuda
 import numpy as np
 import rmm
 import ucp
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -35,15 +35,6 @@ async def send_df(ep, df):
         await ep.send(frame)
 
 
-def device_array(shape, dtype):
-    dtype = np.dtype(dtype)
-    nbytes = np.prod(shape) * dtype.itemsize
-    buf = rmm.DeviceBuffer(size=nbytes)
-    arr = cupy.asarray(buf).view(dtype).reshape(shape)
-    arr = numba.cuda.as_cuda_array(arr)
-    return arr
-
-
 async def recv_df(ep):
     header_nbytes = np.empty((1,), dtype=np.uint64)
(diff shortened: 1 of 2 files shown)


jakirkham:
For both of the following, ran this command on latest nightlies:

```
UCXPY_IFNAME="enp1s0f0" UCX_TLS=tcp,cuda_copy,sockcm python ucx-py/benchmarks/cudf-merge.py --chunks-per-dev 4 --chunk-size 10000 --rmm-init-pool-size 100
```

Without this change I see this performance:

```
cudf merge benchmark
----------------------------
device(s)      | [0]
chunks-per-dev | 4
rows-per-chunk | 10000
data-processed | 1.28 MB
frac-match     | 0.3
============================
Wall-clock     | 48.53 ms
Bandwidth      | 11.62 MB/s
Throughput     | 26.38 MB/s
```

With this change I see this performance:

```
cudf merge benchmark
----------------------------
device(s)      | [0]
chunks-per-dev | 4
rows-per-chunk | 10000
data-processed | 1.28 MB
frac-match     | 0.3
============================
Wall-clock     | 42.80 ms
Bandwidth      | 16.21 MB/s
Throughput     | 29.91 MB/s
```

pentschev:
@jakirkham could you also share results for larger problems (perhaps increasing `chunks-per-dev` or `rows-per-chunk`, or maybe both)?

I see a gain of performance here, but I'm not sure I understand why that happens and how you came to the conclusion that replacing RMM for CuPy would have that effect. Could you elaborate on that?

jakirkham:
Just for clarity both cases are [allocating through RMM]( https://github.com/rapidsai/ucx-py/blob/593fc8d7d1e4b9997be0816466b7ed9a27c53cef/benchmarks/cudf-merge.py#L172-L173 ). It's just a difference in APIs. Am making this change w.r.t. the discussion in issue ( https://github.com/rapidsai/rmm/issues/180 ) to remove this API from RMM.

That said, I was curious of the performance impact. So ran things to see. Agree it is not big (nor would I expect it to be). I don't expect there to be a particularly noticeable scale difference as we are using the same underlying allocator.

pentschev:
Thanks for the info @jakirkham .

jakirkham:
Thanks Peter! 😄

madsbk:
Nice work @jakirkham

PR #397, opened 2020-01-21
Use default UCX_CUDA_IPC_CACHE value again

pentschev:
Using a CUDA memory pool with the CUDA IPC cache doesn't cause an OOM error, this is because the pool should not release/reallocate of memory during runtime.

Given that our most common use case today comprises the use of an RMM pool, we may reenable this to facilitate users' life. It's still possible to disable the cache by setting the environment variable `UCX_CUDA_IPC_CACHE=n`, and this may be required for any uses cases that do not include the use of a memory pool at the risk of running OOM otherwise.

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -20,14 +20,6 @@ if "UCX_MEMTYPE_CACHE" not in os.environ:
     logger.debug("Setting env UCX_MEMTYPE_CACHE=n, which is required by UCX")
     os.environ["UCX_MEMTYPE_CACHE"] = "n"
 
-if "UCX_CUDA_IPC_CACHE" not in os.environ:
-    # See <https://github.com/openucx/ucx/issues/4410>
-    logger.debug(
-        "Setting env UCX_CUDA_IPC_CACHE=n, which is required to avoid NVLink memory "
-        "leaks"
-    )
-    os.environ["UCX_CUDA_IPC_CACHE"] = "n"
-
 if "UCX_SOCKADDR_TLS_PRIORITY" not in os.environ:
     logger.debug(
         "Setting env UCX_SOCKADDR_TLS_PRIORITY=sockcm, "


quasiben:
@madsbk I am going to merge but wanted to make sure you were aware of this change

cjnolet:
@pentschev I'm on the nightly of ucx-py. I have the `UCX_CUDA_IPC_CACHE=n` in my envioronment, but it doesn't show up when I call `ucp.get_config()`. I'm hoping this will solve the following error I'm getting:

```
[deeplearn:47174:0:47227] cuda_ipc_cache.c:198  Fatal: dest:47171: failed to open ipc mem handle. addr:0x7f30e8e00000 len:80000 (Element already exists)
```

quasiben:
@cjnolet it won't show up in `get_config` as this only returns the `UCS` config options.  is that right @madsbk  ?

cjnolet:
@pentschev @jakirkham @quasiben, I just verified this change is what broke cuML's distributed nearest neighbors. It doesn't show in CI yet because it's only single GPU, but i noticed last week the pytest would no longer run on my 2-gpu workstation. Setting `UCX_CUDA_IPC_CACHE=n` seems to have no effect.

cjnolet:
False alarm- I'm still trying to root cause. It fixed almost magically by downgrading to 0.12 but upgrading it again doesn't break it again. This is very weird, but unfortunately makes it very hard to suggest a solution when users encounter it.

pentschev:
> @cjnolet it won't show up in get_config as this only returns the UCS config options. is that right @madsbk ?

`ucp.get_config` only returns UCP options, and `UCX_CUDA_IPC_CACHE` is a UCT variable.

> @pentschev @jakirkham @quasiben, I just verified this change is what broke cuML's distributed nearest neighbors. It doesn't show in CI yet because it's only single GPU, but i noticed last week the pytest would no longer run on my 2-gpu workstation. Setting UCX_CUDA_IPC_CACHE=n seems to have no effect.

This may be very well the case, as that is an error that can only happen with the CUDA IPC cache enabled. While you can probably solve that one issue by disabling the CUDA IPC cache, performance will likely be _very_ poor, unless you happen to be only transferring reasonably large buffers (1GB+).

> False alarm- I'm still trying to root cause. It fixed almost magically by downgrading to 0.12 but upgrading it again doesn't break it again. This is very weird, but unfortunately makes it very hard to suggest a solution when users encounter it.

It could be that you downgraded and since we were enforcing CUDA IPC cache to be disabled back then, thus it didn't error out. If you do disable the CUDA IPC cache, you must make sure that all processes have it disabled as well, and that could be a bit cumbersome if you're starting Dask scheduler/worker/client in different processes. One thing you could do to be 100% sure in 0.13 that you're indeed disabling the cache is to add that configuration directly to `$CONDA_PREFIX/lib/python3.7/site-packages/ucp/__init__.py`, this way all processes will ensure that's disabled.

PR #414, opened 2020-02-10
Rename package to ucx-py in setup.py

pentschev:
The package name is `ucx-py` for both [rapidsai](https://anaconda.org/rapidsai/ucx-py) and [conda-forge](https://anaconda.org/conda-forge/ucx-py) channels, therefore I think renaming it is good to make things consistent.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ ext_modules = [
 ]
 
 setup(
-    name="ucxpy",
+    name="ucx-py",
     packages=["ucp"],
     ext_modules=ext_modules,
     cmdclass=versioneer.get_cmdclass(),


pentschev:
> I would prefer if we could change everything to ucxpy, but if that is too cumbersome, let's change the package name back to ucx-py.

I guess the package name comes from the way we've been writing UCX-Py, so `ucx-py` makes sense, although `ucxpy` also makes sense (consistent with `numpy`, `cupy`, etc.). That said, I don't have a strong opinion either way, but @jakirkham may be able to tell us if changing would be cumbersome if we decide to that.

jacobtomlinson:
I also prefer `uxcpy`.

quasiben:
If we make this change we'll need to update at least the following.  Probably others as well

## Install Dependencies 

- https://github.com/rapidsai/ucx-split-feedstock/blob/master/recipe/meta.yaml
- https://github.com/rapidsai/cuml/blob/94a48421b6dc4fac112c329b2fe9c3688c5feb98/conda/recipes/cuml/meta.yaml#L37
- https://github.com/rapidsai/build/blob/42a2e8ca70518428d6fa7d7d20c89ed6bb8e1d4f/templates/docker/conda_install_devel_deps#L49
- https://github.com/rapidsai/build/blob/42a2e8ca70518428d6fa7d7d20c89ed6bb8e1d4f/generatedDockerfiles/Dockerfile.ubuntu-devel
- https://github.com/rapidsai/build/blob/42a2e8ca70518428d6fa7d7d20c89ed6bb8e1d4f/generatedDockerfiles/Dockerfile.centos7-devel

## CI Install Dependencies 
- https://github.com/rapidsai/dask-cuda/blob/branch-0.13/ci/gpu/build.sh#L68

pentschev:
> If we make this change we'll need to update at least the following. Probably others as well

Just to be clear, by "this change" I think you mean to rename the package to `ucxpy`, and not the change proposed in this PR, right?

quasiben:
Ah, i miss read the change -- I thought it was going from `ucx-py` to `ucxpy`.  Please ignore my comment above

jakirkham:
This seems less cumbersome then dropping the `-` everywhere 🙂

Yeah `-` should be fine. It's `_` that needs to be avoided in package names

pentschev:
Alright, given the complication in renaming everywhere to `ucxpy` I'll go ahead and merge this for the short term. @madsbk and @jacobtomlinson I would encourage you to file an issue if you feel like we should indeed rename the package.

jakirkham:
Should add I've now grabbed the PyPI name for `ucx-py` as well.

pentschev:
Thanks everyone for the comments here!

jakirkham:
Thanks Peter! 😄

Issue #413, opened 2020-02-08
Sequence of UCX calls may be incorrect


pentschev:
@Akshay-Venkatesh and I have identified a hang in `local-send-recv.py`, the hang is non-deterministic (happening about every 5-10 runs for me on a DGX-1) and can be reproduced with the command below on UCX master branch:

```
UCX_RNDV_SCHEME=get_zcopy UCX_TLS=tcp,sockcm,cuda_copy,cuda_ipc UCX_SOCKADDR_TLS_PRIORITY=sockcm python local-send-recv.py -o cupy -d 0 -e 1
```

@Akshay-Venkatesh has worked on a potential fix in https://github.com/openucx/ucx/pull/4748, but it seems like we may be executing the [`ucp_worker_arm` + `ucp_tag_recv_nb` + `epoll_wait` sequence that is illegal in UCX](https://github.com/openucx/ucx/pull/4748#issuecomment-583778695). That being said, we need to verify if we are not doing anything incorrect in UCX-Py, as the aforementioned fix may not be incorporated given the illegality of the sequence.

Issue #411, opened 2020-02-08
msgpack unpack received extra data - crashes worker


jakirkham:
Am seeing workers getting killed due to an issue parsing headers in `msgpack` when running the workflow in issue ( https://github.com/rapidsai/ucx-py/issues/402 ). The workers don't appear to come back either. If I had to guess, the problem is similar to issue ( https://github.com/rapidsai/rmm/issues/223 ) and a workaround like the one found in PR ( https://github.com/rapidsai/rmm/pull/224 ) would be sufficient.

Note: It's important that this is only applied to host buffers and not device ones. Attempting to fix the latter would likely crash the program.

```python
distributed.core - ERROR - unpack(b) received extra data.
Traceback (most recent call last):
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/core.py", line 412, in handle_comm
    result = await result
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/scheduler.py", line 1731, in add_worker
    await self.handle_worker(comm=comm, worker=address)
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/scheduler.py", line 2688, in handle_worker
    await self.handle_stream(comm=comm, extra={"worker": worker})
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/core.py", line 456, in handle_stream
    msgs = await comm.read()
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/comm/ucx.py", line 196, in read
    frames, deserialize=self.deserialize, deserializers=deserializers
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/comm/utils.py", line 69, in from_frames
    res = _from_frames()
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/comm/utils.py", line 55, in _from_frames
    frames, deserialize=deserialize, deserializers=deserializers
  File "/datasets/jkirkham/miniconda/envs/rapids13dev/lib/python3.7/site-packages/distributed/protocol/core.py", line 101, in loads
    msg = loads_msgpack(small_header, small_payload)
[...]

PR #415, opened 2020-02-10
[Merge] Epoll wait follows worker arming 

madsbk:
This PR fixes https://github.com/rapidsai/ucx-py/issues/413 by making sure that asyncio's next state after calling `ucp_worker_arm()` is epoll wait.

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -503,16 +503,36 @@ cdef class ApplicationContext:
         """Bind an asyncio reader to a UCX epoll file descripter"""
         assert self.blocking_progress_mode is True
 
-        def _fd_reader_callback():
-            cdef ucs_status_t status
+        async def _arm():
+            # When arming the worker, the following must be true:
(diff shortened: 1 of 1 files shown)


pentschev:
Thanks @madsbk for working on this. I've tested changes locally and I can confirm there's no hang anymore.

I also discussed with @Akshay-Venkatesh and he's going to work on a patch to assert that arm isn't called before verifying that the worker is drained first. I think this patch will allow us to later confirm that the fixes here are everything we need.

Are you expecting any more changes here @madsbk ?

jakirkham:
@pentschev, could you please take another look? 🙂

jakirkham:
Appears this fixes issue ( https://github.com/rapidsai/ucx-py/issues/411 ).

jakirkham:
Thanks for working on this @madsbk! 😄

Was able to get up and running with this locally easily and it seems to have fixed a painful issue. 🎉

If you find any issues with this @pentschev, please let us know. 🙂

PR #416, opened 2020-02-11
Allow `ucx-proc` to be unspecified

jakirkham:
Based on UCX's `dlopen` behavior, it should be ok to install the GPU build on a CPU only machine. So leave `ucx-proc` unspecified and allow Conda to just pick one. This should help with our doc building issues where `ucx-proc`'s CPU build is not found.

cc @quasiben

### DIFF
--- a/conda/environments/builddocs_py37.yml
+++ b/conda/environments/builddocs_py37.yml
@@ -10,7 +10,6 @@ dependencies:
 - sphinxcontrib-websupport
 - nbsphinx
 - numpydoc
-- ucx-proc=*=cpu
 - recommonmark
 - pandoc=<2.0.0
 - pip


madsbk:
@jakirkham, is this ready to be merged?

jakirkham:
I think when @quasiben and I talked offline we concluded this might not help the underlying doc building issue. Hence why this has lingered.

That said, no objections to merging if it's useful 🙂

quasiben:
I think we could merge this.  The underlying doc build still relies on a custom built conda package which doesn't use libnuma -- we can't get libnuma installed on the RTD machines/conatiners.

jakirkham:
That's right. We have issue ( https://github.com/openucx/ucx/issues/4570 ) open upstream, which would be another way to resolve this.

jakirkham:
Don't think this is still needed. So going to close it out. Though please let me know if that's not the case and can always reopen.

PR #417, opened 2020-02-11
Correct UCX_RNDV_THRESH environment variable name

jakirkham:
From the wiki, this appears to be named `UCX_RNDV_THRESH`. So have updated it here. However not totally sure about this. So please check.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -53,9 +53,9 @@ This is a UCX CUDA Memory optimization which enables/disables a remote endpoint 
 
 Values: n/y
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Yeah actually now I see it [here]( https://github.com/openucx/ucx/blob/9d06c3aa68bb23fcd8412d94a179da511c4b9c40/src/ucp/core/ucp_context.c#L120-L122 ). So this seems right. @quasiben, could you please confirm? 🙂

quasiben:
thanks @jakirkham

PR #418, opened 2020-02-19
Splitting the asyncio and the UCX specific API into different files

madsbk:
This PR splits the implementation into a low-level part `ucx_api.pyx` that follows the semantic of UCX closely and a high-level part that introduce asyncio semantic and tries to make it easy to use for Python users.
This will also make it a lot easier to merge UCX-Py into upstream UCX because we can start with the low-level part that is fairly simple, follows the semantic of UCX, and is easy to maintain. 

#### low-level
Python wrappers of the UCX API where handlers such as `ucp_worker_h`, `ucp_listener_h` and  `ucp_ep_h` are wrapped in Python classes. It does not use asyncio and all functions is non-blocking.
The only major difference in semantic compared to UCX is that all the send/receive functions _always_ calls the callback function even when the communication finished immediately.

#### high-level
New semantic based on asyncio where communication calls return futures that the use later can `await`. Generally, at this level we try to make the common use case as easy as possible; taking the liberty to change the API to fit Python users.


madsbk:
rerun tests

ucp/public_api.py
```diff
@@ -177,7 +183,7 @@ def get_config():
     """
 
     if _ctx is None:
-        return core.get_config()
+        return ucx_api.get_default_options()
```
pentschev: Will this do the right thing? What if environment variables are passed, will those be read?

ucp/public_api.py
```diff
@@ -261,18 +293,58 @@ def abort(self):
...
+                    len(msg),
+                    self._ctrl_tag_send,
+                    pending_msg=self.pending_msg_list[-1],
+                )
+            except exceptions.UCXError:
+                pass  # The peer might already be shutting down
+            # Give all current outstanding send() calls a chance to return
+            self._ctx.progress()
+            await asyncio.sleep(0)
+        finally:
```
pentschev: Just checking, did you really intend try/finally without an except?

ucp/public_api.py
```diff
@@ -298,27 +389,56 @@ def closed(self):
...
         """
-        return self._ep.get_ucp_worker()
+        return self._worker.handle
 
     def get_ucp_endpoint(self):
         """Returns the underlying UCP endpoint handle (ucp_ep_h)
         as a Python integer.
         """
-        return self._ep.get_ucp_endpoint()
+        return self._ep.handle
```
pentschev: Is this valid if the endpoint is closed or should there be a check/exception before?

tests/test_custom_send_recv.py
```diff
@@ -59,12 +59,9 @@ def __init__(self, ep):
             )
             await self.ep.send(np.array([nbytes(f) for f in frames], dtype=np.uint64))
             # Send frames
-            # breakpoint()
             for frame in frames:
                 if nbytes(frame) > 0:
                     await self.ep.send(frame)
-                else:
-                    await self.ep.send(np.empty(0, dtype=np.uint8))
```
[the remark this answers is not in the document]
madsbk: When `nbytes(frame) == 0` no frame is send/recv, which is also the behavior of Dask/Distributed.

tests/test_reset.py
```diff
@@ -59,7 +62,10 @@ def server(ep):
     lt = ucp.create_listener(server)
     ep = await ucp.create_endpoint(ucp.get_address(), lt.port)
     del lt
-    with pytest.raises(ucp.exceptions.UCXError, match="_ucp_endpoint"):
+    with pytest.raises(
+        ucp.exceptions.UCXError,
+        match="The following objects are still referencing ApplicationContext",
+    ):
```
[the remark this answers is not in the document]
madsbk: Yes, but the specific object pointers varies thus the error varies as well. I think this checking is enough to make sure that the right error messaged has been raised.

ucp/public_api.py
madsbk: Yes, it calls `ucp_config_print()`, which also reads environment variables but I can see that the name is very misleading :)
Renamed it to `get_current_options()`

ucp/public_api.py
madsbk: Yes, on close error we want to both call `self.abort()` and propagate the close error.

ucp/public_api.py
madsbk: You are right, I am adding some close checks

madsbk:
Thanks for the review @pentschev.
The CI fails because of https://github.com/dask/distributed/pull/3549

madsbk:
Much of `ucx_apy.pyx` is new and should be able to standalone. This is what I am referring to as the low-level package in https://github.com/rapidsai/ucx-py/issues/429.

tests/test_reset.py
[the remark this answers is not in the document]
pentschev: My question was more regarding "The _following_ objects...", will there be objects following that error message?

ucp/public_api.py
pentschev: Ok, this is what I thought, indeed renaming makes more sense. Thanks for adjusting that!

ucp/public_api.py
pentschev: Thanks for adding those.

tests/test_reset.py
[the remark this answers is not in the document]
madsbk: Yes, the error message lists the objects still holding a reference to `ApplicationContext`.
`pytest.raises(match=<str>)` passes if `<str>` is somewhere in the error message.

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,475 @@
...
+    """Python representation of `ucp_ep_h`
+    Please use `ucx_ep_create()` to contruct an instance of this class
+    """
+    cdef:
+        ucp_ep_h _handle
+        bint initialized
+
+    def close(self, UCXWorker worker):
+        cdef ucs_status_ptr_t status
+        if self.initialized:
```
pentschev: I see `self.initialized` is set in `ucx_ep_create`, but what if it isn't called? Is it guaranteed that the default for `bint` is `False`?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,475 @@
...
+    def handle(self):
+        assert self.initialized
+        return int(<uintptr_t><void*>self._handle)
+
+
+cdef UCXEndpoint ucx_ep_create(ucp_ep_h ep):
+    ret = UCXEndpoint()
+    ret._handle = ep
+    ret.initialized = True
+    return ret
```
pentschev: Since you're making this the "Pythonic" side of things, wouldn't it make sense that creating a `UCXEndpoint` from a class instance suffice and not requiring `ucx_ep_create` anymore?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,475 @@
...
+    if req.data == NULL:
+        # This callback function was called before handle_comm_result
+        # had a chance to set req.data
+        req.finished = True
+        req.received = length
+        return
+    cdef object req_data = <object> req.data
+
+    log_str = None
+    msg = "Error receiving%s " %(" \"%s\":" % log_str if log_str else ":")
```
pentschev: Maybe move these two lines inside `elif status != UCX_OK`?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,475 @@
...
+                                 ucp_tag_recv_info_t *info):
+    _ucx_recv_callback(request, status, info.length)
+
+
+cdef uintptr_t handle_comm_result(
+    ucs_status_ptr_t status, dict req_data, expected_receive=None
+):
+    log_str = None
+    msg = "Comm Error%s " %(" \"%s\":" % log_str if log_str else ":")
+    exception = None
```
pentschev: Maybe move these two lines inside `elif UCS_PTR_IS_ERR(status)` as well?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,475 @@
...
+    else:
+        exception = None
+
+    ucp_request_reset(request)
+    ucp_request_free(request)
+    req_data["cb_func"](exception, *req_data["cb_args"])
+    Py_DECREF(req_data)
+
+
+def ucx_tag_send(UCXEndpoint ep, buffer, size_t nbytes,
```
pentschev: I see `ucx_stream_{send,recv}`, but only `ucx_tag_send`, is `ucx_tag_recv` missing or is this intended?

ucp/_libs/ucx_api.pyx
madsbk: Yes that is guaranteed, but to make it more readable I have added code that initialize them explicitly.

ucp/_libs/ucx_api.pyx
madsbk: `ucx_tag_recv()` was a method instead `UCXWorker`, moved it out to the other functions

jakirkham:
rerun tests

ucp/_libs/ucx_api.pyx
madsbk: Fixed

madsbk:
rerun tests

MattBBaker:
Oh nice, looks like it will be even smoother to get the RDMA work in. Only quibble is that in the copyright dates they all say 2019, and it's now 2020.

Also, with this split, once development settles down a bit, would it make sense to upstream the really low level parts that touch the C api to the core ucx?

madsbk:
> Oh nice, looks like it will be even smoother to get the RDMA work in. Only quibble is that in the copyright dates they all say 2019, and it's now 2020.

Actually, I don't know how copyright dates works; are we supposed to update the dates every year? Or only on new files?

> Also, with this split, once development settles down a bit, would it make sense to upstream the really low level parts that touch the C api to the core ucx?

Yes! this is one of the motivations for the spitting. The idea is to make two separate Python packages: https://github.com/rapidsai/ucx-py/issues/429

MattBBaker:
Oh nice, I missed that issue.

For copyright dates, on the UCX end we up the year when a file is touched.

pentschev:
> Actually, I don't know how copyright dates works; are we supposed to update the dates every year? Or only on new files?

From our internal slack:

> Remember that if you're creating a new file with a new license header that the year is now 2020, so please use that in the license 

> I believe updates should include both years. That is, if you update a 2019 file, then change it to 2019-2020. If you update a 2018 file, then change it to 2018-2020. Likewise, if you update a 2018-2019 file, then change it to 2018-2020. If it is a new file then it should just be 2020.

quasiben:
I tested this PR with `local_cudf_merge.py` and I'm seeing a seg fault (pasted below).  I see similar errors when only using nvlink (--disable-infiniband).

```
(20200306) bzaitlen@dgx13:/datasets/bzaitlen/GitRepos/dask-cuda/dask_cuda/benchmarks$ python local_cudf_merge.py -p ucx -d 0,1,2,3,4,5,6,7 -c 100000000 --enable-nvlink --enable-tcp-over-ucx --enable-infiniband  --runs 1
ERROR:Ignored except: <class 'SystemError'> <method 'progress' of 'ucp._libs.ucx_api.UCXWorker' objects> returned a result with an error set
[dgx13:28962:0:28962] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))
==== backtrace ====
    0  /datasets/bzaitlen/miniconda3/envs/20200306/lib/libucs.so.0(+0x22631) [0x7fe801781631]
    1  /datasets/bzaitlen/miniconda3/envs/20200306/lib/libucs.so.0(+0x22802) [0x7fe801781802]
    2  /lib/x86_64-linux-gnu/libpthread.so.0(+0x12890) [0x7fe93107a890]
...
===================
distributed.nanny - WARNING - Restarting worker
[dgx13:28740:0:28868] rc_verbs_iface.c:69   send completion with error: transport retry counter exceeded qpn 0x396a wrid 0x6 vendor_err 0x81
==== backtrace ====
    0  /datasets/bzaitlen/miniconda3/envs/20200306/lib/libucs.so.0(ucs_fatal_error_message+0x51) [0x7fcb695eb171]
    1  /datasets/bzaitlen/miniconda3/envs/20200306/lib/libucs.so.0(+0x22f37) [0x7fcb695eef37]
    2  /datasets/bzaitlen/miniconda3/envs/20200306/lib/libucs.so.0(ucs_log_dispatch+0xd9) [0x7fcb695ef069]
    3  /datasets/bzaitlen/miniconda3/envs/20200306/lib/ucx/libuct_ib.so.0(+0x1b42e) [0x7fcb65ba442e]
    4  /datasets/bzaitlen/miniconda3/envs/20200306/lib/ucx/libuct_ib.so.0(+0x1ba6e) [0x7fcb65ba4a6e]
```

jakirkham:
Did you merge with `branch-0.13`?

quasiben:
No, I tested just this branch

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+        log = logging.warning
+    else:
+        log = logging.error
+    log("Ignored except: %s %s" % (type(msg), msg))
+
+
+async def exchange_peer_info(ucp_endpoint, msg_tag, ctrl_tag, guarantee_msg_order):
+    """Help function that exchange endpoint information"""
+
+    my_info = struct.pack("QQi", msg_tag, ctrl_tag, guarantee_msg_order)
```
quasiben: I think this needs to be `QQ?`. `guarantee_msg_order` is a bint which, I think, should be `?` for boolean on the Python side

jakirkham: As I can never remember the codes for these things, here's the [table]( https://docs.python.org/3.8/library/struct.html#format-characters ).

Should we be worried about [endianness]( https://docs.python.org/3.8/library/struct.html#byte-order-size-and-alignment )?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+    return ret
+
+
+def get_ucx_version():
+    cdef unsigned int a, b, c
+    ucp_get_version(&a, &b, &c)
+    return (a, b, c)
+
+
+cdef void _listener_callback(ucp_ep_h ep, void *args) except *:
```
jakirkham: As I just recently learned, this can sometimes behave funny when asyncio is involved as it well check exception set by asyncio and raise. Doing `cdef` without `void` or `except *` fixes that.

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
+# cython: language_level=3
+
+import socket
+from libc.stdio cimport FILE, fflush, fclose
+from libc.stdlib cimport free
+from libc.string cimport memset
+from libc.stdint cimport uintptr_t
```
jakirkham: As it is used below and I don't think it gets imported by default (though could be wrong here).

```suggestion
from libc.stdint cimport uint16_t, uintptr_t
```

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+    return config
+
+
+cdef ucx_config_to_dict(ucp_config_t *config):
+    """Returns a dict of a UCX config"""
+    cdef char *text
+    cdef size_t text_len
+    cdef FILE *text_fd = open_memstream(&text, &text_len)
+    assert(text_fd != NULL)
+    ret = {}
```
jakirkham: Cython can optimize some things if we type this.

```suggestion
    cdef dict ret = {}
```

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+            raise UCXConfigError(msg)
+    return config
+
+
+cdef ucx_config_to_dict(ucp_config_t *config):
+    """Returns a dict of a UCX config"""
+    cdef char *text
+    cdef size_t text_len
+    cdef FILE *text_fd = open_memstream(&text, &text_len)
+    assert(text_fd != NULL)
```
jakirkham: Thoughts on raising an `IOError` instead here?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+
+
+def get_ucx_version():
+    cdef unsigned int a, b, c
+    ucp_get_version(&a, &b, &c)
+    return (a, b, c)
+
+
+cdef void _listener_callback(ucp_ep_h ep, void *args) except *:
+    cdef dict cb_data = <dict> args
```
jakirkham: How does this work? Do we not need to cast through `PyObject*` first?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+        cdef ucp_listener_accept_callback_t _listener_cb = (
+            <ucp_listener_accept_callback_t>_listener_callback
+        )
+        if c_util_get_ucp_listener_params(&params,
+                                          port,
+                                          _listener_cb,
+                                          <void*> self.cb_data):
+            raise MemoryError("Failed allocation of ucp_ep_params_t")
+
+        logging.info("create_listener() - Start listening on port %d" % port)
```
jakirkham: Have we setup some kind of logging handler somewhere?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+        if self.initialized:
+            self.initialized = False
+            ucp_listener_destroy(self.handle)
+
+
+cdef class UCXContext:
+    cdef:
+        ucp_context_h _context
+        bint _initialized
+        object _config
```
jakirkham: IIUC this is just a `dict`. Is that right?

```suggestion
        dict _config
```

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+        ucp_params.features = (UCP_FEATURE_TAG |  # noqa
+                               UCP_FEATURE_WAKEUP |  # noqa
+                               UCP_FEATURE_STREAM)
+
+        ucp_params.request_size = sizeof(ucp_request)
+        ucp_params.request_init = (
+            <ucp_request_init_callback_t>ucp_request_reset
+        )
+
+        cdef ucp_config_t *config = _read_ucx_config(config_dict)
```
jakirkham: Should we do some sort of `try...finally...` thing below to make sure this gets freed if some other error comes up?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+        # descriptor that we can wait on later.
+        cdef ucs_status_t status
+        cdef int ucp_epoll_fd
+        cdef epoll_event ev
+        cdef int err
+        status = ucp_worker_get_efd(self._handle, &ucp_epoll_fd)
+        assert_ucs_status(status)
+        self.arm()
+        epoll_fd = epoll_create(1)
+        assert(epoll_fd != -1)
```
jakirkham: Should we raise some sort of `IOError` or other appropriate error here?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+        ev.data.fd = ucp_epoll_fd
+        ev.data.ptr = NULL
+        ev.data.u32 = 0
+        ev.data.u64 = 0
+        ev.events = EPOLLIN
+        err = epoll_ctl(epoll_fd, EPOLL_CTL_ADD, ucp_epoll_fd, &ev)
+        assert(err == 0)
```
jakirkham: Same question about raising an `IOError` or other error.

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+                ucp_request_free(status)
+
+    def info(self):
+        assert self.initialized
+        # Making `ucp_ep_print_info()` write into a memstream,
+        # convert it to a Python string, clean up, and return string.
+        cdef char *text
+        cdef size_t text_len
+        cdef FILE *text_fd = open_memstream(&text, &text_len)
+        assert(text_fd != NULL)
```
jakirkham: Should we raise an `IOError` here?

ucp/_libs/ucx_api.pyx
```diff
@@ -0,0 +1,490 @@
...
+        ucp_ep_print_info(self._handle, text_fd)
+        fflush(text_fd)
+        cdef unicode py_text = text.decode()
+        fclose(text_fd)
+        free(text)
```
jakirkham: Should we have some `try...finally...` block above to make sure we cleanup if an error comes up in the intervening code?

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+    ep.pending_msg_list.append({"log": log})
+    shutdown_fut = send_recv.tag_recv(
+        ep._worker,
+        msg,
+        len(msg),
+        ep._ctrl_tag_recv,
+        pending_msg=ep.pending_msg_list[-1],
+    )
+
+    shutdown_fut.add_done_callback(partial(handle_ctrl_msg, weakref.ref(ep), log, msg))
```
jakirkham: Why do we use a `weakref` here?

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+def listener_handler(ucp_endpoint, ctx, worker, func, guarantee_msg_order):
+    async def run(ucp_endpoint, ctx, worker, func, guarantee_msg_order):
+        loop = asyncio.get_event_loop()
+        # TODO: exceptions in this callback is never showed when no
```
jakirkham:
```suggestion
        # TODO: exceptions in this callback is never shown when no
```

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+            async def _func(ep):  # coroutine wrapper
+                func(ep)
+
+            await _func(ep)
+
+    asyncio.ensure_future(run(ucp_endpoint, ctx, worker, func, guarantee_msg_order))
+
+
+async def _non_blocking_mode(weakref_ctx):
+    """This help function maintains a UCX progress loop.
```
jakirkham:
```suggestion
    """This helper function maintains a UCX progress loop.
```

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+        ctx = weakref_ctx()
+        if ctx is None:
+            return
+        ctx.progress()
+        del ctx
+        await asyncio.sleep(0)
+
+
+async def _arm_worker(weakref_ctx, rsock, event_loop):
+    """This help function args the worker.
```
jakirkham:
```suggestion
    """This help function arms the worker.
```

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+    Notice, it only keeps a weak reference to `ApplicationContext`, which makes it
+    possible to call `ucp.reset()` even when this loop is running.
+    """
+    while True:
+        ctx = weakref_ctx()
+        if ctx is None:
+            return
+        ctx.progress()
+        del ctx
+        await asyncio.sleep(0)
```
jakirkham: Couple questions. Why do we need the explicit `del`? Also why is `sleep(0)` needed?

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+        if ctx is None:
+            return
+        ctx.progress()
+        # This IO task returns when all non-IO tasks are finished.
+        await event_loop.sock_recv(rsock, 1)
+        if ctx._worker.arm():
+            # At this point we know that asyncio's next state is
+            # epoll wait.
+            break
+        del ctx
```
jakirkham: Won't this scope out normally without the `del`? Or is there some reason it is explicitly needed?

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+            task.cancel()
+        if self.dangling_arm_task is not None:
+            self.dangling_arm_task.cancel()
+        # Notice, worker and context might never have been created
+        if self._worker is not None:
+            self._worker.destroy()
+        if self._ctx is not None:
+            self._ctx.destroy()
+        if self.blocking_progress_mode and self.epoll_fd != -1:
+            close_fd(self.epoll_fd)
```
jakirkham: Is it possible to move the cleanup stuff into Cython (for example) to avoid the need for an explicit `__del__`?

ucp/application_context.py
```diff
@@ -0,0 +1,376 @@
...
+
+    def create_listener(self, callback_func, port, guarantee_msg_order):
+        self.continuous_ucx_progress()
+        if port in (None, 0):
+            # Get a random port number and check if it's not used yet. Doing this
+            # without relying on `socket` allows preventing UCX errors such as
+            # "none of the available transports can listen for connections", due
+            # to the socket still being in TIME_WAIT state.
+            try:
+                with open("/proc/sys/net/ipv4/ip_local_port_range") as f:
```
jakirkham: This feels a little fragile just because there may be systems that don't have this. However given your point above, I'm not sure of another option. Just noting this concern for now.

ucp/public_api.py
```diff
@@ -261,18 +293,59 @@ def abort(self):
...
+                    msg,
+                    len(msg),
+                    self._ctrl_tag_send,
+                    pending_msg=self.pending_msg_list[-1],
+                )
+            except exceptions.UCXError:
+                pass  # The peer might already be shutting down
+            # Give all current outstanding send() calls a chance to return
+            self._ctx.progress()
+            await asyncio.sleep(0)
```
jakirkham: Why is `sleep(0)` needed here?

jakirkham:
Thanks Mads! 😄

This seems like a cleaner separation of concerns to me. Very nice to see async functions moving to Python and Cython being reserved for C functions. Had a few observations above. Mostly in places where we might be able to tighten things up a bit and ensure things are robust to different failure modes. Otherwise agree this would be nice to see land.

Should add the crash Ben is seeing is a little concerning, but I think this is the same `cdef void ... except *` problem identified a little bit ago. So hopefully a mixture of that fix and the comments above should address fix that behavior.

ucp/application_context.py
pentschev: This is not really part of this PR, but was added by me in https://github.com/rapidsai/ucx-py/pull/423. For systems that don't have this, we have default values in the `except` block.

ucp/application_context.py
madsbk: > I think this needs to be `QQ?`. `guarantee_msg_order` is a bint which, I think, should be `?` for boolean on the Python side

I think `"i"` is correct, according to [this](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#types) `bint` is a `int` interpreted as a Python boolean.

madsbk: > Should we be worried about [endianness](https://docs.python.org/3.8/library/struct.html#byte-order-size-and-alignment)?

I think by omitting byte order symbol, we get automatic byte-order swap if the two endpoints have different endianness.
https://stackoverflow.com/questions/27506474/how-to-byte-swap-a-32-bit-integer-in-python

ucp/_libs/ucx_api.pyx
madsbk: It's imported in https://github.com/madsbk/ucx-py/blob/cython_refactoring/ucp/_libs/ucx_api_dep.pxd

ucp/application_context.py
madsbk: We need to delete `ctx` to avoid holding a reference to `ApplicationContext`, which would make `ucp.reset()` fails.

We use `sleep(0)` to give other co-routines a chance to run.

ucp/application_context.py
madsbk: To avoid holding a reference to ApplicationContext, which would make ucp.reset() fails.

ucp/_libs/ucx_api.pyx
madsbk: This isn't  necessary, the difference in the generated C code is:
```c
// Single cast
  __pyx_t_1 = ((PyObject *)__pyx_v_args);

/ Double Cast
  __pyx_t_1 = ((PyObject *)((PyObject *)__pyx_v_args));
```

ucp/_libs/ucx_api.pyx
madsbk: Yes here: https://github.com/rapidsai/ucx-py/blob/branch-0.13/ucp/__init__.py#L15
We need to clean up the logging and follow best practice, which i honestly don't know what is :)
But let's do that in a new PR.

ucp/_libs/ucx_api.pyx
madsbk: Fixed

ucp/_libs/ucx_api.pyx
madsbk: Fixed

ucp/_libs/ucx_api.pyx
madsbk: Fixed

ucp/application_context.py
madsbk: You are right, fixed

ucp/application_context.py
madsbk: I thought it was better to have it in `__del__()` than `__dealloc__()` ?

ucp/public_api.py
madsbk: To give other co-routines a chance to run before calling `self.abort()`, which cancel pending messages.

(Updated the code comment )

ucp/_libs/ucx_api.pyx
madsbk: Fixed

ucp/_libs/ucx_api.pyx
madsbk: Fixed

ucp/_libs/ucx_api.pyx
madsbk: I have rewritten the error handling, we now use https://github.com/madsbk/ucx-py/blob/cython_refactoring/ucp/_libs/ucx_api.pyx#L25 to handle callback exceptions.

tests/test_reset.py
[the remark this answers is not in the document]
jakirkham: Does this answer you question @pentschev or did you have more questions?

ucp/_libs/ucx_api.pyx
jakirkham: Did this get resolved @madsbk?

ucp/application_context.py
jakirkham: Interesting. So `struct` is somehow tracking endianness?

ucp/_libs/ucx_api.pyx
jakirkham: That sounds like a nice solution. Thanks Mads! 😄

ucp/_libs/ucx_api.pyx
jakirkham: Ok. Sounds good.

ucp/application_context.py
jakirkham: Got it. Thanks!

tests/test_reset.py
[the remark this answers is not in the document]
pentschev: Yes, thanks both.

ucp/application_context.py
jakirkham: Does `0` give them enough of a chance? Should this be something larger?

ucp/application_context.py
jakirkham: Well they both have their issues. `__dealloc__` relies on the Python interpreter to run the cleanup step, which may not happen during a bad shutdown. Though the same could be said of `__del__`. However `__del__` has [even more issues]( https://stackoverflow.com/a/1481512/3877089 ). So unless we are particularly worried about the bad shutdown case, would go the `__dealloc__` route.

ucp/application_context.py
jakirkham: Ok I'll raise an issue then.

ucp/public_api.py
jakirkham: Is `0` enough time or should we be choosing some small number here?

jakirkham:
cc @cjnolet (for visibility)

ucp/application_context.py
```diff
@@ -0,0 +1,373 @@
...
+        self._ctx = None
+        self._worker = None
+        if blocking_progress_mode is not None:
+            self.blocking_progress_mode = blocking_progress_mode
+        elif "UCXPY_NON_BLOCKING_MODE" in environ:
+            self.blocking_progress_mode = False
+        else:
+            self.blocking_progress_mode = True
+
+        # Create context and worker, whic might fail
```
jakirkham:
```suggestion
        # Create context and worker, which might fail
```

ucp/public_api.py
```diff
@@ -82,7 +88,7 @@ def create_listener(callback_func, port=None, guarantee_msg_order=True):
     argument -- the Endpoint connected to the client.
 
     In order to call ucp.reset() inside callback_func remember to
-    close the Endpoint given as an argument. It is not enough to
+    close the Endpoint given as an argument.
```
jakirkham: Was this meant to be deleted?

jakirkham:
Thanks Mads! Looks good. Had only a couple of minor questions on existing threads. Think this is close to completion.

Have you tried the benchmarks in this repo on this change? If so, how do they fair?

jakirkham:
Looks like [this build is failing]( https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/1382/pipeline ) with this exception:

```python
tests/test_send_recv_two_workers.py::test_send_recv_cu[dataframe] Process client:
  File "/conda/envs/gdf/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/conda/envs/gdf/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/var/lib/jenkins/workspace/rapidsai/gpuci/ucx-py/prb/ucx-py-gpu-build/tests/test_send_recv_two_workers.py", line 53, in client
    before_rx, before_tx = total_nvlink_transfer()
  File "/var/lib/jenkins/workspace/rapidsai/gpuci/ucx-py/prb/ucx-py-gpu-build/tests/test_send_recv_two_workers.py", line 264, in total_nvlink_transfer
    transfer = pynvml.nvmlDeviceGetNvLinkUtilizationCounter(handle, i, 0)
  File "/conda/envs/gdf/lib/python3.6/site-packages/pynvml/nvml.py", line 2006, in nvmlDeviceGetNvLinkUtilizationCounter
    check_return(ret)
  File "/conda/envs/gdf/lib/python3.6/site-packages/pynvml/nvml.py", line 366, in check_return
    raise NVMLError(ret)
pynvml.nvml.NVMLError_NotSupported: Not Supported
Build timed out (after 20 minutes). Marking the build as failed.
```

madsbk:
This PR is getting too complex. I am splitting it up into smaller pieces.

quasiben:
Ok, happy to help with reviews/fixes to push this work along if need be.

PR #423, opened 2020-03-02
Use psutil instead of socket to get a random port number

pentschev:
This change will get a random port number without creating a socket on it, that sometimes prevents a dask worker from starting due to the socket being in a `TIME_WAIT` state.

We would see messages like the one below:

```
[1583153840.078159] [dgx12:67177:0]   ucp_listener.c:431  UCX  ERROR none of the available transports can listen for connections on 0.0.0.0:41245
```

and ultimately leading the start of a dask worker to fail with the following traceback:

```python-traceback
Traceback (most recent call last):
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/bin/dask-cuda-worker", line 8, in <module>
    sys.exit(go())
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/dask_cuda/dask_cuda_worker.py", line 382, in go
    main()
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/dask_cuda/dask_cuda_worker.py", line 373, in main
    loop.run_sync(run)
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/tornado/ioloop.py", line 532, in run_sync
    return future_cell[0].result()
  File "/home/pentschev/miniconda3/envs/rn-101-0.13.0b200301/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
[...]

PR #426, opened 2020-03-03
Add psutil as a dependency

pentschev:
`psutil` was introduced as a dependency in https://github.com/rapidsai/ucx-py/pull/423, but it was not updated as such, this PR addresses it.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -56,7 +56,7 @@ conda install "cudatoolkit=$CUDA_REL" \
 conda install -c conda-forge "async_generator" "automake" "libtool" \
                               "cmake" "automake" "autoconf" "cython>=0.29.14,<3.0.0a0" \
                               "pytest" "pkg-config" "pytest-asyncio" \
-                              "pynvml" "libhwloc"
+                              "pynvml" "libhwloc" "psutil"
 
 # Install the master version of dask and distributed
 logger "pip install git+https://github.com/dask/distributed.git --upgrade --no-deps"
--- a/conda/environments/builddocs_py37.yml
+++ b/conda/environments/builddocs_py37.yml
@@ -14,6 +14,7 @@ dependencies:
 - recommonmark
 - pandoc=<2.0.0
 - pip
+- psutil
 - libhwloc
 - ucx
 - ucx-py
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -46,7 +46,7 @@ Build Dependencies
 
     conda create -n ucx -c conda-forge \
         automake make libtool pkg-config \
-        libhwloc \
+        libhwloc psutil \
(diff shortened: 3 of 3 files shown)


quasiben:
Thanks @pentschev -- I should've caught the first time around

pentschev:
> Thanks @pentschev -- I should've caught the first time around

No worries, I also didn't catch it because I think it happens to be a dependency to RAPIDS somewhere, that's why builds didn't fail either.

jakirkham:
Could you also add `install_requires` to `setup.py` and include `psutil` there?

https://github.com/pypa/sampleproject/blob/90706c518dbcd3824d82002eeed9fdc29c1a258e/setup.py#L147-L153

pentschev:
I thought UCX-Py doesn't have requirements as it's always based on a conda environment: https://ucx-py.readthedocs.io/en/latest/install.html#source ?

I think if we decide to add `install_requires` to `setup.py`, then we should include all dependencies and update documentation above as well, but I would argue that this is of lesser importance pre-GTC.

jakirkham:
It's not about non-Conda installs. It's about clear documentation of the dependencies. This is where I would look to see new UCX-Py requirements for example 😉. Should add this would affect dev installs as well.

jakirkham:
Should add we also do this in other RAPIDS projects. Here are the relevant lines from cuDF for example:

https://github.com/rapidsai/cudf/blob/64de40005f0c6459007771caf9945c2e0c9ef96b/python/cudf/setup.py#L13

https://github.com/rapidsai/cudf/blob/64de40005f0c6459007771caf9945c2e0c9ef96b/python/cudf/setup.py#L92

pentschev:
I agree with that, but the problem is we need other dependencies as well, for instance `libhwloc` and `ucx` which aren't even available as pip packages, how would we deal with that?

jakirkham:
We skip them. This wasn’t something `install_requires` was really designed for.

pentschev:
> We skip them. This wasn’t something install_requires was really designed for.

Ok, then could you explain me what was `install_requires` designed for? To me the description in https://github.com/pypa/sampleproject/blob/90706c518dbcd3824d82002eeed9fdc29c1a258e/setup.py#L147-L149 seems to say that it should list all dependencies for a project, which for UCX-Py includes `ucx` and `libhwloc`, among others.

jakirkham:
It's a way of listing Python dependencies.

pentschev:
> It's a way of listing Python dependencies.

So why would we skip other packages such as `ucx` and `libhwloc`, they're both Python dependencies to this package.

pentschev:
rerun tests

pentschev:
I'm merging this now, but we can address your concerns in a different issue/PR if you so wish @jakirkham .

Issue #424, opened 2020-03-02
Document and Define more UCX config options


quasiben:
We often are running with the following config: UCX_RNDV_THRESHOLD=8092 UCX_TCP_TX_SEG_SIZE=8M UCX_TCP_RX_SEG_SIZE=8M.  We should update defaults in ucx-py.  We might not need defaults for `UCX_TCP_...` but wanted to include it in the issue for discussion.

cc @pentschev

quasiben:
resolved by https://github.com/rapidsai/ucx-py/pull/427

PR #427, opened 2020-03-03
Add default values for rndv_thresh rx/tx seg size and update docs

quasiben:
https://github.com/rapidsai/ucx-py/issues/424

cc @pentschev

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -57,6 +57,8 @@ Values: n/y
 
 This is a configurable parameter used by UCX to help determine which transport method should be used.  For example, on machines with multiple GPUs, and with NVLink enabled, UCX can deliver messages either through TCP or NVLink.  Sending GPU buffers over TCP is costly as it triggers a device-to-host on the sender side, and then host-to-device transfer on the receiver side --  we want to avoid these kinds of transfers when NVLink is available.  If a buffer is below the threshold, `Rendezvous-Protocol <https://github.com/openucx/ucx/wiki/Rendezvous-Protocol>`_ is triggered and for UCX-Py users, this will typically mean messages will be delivered through TCP.  Depending on the application, messages can be quite small, therefore, we recommend setting a small value if the application uses NVLink or InfiniBand: ``UCX_RNDV_THRESH=8192``
 
+Values: Int (default: 8192)
+
 
 ``UCX_RNDV_SCHEME``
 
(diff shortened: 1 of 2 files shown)


docs/source/configuration.rst
```diff
@@ -57,6 +57,8 @@ Values: n/y
 
 This is a configurable parameter used by UCX to help determine which transport method should be used.  For example, on machines with multiple GPUs, and with NVLink enabled, UCX can deliver messages either through TCP or NVLink.  Sending GPU buffers over TCP is costly as it triggers a device-to-host on the sender side, and then host-to-device transfer on the receiver side --  we want to avoid these kinds of transfers when NVLink is available.  If a buffer is below the threshold, `Rendezvous-Protocol <https://github.com/openucx/ucx/wiki/Rendezvous-Protocol>`_ is triggered and for UCX-Py users, this will typically mean messages will be delivered through TCP.  Depending on the application, messages can be quite small, therefore, we recommend setting a small value if the application uses NVLink or InfiniBand: ``UCX_RNDV_THRESH=8192``
 
+Values: Int (default: 8192)
```
pentschev: Maybe we should make it clear that the `8192` default value applies to UCX-Py (not UCX core).

docs/source/configuration.rst
```diff
@@ -68,6 +70,15 @@ Values:
...
 
+``UCX_TCP_RX_SEG_SIZE``
+Size of send copy-out buffer when receiving.  This environment variable controls the size of the buffer on the host when receiving data over TCP.
+
+``UCX_TCP_TX_SEG_SIZE``
+
+Size of send copy-out buffer when transmitting.  This environment variable controls the size of the buffer on the host when sending data over TCP.
+
+When mixing TCP with other transports methods, users should take care to properly tune these parameters.  UCX-Py uses ``8M`` as the default value for both RX/TX.
+
```
pentschev: I would also add a note like:

Note that these variables will impact CUDA transfers as well when no NVLink or InfiniBand is available between two UCX-Py processes. The `UCX_TCP_RX_SEG_SIZE` and `UCX_TCP_TX_SEG_SIZE` will cause the `HostToDevice` and `DeviceToHost` copies of buffers to be broken down in several chunks when the size of a buffer exceeds the size defined by these two variables. If an application is expected to transfer very large buffers, increasing such values may improve overall performance.

quasiben:
@pentschev thanks for the note.  I think that is especially helpful. Here is how the update is rendered
Screen Shot 2020-03-03 at 5 19 52 PM

docs/source/configuration.rst
```diff
@@ -68,8 +74,27 @@ Values:
...
+UCX_TCP_TX_SEG_SIZE
+```````````````````
+
+Size of send copy-out buffer when transmitting.  This environment variable controls the size of the buffer on the host when sending data over TCP.
+
+UCX-Py uses ``8M`` as the default value for both RX/TX.
+
+.. note::
+    Users should take care to properly tune ``UCX_TCP_{RX/TX}_SEG_SIZE`` parameters when mixing when mixing TCP with other transports methods as well as when
```
pentschev:
```suggestion
    Users should take care to properly tune ``UCX_TCP_{RX/TX}_SEG_SIZE`` parameters when mixing TCP with other transports methods as well as when
```

quasiben: Thanks @pentschev

Issue #430, opened 2020-03-05
Destination is unreachable when using hostname


pentschev:
When using the server's hostname to connect from client, we get a `Destination is unreachable` error:

```python-traceback
$ python recv-into-client.py -s dgx12
[1583366523.058256] [dgx12:52644:0]           sock.c:164  UCX  ERROR connect(fd=83, dest_addr=255.255.255.255:13337) failed: Network is unreachable
Traceback (most recent call last):
  File "recv-into-client.py", line 206, in <module>
    loop.run_until_complete(main())
  File "/datasets/pentschev/miniconda3/envs/ucx-src-101-0.13.0b200304/lib/python3.7/asyncio/base_events.py", line 583, in run_until_complete
    return future.result()
  File "recv-into-client.py", line 138, in connect
    ep = await ucp.create_endpoint(host, port)
  File "/datasets/pentschev/miniconda3/envs/ucx-src-101-0.13.0b200304/lib/python3.7/site-packages/ucp/public_api.py", line 126, in create_endpoint
    return await _get_ctx().create_endpoint(ip_address, port, guarantee_msg_order)
  File "ucp/_libs/core.pyx", line 470, in create_endpoint
  File "ucp/_libs/core.pyx", line 32, in ucp._libs.core.assert_ucs_status
ucp.exceptions.UCXError: Destination is unreachable
```

The same test above works when using the IP address of the main network interface, and we can see the same behavior with `dask-cuda-worker` (see https://github.com/rapidsai/dask-cuda/issues/240).

I'm not entirely sure, but I think it may be due to the hostname resolving to `127.0.0.1`, thus using the `lo` interface. However, if we use the hostname with `ucx_perftest` things work, so this is potentially an issue on the UCX-Py side.

jakirkham:
Fixed by PR ( https://github.com/rapidsai/ucx-py/pull/431 ).

PR #431, opened 2020-03-05
Support host names

madsbk:
Fixes issue #430

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -461,6 +461,7 @@ cdef class ApplicationContext:
         self.continuous_ucx_progress()
 
         cdef ucp_ep_params_t params
+        ip_address = socket.gethostbyname(ip_address)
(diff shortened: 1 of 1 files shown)


pentschev:
Wow, that was quick! Thanks @madsbk , I'm sure @randerzander will appreciate it.

pentschev:
rerun tests

quasiben:
We are working on the fix for distributed -- alternatively, we can comment out the cupy test while we sort out issues:
https://github.com/rapidsai/ucx-py/blob/branch-0.13/ci/gpu/build.sh#L134

pentschev:
If you think this will take another day or two only, I think we can wait instead of removing tests. It's not critical since we can still use IP addresses.

jakirkham:
rerun tests

quasiben:
rerun tests

jakirkham:
Thanks all! 😄

Issue #435, opened 2020-03-06
Confusing Breakage


quasiben:
Since the merge of https://github.com/rapidsai/ucx-py/pull/433 we are seeing a Ignored error:

> ERROR:Ignored except: <class 'SystemError'> <built-in function isinstance> returned a result with an error set

This can be generated on any DGX machine with the following:

> python local_cudf_merge.py -p ucx -d 0,1 -c 100000000 --disable-nvlink --enable-tcp-over-ucx --enable-infiniband  --runs 1

Perhaps more importantly, often, though not always, the job hangs.  Two things could be possible:

1) A new bug was introduced
2) We have exposed a previously undetected bug

It should also be noted the prior to this PR #433, we see ignored statements like the following:

```
Exception ignored in: 'ucp._libs.send_recv._stream_recv_callback'
asyncio.base_futures.InvalidStateError: invalid state
asyncio.base_futures.InvalidStateError: invalid state
Exception ignored in: 'ucp._libs.send_recv._send_callback'
asyncio.base_futures.InvalidStateError: invalid state
```

jakirkham:
Do you have a traceback? Would be good to get more context. 🙂

FWIW that PR only makes those errors visible to us. They were always there before. It also avoids a segfault encountered by not handling those errors on occasion.

quasiben:
No traceback yet -- it hangs

pentschev:
> It also avoids a segfault encountered by not handling those errors on occasion.

What segfault are you referring to?

quasiben:
Is it possible that it should be `except +` instead of `except *` ?

jakirkham:
`except +` is meant for C++ exceptions. `except *` is used to indicate Python exceptions.

jakirkham:
> What segfault are you referring to

In pure Python it's impossible to not propagate an exception. However in C this is really easy to do. In fact you just never check for the exception. However the Python interpreter doesn't handle this well. Particularly when lots of exceptions are thrown.

jakirkham:
Where are you running this?

quasiben:
I am running on DGX13 but should be reproducible anywhere

quasiben:
Excuse the ignorance of my Cython question.  Was hunting around and naively came across `+`

pentschev:
> > What segfault are you referring to
> 
[...]

PR #436, opened 2020-03-07
Drop `void` return type when raising exceptions

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/435

After poking at this a bit locally, it seems Cython's `except *` is picking up exceptions set through `asyncio` even when it shouldn't. Not sure if this behavior is due to Python or Cython. That said, we can avoid using `except *` here by using `cdef` without the `void` return type (this is the same as `cdef object`). Based on running things locally this seems to behave correctly for me. Though please check as well.

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -255,7 +255,7 @@ async def listener_handler(ucp_endpoint, ctx, ucp_worker, func, guarantee_msg_or
         await _func(pub_ep)
 
 
-cdef void _listener_callback(ucp_ep_h ep, void *args) except *:
+cdef _listener_callback(ucp_ep_h ep, void *args):
     cdef _listener_callback_args *a = <_listener_callback_args *> args
     cdef object ctx = <object> a.py_ctx
     cdef object func = <object> a.py_func
--- a/ucp/_libs/send_recv.pyx
+++ b/ucp/_libs/send_recv.pyx
@@ -53,7 +53,7 @@ cdef create_future_from_comm_status(ucs_status_ptr_t status,
     return ret
 
 
-cdef void _send_callback(void *request, ucs_status_t status) except *:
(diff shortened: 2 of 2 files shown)


jakirkham:
cc @quasiben @pentschev

quasiben:
This still hangs for me with the dask-cuda merge benchmark.  My Cython expertise is quite low but is there now conflict with argument and definitions in `core_dep.pxd`
https://github.com/rapidsai/ucx-py/blob/9e3b3c0ac54a741fb52681aafb372f3cf5bb5247/ucp/_libs/core_dep.pxd#L136 ?

jakirkham:
Where are you running this? FWIW I ran this on dgx-15 without issues. Just want to exclude the possibility of hardware issues getting in the mix.

That's a good point. So I guess the exception will get ignored in those case anyways. Yeah it's hard to solve this correctly without more-or-less doing what Mads has already done. If we are not happy with this, maybe we can try to help get his work in? Sorry I don't have a better answer 😞

jakirkham:
Should add when I tested this I was using Vibhu's MRE from PR ( https://github.com/rapidsai/ucx-py/issues/402 ). This seems potentially more complex than the merge benchmark. So I'm pretty sure I'm exercise all the same things, but please let me know if that's incorrect.

pentschev:
Yeah, there's no error with the merge benchmark, Ben and I were testing both that and the merge benchmark. It's intriguing why it doesn't fail with that, in fact the tests for `u1` and `u16` fail, but `u256` and larger pass, so it's probably due to the size of the transfer, but right now I have no clue where the issue may lie.

madsbk:
This should be fixed in #418

madsbk:
https://github.com/rapidsai/ucx-py/pull/439 replaces this PR

PR #438, opened 2020-03-10
CI: ignoring the two worker tests

madsbk:
It is simply not reliable to run in CI


quasiben:
Agreed.  After GTC we should figure out what is going on

quasiben:
@rapidsai/ops can you merge this ?

PR #443, opened 2020-03-11
Backporting test fixes from #418

madsbk:
The PR backport some test fixes from #418

PR #444, opened 2020-03-11
Make sure we use the right event loop when modifying futures

madsbk:
This PR makes sure that when modifying a future, the event loop used when creating the future is open.


quasiben:
Any ideas why this and PR #443 are failing in the same spot?  These tests pass locally

madsbk:
> Any ideas why this and PR #443 are failing in the same spot? These tests pass locally

No, just tried in a clean up-to-date conda environment and everything passes.

quasiben:
rerun tests

PR #451, opened 2020-03-12
Cleaning up the shutdown of Context, Endpoint, and Listener

madsbk:
This PR makes sure that the shutdown order of Context, Endpoints, and Listeners is correct.

Context now has a weak reference to all _child_ objects and makes sure to abort them before destroying the UCX context.

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -224,6 +224,7 @@ async def listener_handler(ucp_endpoint, ctx, ucp_worker, func, guarantee_msg_or
         ctrl_tag_recv=ctrl_tag,
         guarantee_msg_order=guarantee_msg_order
     )
+    ctx.children.append(weakref.ref(ep))
 
     logging.debug(
         "listener_handler() server: %s, msg-tag-send: %s, "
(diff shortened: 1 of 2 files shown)


quasiben:
When running this PR locally with IB I am seeing failures:

> UCX_NET_DEVICES=mlx5_0:1 UCX_MEMTYPE_CACHE=n UCX_SOCKADDR_TLS_PRIORITY=sockcm UCX_TLS=rc,cuda_copy,tcp,sockcm py.test -s -v tests/test_custom_send_recv.py


```
tests/test_custom_send_recv.py::test_send_recv_cudf[<lambda>0] [1584028521.945038] [prm-dgx-20:63935:0]         parser.c:1578 UCX  WARN  unused env variable: UCX_TCP_TX_SEG_SIZE,... (set UCX_WARN_UNUSED_ENV_VARS=n to suppress this warning)
Exception ignored in: <function Listener.__del__ at 0x7fa3f4ed8050>
  File "/gpfs/fs1/bzaitlen/miniconda3/envs/20200312/lib/python3.7/site-packages/ucp/public_api.py", line 220, in __del__
    self.close()
  File "/gpfs/fs1/bzaitlen/miniconda3/envs/20200312/lib/python3.7/site-packages/ucp/public_api.py", line 234, in close
    self._b.abort()
AttributeError: 'ucp._libs.core._Listener' object has no attribute 'abort'
FAILED
```

quasiben:
Apologies, I had the wrong branch.  My tests are looking good now with IB

ucp/_libs/core.pyx
```diff
@@ -398,10 +406,16 @@ cdef class ApplicationContext:
 
     def __dealloc__(self):
         if self.initiated:
-            for task in self.progress_tasks:
+            # Aborting all objects working in this context
+            for child in (self.children if self.children else []):
```
jakirkham: Since it is a `list` type and is initialized to be empty, seems like this should work 🙂

```suggestion
            for child in self.children:
```

ucp/_libs/core.pyx
```diff
@@ -317,12 +323,13 @@ cdef class ApplicationContext:
         int epoll_fd
         object event_loops_binded_for_progress
         object progress_tasks
```
jakirkham:
```suggestion
        list progress_tasks
```

ucp/_libs/core.pyx
```diff
@@ -398,10 +406,16 @@ cdef class ApplicationContext:
 
     def __dealloc__(self):
         if self.initiated:
-            for task in self.progress_tasks:
+            # Aborting all objects working in this context
+            for child in (self.children if self.children else []):
+                child = child()
+                if child is not None:
+                    child.abort()
+            for task in (self.progress_tasks if self.progress_tasks else []):
```
jakirkham: Same simplification here 🙂

```suggestion
            for task in self.progress_tasks:
```

madsbk: I added this because sometimes `self.progress_tasks` will be None. We are not setting it to None anywhere so my guess is that it is part of the Python-destructor-order-issue.

jakirkham: Thanks for clarifying. Would it make sense to set it to `[]` in those cases instead? Not critical if that seems like too much for this PR.

madsbk: I am not sure that it is possible?

jakirkham: Could we add this in an `__init__` or `__cinit__` perhaps?

madsbk: I am not sure I follow. Do you want to wrap `self.progress_tasks` and `self.children` in a class?

jakirkham: Am suggesting that `_Listener` have a `__cinit__` where these are initialized with `[]`. Though this is not critical. We can follow-up in another PR if needed. 🙂

madsbk: I see, yeah let's wait with that to another PR.

PR #456, opened 2020-03-13
[REVIEW] Clean up of UCX progress

madsbk:
This PR moves ucx progress in its own file and use classes to simplify the clean up process.

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -20,6 +20,7 @@ from ..exceptions import (
     UCXWarning,
     UCXConfigError,
 )
+from .. import continuous_ucx_progress
 
 from .send_recv import tag_send, tag_recv, stream_send, stream_recv, log_errors
 from .utils import get_buffer_nbytes
@@ -272,20 +273,6 @@ cdef void _listener_callback(ucp_ep_h ep, void *args):
         )
 
 
-async def _non_blocking_mode(weakref_ctx):
-    """This help function maintains a UCX progress loop.
-    Notice, it only keeps a weak reference to `ApplicationContext`, which makes it
-    possible to call `ucp.reset()` even when this loop is running.
-    """
-    while True:
-        ctx = weakref_ctx()
-        if ctx is None:
-            return
-        ctx.progress()
-        del ctx
-        await asyncio.sleep(0)
-
-
 cdef class _Listener:
     """This represents the private part of Listener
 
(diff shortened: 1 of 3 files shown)


quasiben:
rerun tests

madsbk:
rerun tests

ucp/continuous_ucx_progress.py
```diff
@@ -0,0 +1,108 @@
...
+        event_loop: asyncio.EventLoop
+            The event loop to do progress in.
+        """
+        self.weakref_ctx = weakref.ref(ctx)
+        self.event_loop = event_loop
+        self.asyncio_task = None
+
+    def __del__(self):
+        if self.asyncio_task is not None:
+            self.asyncio_task.cancel()
```
jakirkham: What happens if the task scopes out normally?

ucp/continuous_ucx_progress.py
```diff
@@ -0,0 +1,108 @@
...
+        return hash(self) == hash(other)
+
+
+class NonBlockingMode(ProgressTask):
+    def __init__(self, ctx, event_loop):
+        super().__init__(ctx, event_loop)
+        self.asyncio_task = event_loop.create_task(self._progress_task())
+
+    async def _progress_task(self):
+        """This help function maintains a UCX progress loop."""
```
jakirkham:
```suggestion
        """This helper function maintains a UCX progress loop."""
```

ucp/_libs/core.pyx
```diff
@@ -533,75 +513,32 @@ cdef class ApplicationContext:
...
         else:
-            self._non_blocking_progress_mode(loop)
+            task = continuous_ucx_progress.NonBlockingMode(self, loop)
+        self.progress_tasks.append(task)
 
     def get_ucp_worker(self):
         return int(<uintptr_t><void*>self.worker)
 
     def get_config(self):
         return self.config
```
jakirkham: Trimming some whitespace 🙂

```suggestion
```

ucp/_libs/core.pyx
```diff
@@ -411,10 +394,7 @@ cdef class ApplicationContext:
                 child = child()
                 if child is not None:
                     child.abort()
-            for task in (self.progress_tasks if self.progress_tasks else []):
-                task.cancel()
-            if self.dangling_arm_task is not None:
-                self.dangling_arm_task.cancel()
+            self.progress_tasks = None
```
jakirkham: Should this be assigned an empty `list`?

```suggestion
            self.progress_tasks = []
```

ucp/continuous_ucx_progress.py
madsbk: That is fine, when this goes out of scope the continuous progress will stop

jakirkham: Ok, am just wondering do we need to call `.cancel()` explicitly or does this already happens for us?

madsbk: In some cases we have to call it explicitly.
PS: note that it is fine to call `.cancel()` even if the future is already done.

quasiben:
Thanks @madsbk and thank you for the review @jakirkham

PR #457, opened 2020-03-15
Add test that causes RMM deallocate segfault

pentschev:
DO NOT MERGE!

This is a fork of #453 to demonstrate the RMM deallocate segfault seen with InfiniBand. The segfault has been previously experienced in some much more complex workflows with cuDF, but this seems to simulate the same segfault.

Removing https://github.com/rapidsai/ucx-py/compare/branch-0.13...pentschev:rmm-deallocate-segfault?expand=1#diff-17c6de8e1db1dd4d3eb6f8b14fb4303eR43-R44 or disabling the RMM pool avoids the segfault.

It can be run on a machine with IB properly configured and it only happens if InfiniBand is enabled. An example of how to run it is below:

```
# server
UCX_HANDLE_ERRORS=debug UCX_NET_DEVICES=mlx5_1:1 CUDA_VISIBLE_DEVICES=3 UCX_MEMTYPE_CACHE=n UCX_TLS=rc,tcp,cuda_copy,sockcm UCX_SOCKADDR_TLS_PRIORITY=sockcm python tests/debug-tests/send.py

# client
UCX_NET_DEVICES=mlx5_3:1 CUDA_VISIBLE_DEVICES=7 UCX_MEMTYPE_CACHE=n UCX_TLS=rc,tcp,cuda_copy,sockcm UCX_SOCKADDR_TLS_PRIORITY=sockcm  python tests/debug-tests/recv.py
```

For me it has failed almost every time, succeeding only every 5-10 runs. The crash will happen after just before exiting the program, which will open/close an endpoint 10 times and send/recv 10 frames each time, it will take about 20 seconds from the beginning until the segfault. Below are the first 20 frames of the failing thread's stack:

```
#0  0x00007f9e91b9e23a in __waitpid (pid=pid@entry=26739, stat_loc=stat_loc@entry=0x7f9dead6afcc,
    options=options@entry=0) at ../sysdeps/unix/sysv/linux/waitpid.c:30
#1  0x00007f9dec1ad6b5 in ucs_debugger_attach () at debug/debug.c:703
#2  0x00007f9dec1b0207 in ucs_handle_error (message=message@entry=0x7f9dec1c238f "address not mapped to object")
    at debug/debug.c:985
#3  0x00007f9dec1b0631 in ucs_debug_handle_error_signal (signo=signo@entry=11,
    cause=0x7f9dec1c238f "address not mapped to object", fmt=fmt@entry=0x7f9dec1c2431 " at address %p")
    at debug/debug.c:941
#4  0x00007f9dec1b0802 in ucs_error_signal_handler (signo=11, info=0x7f9dead6b5f0, context=<optimized out>)
    at debug/debug.c:963
#5  <signal handler called>
#6  0x0000000000000000 in ?? ()
#7  0x00007f9de4135bf8 in rmm::mr::device_memory_resource::deallocate (stream=0x0, bytes=<optimized out>,
    p=<optimized out>, this=<optimized out>)
   from /datasets/pentschev/miniconda3/envs/rn-101-0.13.0b200315/lib/python3.7/site-packages/rmm/_lib/device_buffer.cpython-37m-x86_64-linux-gnu.so
#8  rmm::device_buffer::deallocate (this=0x564f70f1adc0) at rmm/_lib/device_buffer.cpp:413
#9  rmm::device_buffer::~device_buffer (this=0x564f70f1adc0, __in_chrg=<optimized out>)
    at rmm/_lib/device_buffer.cpp:253
#10 std::default_delete<rmm::device_buffer>::operator() (this=<optimized out>, __ptr=0x564f70f1adc0)
    at rmm/_lib/device_buffer.cpp:76
#11 std::unique_ptr<rmm::device_buffer, std::default_delete<rmm::device_buffer> >::~unique_ptr (this=0x7f9da9c8df48,
    __in_chrg=<optimized out>) at rmm/_lib/device_buffer.cpp:236
#12 __Pyx_call_destructor<std::unique_ptr<rmm::device_buffer, std::default_delete<rmm::device_buffer> > > (x=...)
    at rmm/_lib/device_buffer.cpp:334
#13 __pyx_tp_dealloc_3rmm_4_lib_13device_buffer_DeviceBuffer (o=0x7f9da9c8df30) at rmm/_lib/device_buffer.cpp:20768
#14 0x0000564f6bb293c2 in list_dealloc (op=0x7f9d9d3b2dc0)
    at /home/conda/feedstock_root/build_artifacts/python_1583419343668/work/Objects/listobject.c:324
#15 0x0000564f6bb15428 in cell_dealloc (op=0x7f9dc39335d0)
    at /home/conda/feedstock_root/build_artifacts/python_1583419343668/work/Objects/cellobject.c:52
#16 0x0000564f6bb308dc in frame_dealloc (f=0x564f6c99e0c0)
    at /home/conda/feedstock_root/build_artifacts/python_1583419343668/work/Objects/frameobject.c:470
#17 _PyEval_EvalCodeWithName ()
    at /home/conda/feedstock_root/build_artifacts/python_1583419343668/work/Python/ceval.c:3946
#18 0x0000564f6bb80527 in _PyFunction_FastCallKeywords ()
    at /home/conda/feedstock_root/build_artifacts/python_1583419343668/work/Objects/call.c:433
#19 0x0000564f6bbe8806 in call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>)
    at /home/conda/feedstock_root/build_artifacts/python_1583419343668/work/Python/ceval.c:4616
```

It's not clear what's the cause of the issue, but it clearly relates to RMM and IB when both are used together.

cc @quasiben @jrhemstad @harrism @Akshay-Venkatesh


harrism:
@pentschev could you use a debug build of RMM to get a more accurate call stack at the point where it is in the RMM C++ code? This would also allow stepping in the debugger...

pentschev:
Could you be more specific to what would be a more accurate call stack? We can step in with that build, with a debug build I don't see any significant difference though, as optimized values seem to be due to inlining.

pentschev:
In this case here, I realized the issue is that we store RMM objects in `l` (https://github.com/rapidsai/ucx-py/pull/457/files#diff-17c6de8e1db1dd4d3eb6f8b14fb4303eR31) and in every loop we `rmm.reinitialize()` (https://github.com/rapidsai/ucx-py/pull/457/files#diff-17c6de8e1db1dd4d3eb6f8b14fb4303eR38), which is obviously invalid. Nevertheless, the error seems to be the same as we see when running a Dask workflow, so I'm now thinking something similar must be happening. Another thing I can't explain is why this only happens when InfiniBand is used, but not with TCP over UCX.

pentschev:
This is not relevant anymore, closing.

PR #462, opened 2020-03-25
Use Cython's `new_build_ext` (if available)

jakirkham:
Try to use Cython's `new_build_ext` (planned to replace it's current `build_ext`) in order to benefit from features (like parallel builds). Though maintain a fallback to Setuptools in cases where Cython might not be present as needed for building (like checking the version). Note Setuptools would like to adopt this behavior in the future, but hasn't yet ( https://github.com/pypa/setuptools/issues/1270 ).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,10 @@ import versioneer
 from setuptools import setup
 from setuptools.extension import Extension
 
-from setuptools.command.build_ext import build_ext
+try:
+    from Cython.Distutils.build_ext import new_build_ext as build_ext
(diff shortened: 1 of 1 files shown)


quasiben:
Does this require any changes to the recipe ?  

https://github.com/rapidsai/ucx-split-feedstock/blob/master/recipe/meta.yaml

jakirkham:
Nope. Cython was already a build requirement.

quasiben:
Thanks John.  Merging

Issue #468, opened 2020-03-30
[BUG] Error in sys.excepthook on process exit


randerzander:
In the latest conda nightlies, when terminating a Python process in which a UCX client connection was established, a confusing error is printed:
```
>>> from dask.distributed import Client
>>> client = Client('ucx://10.150.162.155:8786')
>>> client
<Client: 'ucx://10.150.162.155:8786' processes=136 threads=136, memory=1.02 TB>
>>> exit()
Error in sys.excepthook:

Original exception was:
```

This does not happen for TCP clusters.

quasiben:
Thanks @randerzander .  I was able to confirm the issue.

jakirkham:
Was this solved by PR ( https://github.com/rapidsai/ucx-py/pull/469 )?

quasiben:
yes, i believe.  @randerzander if you are still seeing the issue, please re-open.

PR #469, opened 2020-03-30
AppContext use finalizer instead of __dealloc__()

madsbk:
Fixes issue #468 by using a `weakref.finalize` instead of `ApplicationContext.__dealloc__()`.


quasiben:
Thanks @madsbk

ucp/_libs/core.pyx
```diff
@@ -175,13 +175,26 @@ async def listener_handler(endpoint, ctx, func, guarantee_msg_order):
...
+        func(pub_ep)
+
+
+def application_context_finalizer(children, worker, context, epoll_fd):
+    """
+    Finalizer function for `ApplicationContext` object, which is
+    more reliable than __dealloc__.
+    """
+    for child in children:
+        child = child()
```
jakirkham: Should we disambiguate the name here somehow? It's not totally clear (at least not to me) what's happening

ucp/continuous_ucx_progress.py
```diff
@@ -7,21 +7,21 @@
...
-        not prevent `ctx` to be garbage collected.
+        reference to `worker` so that a danling progress task will
+        not prevent `worker` to be garbage collected.
 
         Parameters
         ----------
-        ctx: ApplicationContext
-            The application context to progress
+        worker: UCXWorker
+            The UCX worker context to progress
```
jakirkham: Is this breaking API? If so, are there consequences for downstream users (like Distributed or cuML)?

madsbk: This is not part of the public API and AFAIK nobody are using it downstream.

ucp/_libs/core.pyx
madsbk: Good point, will fix it :)

madsbk: Furthermore as part of the implementation of https://github.com/rapidsai/ucx-py/pull/418, I will streamline the cleanup processes.

ucp/_libs/ucx_api.pyx
```diff
@@ -100,15 +100,17 @@ def get_ucx_version():
 cdef class UCXContext:
     """Python representation of `ucp_context_h`"""
     cdef:
+        object __weakref__
         ucp_context_h _handle
-        bint _initialized
         dict _config
+    cdef public:
+        bint initialized
```
pentschev: Why `_initialized` has to be made public?

ucp/_libs/ucx_api.pyx
```diff
@@ -139,39 +141,43 @@ cdef class UCXContext:
...
 
 cdef class UCXWorker:
     """Python representation of `ucp_worker_h`"""
     cdef:
+        object __weakref__
         ucp_worker_h _handle
-        bint _initialized
         UCXContext _context
+    cdef public:
+        bint initialized
```
pentschev: Same question as in `UCXContext`.

ucp/_libs/ucx_api.pyx
```diff
@@ -240,10 +251,11 @@ cdef class UCXEndpoint:
     Please use `ucx_ep_create()` to contruct an instance of this class
     """
     cdef:
+        object __weakref__
         ucp_ep_h _handle
-        bint initialized
 
     cdef public:
+        bint initialized
```
pentschev: And here as well.

ucp/_libs/ucx_api.pyx
madsbk: It is used by [`ProgressTask()`](https://github.com/madsbk/ucx-py/blob/app_ctx_finalizer/ucp/continuous_ucx_progress.py#L9).

pentschev: This is probably more semantic, but it seems that it's used read-only, right? Would it make sense to still prefix it with `_` to identify that shouldn't be modified? Maybe a `isInitialized` method would be even better.

jakirkham: Well if it's `readonly`, Cython has a way to [declare that]( https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#readonly ). So maybe we should just mark it as `readonly`?

jakirkham:
PR ( https://github.com/rapidsai/ucx-py/pull/471 ) should fix the style issue.

jakirkham:
rerun tests

ucp/_libs/ucx_api.pyx
jakirkham: Trying to address with PR ( https://github.com/rapidsai/ucx-py/pull/472 ). Please take a look 🙂

PR #472, opened 2020-03-31
Mark `initialized` as `readonly`

jakirkham:
As it sounds like `initialized` should be visible externally, but should not be modifiable. This makes makes it `readonly` so that it can be checked in external usage, but not changed.

cc @pentschev @madsbk


jakirkham:
Agreed let's wait for Mads. 🙂

PR #477, opened 2020-04-03
Endpoint: shutdown peer once

madsbk:
This PR makes sure that an endpoint only tells peer to shutdown once.

This PR should also fix the close issue seen in https://github.com/rapidsai/ucx-py/pull/473. 
@quasiben, can you to confirm that you don't see the error?:

    File "ucp/_libs/core.pyx", line 640, in close
    AttributeError: 'NoneType' object has no attribute 'progress'

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -372,6 +372,7 @@ class _Endpoint:
         self._recv_count = 0  # Number of calls to self.recv()
         self._finished_recv_count = 0  # Number of returned (finished) self.recv() calls
         self._closed = False
+        self._shutting_down_peer = False  # Told peer to shutdown down
         self.pending_msg_list = []
         # UCX supports CUDA if "cuda" is part of the TLS or TLS is "all"
         tls = ctx.get_config()["TLS"]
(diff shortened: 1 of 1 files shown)


ucp/_libs/core.pyx
```diff
@@ -403,10 +404,15 @@ class _Endpoint:
         cdef CtrlMsg msg
         cdef CtrlMsgData[::1] ctrl_msg_mv
         try:
+            # Making sure we only tell peer to shutdown once
+            if self._shutting_down_peer:
+                return
+            self._shutting_down_peer = True
```
pentschev, line 410: Maybe it would make more sense to move this block to the top of the function:

```python
if self._closed or self._shutting_down_peer:
    return
self._shutting_down_peer = True
```

madsbk: It is there to make sure we run:
```python
    finally:
            self._ctx.worker.progress()
            await asyncio.sleep(0)
            self.abort()
```

quasiben:
I'll try out this PR tomorrow morning and confirm

ucp/_libs/core.pyx
pentschev: Interesting, it seems logical now but I didn't realize `finally` would get executed even if there's a return inside the `try` block. Thanks for clarifying @madsbk !

PR #479, opened 2020-04-08
Make logger formatting similar to UCX

pentschev:
This PR adds more timestamp, hostname and PID to logger, making it similar to UCX's logger format.

This is how logger format used to be:

```
DEBUG [Recv #000] ep: 0x7f5bccd83070, tag: 0x130180117a7eba2c, nbytes: 8
```

And how it looks with this PR:

```
[1586355277.471017] [dgx12:33950] UCXPY  DEBUG [Recv #000] ep: 0x7f5bccd83070, tag: 0x130180117a7eba2c, nbytes: 8
```

### DIFF
--- a/ucp/utils.py
+++ b/ucp/utils.py
@@ -1,5 +1,6 @@
 import asyncio
 import fcntl
+import logging
 import multiprocessing as mp
 import os
 import socket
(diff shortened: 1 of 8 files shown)


ucp/utils.py
```diff
@@ -80,6 +81,50 @@ def get_closest_net_devices(gpu_dev):
...
+
+    _level_enum = logging.getLevelName(os.getenv("UCXPY_LOG_LEVEL", "WARNING"))
+    logger = logging.getLogger("ucx")
+
+    # Avoid duplicate logging
+    logger.propagate = False
+
+    class LoggingFilter(logging.Filter):
+
+        def filter(self, record):
```
quasiben, line 108: Can you elaborate on why we need this filter ?  It's always returning `True` so is it filtering anything ?

pentschev: it returns `True` because we want to apply it for all logging messages. It could return `False` depending on some condition to avoid applying that to certain messages. For reference: https://docs.python.org/3/library/logging.html#logging.Logger.filter .

quasiben:
Thanks @pentschev this looks good.  I created a PR to add a logging test -- feel free to use it or write your own but I think we should also test logging:

https://github.com/pentschev/ucx-py/pull/1

pentschev:
Thanks for taking the time to add tests @quasiben , much appreciated!

pentschev:
rerun tests

PR #481, opened 2020-04-10
Set rsock/wsock to blocking

### LINKED ISSUE #480: Non-blocking Socket error
@pentschev recently noticed that when starting dask-cuda-workers with `python -X` we get a backtrace with

```
  File "/datasets/pentschev/miniconda3/envs/ucx-src-101-0.14.0b200408/lib/python3.7/site-packages/ucp/continuous_ucx_progress.py", line 88, in _fd_reader_callback
    self.asyncio_task = self.event_loop.create_task(self._arm_worker())
  File "/datasets/pentschev/miniconda3/envs/ucx-src-101-0.14.0b200408/lib/python3.7/site-packages/ucp/continuous_ucx_progress.py", line 106, in _arm_worker
    await self.event_loop.sock_recv(self.rsock, 1)
  File "/datasets/pentschev/miniconda3/envs/ucx-src-101-0.14.0b200408/lib/python3.7/asyncio/selector_events.py", line 355, in sock_recv
    raise ValueError("the socket must be non-blocking")
ValueError: the socket must be non-blocking
```

Writing `self.rsock.setblocking(0)` and `wsock.setblocking(0)` resolves the issue.  This is not something I would have expected as I didn't think we were creating non-blocking sockets anywhere.  @madsbk is there any danger of ucx-py filling up the Python socket buffer ?  

- [Setting Socket Buffer Size](https://stackoverflow.com/questions/30888397/how-to-set-send-buffer-size-for-sockets-in-python)

pentschev:
It's still unclear whether this is the right solution or if it impacts performance, but it potentially resolves #480 . When running Dask+UCX-Py with `python -X dev` we keep seeing `ValueError: the socket must be non-blocking` being thrown during `BlockingMode._fd_reader_callback`, which calls `ucp_worker_progress` .

### DIFF
--- a/ucp/continuous_ucx_progress.py
+++ b/ucp/continuous_ucx_progress.py
@@ -63,6 +63,8 @@ class BlockingMode(ProgressTask):
         # all non-IO tasks are finished.
         # See <https://stackoverflow.com/a/48491563>.
         self.rsock, wsock = socket.socketpair()
+        self.rsock.setblocking(0)
+        wsock.setblocking(0)
         wsock.close()
 
         # Bind an asyncio reader to a UCX epoll file descripter


quasiben:
@pentschev is there any more work to be done on this PR ?  @madsbk do you have time to review ?

madsbk:
Will take a look tomorrow and merge if everything is well :)

pentschev:
Yeah, it would be good to check it's not breaking anything else, it doesn't seem that way though. Also I didn't check performance, this is something that would be worth testing just in case.

pentschev:
Ok, I'm merging this in for now. I also don't think there's any performance issues with that.

PR #485, opened 2020-04-14
Implementing exchange_peer_info() using struct

madsbk:
Implementing `exchange_peer_info()` using the `struct` module instead of cython's struct.

This PR is part of the effort to restrict the use of Cython to `ucx_api.pyx`


madsbk:
> LGTM. I believe there's no change in behavior, but asking just in case: is this only a cleanup?

Yes, this is only cleanup :)

PR #486, opened 2020-04-14
 Implementing CtrlMsg using struct

madsbk:
Implementing `CtrlMsg` using the `struct` module instead of cython's struct.

This PR is part of the effort to restrict the use of Cython to `ucx_api.pyx`

### DIFF
--- a/ucp/_libs/core.pyx
+++ b/ucp/_libs/core.pyx
@@ -11,6 +11,7 @@ from random import randint
 import psutil
 import uuid
 import socket
+import struct
 import logging
 from os import close as close_fd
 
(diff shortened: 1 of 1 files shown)


madsbk:
> LGTM. There's also no change in behavior here, right?

Yes, this is only cleanup :)

madsbk:
Yes, my plan is to make core.pyx pure python and merge it with public_api.

jakirkham:
Sounds good. Thanks for working on this Mads 😄

ucp/_libs/core.pyx
```diff
@@ -75,52 +76,62 @@ async def exchange_peer_info(endpoint, msg_tag, ctrl_tag, guarantee_msg_order):
...
+    """
 
-cdef class CtrlMsg:
-    cdef CtrlMsgData data
+    fmt = "QQ"
+    nbytes = struct.calcsize(fmt)
 
+    @staticmethod
+    def serialize(opcode, close_after_n_recv):
+        return struct.pack(CtrlMsg.fmt, int(opcode), int(close_after_n_recv))
```
quasiben, line 92: I know this is already merged but wanted to say thanks to @madsbk for this work.  I raised some concerns in a [previous PR](https://github.com/rapidsai/ucx-py/pull/418#discussion_r389231002) (probably wrong) and I appreciate you taking a look.

jakirkham: Yeah Mads is right that it should have been `"i"` in that context. Though it seems we have dropped that value since.

PR #488, opened 2020-04-15
Renaming core.pyx => core.py

madsbk:
Renaming `core.pyx` => `core.py`. No change to behavior.

### DIFF
--- a/tests/test_guarantee_msg_order.py
+++ b/tests/test_guarantee_msg_order.py
@@ -9,7 +9,7 @@ def test_mismatch(server_guarantee_msg_order):
 
     # We use an exception handle to catch errors raised by the server
     def handle_exception(loop, context):
-        msg = context.get("exception", context["message"])
+        msg = str(context.get("exception", context["message"]))
         loop.test_failed = msg.find(loop.error_msg_expected) == -1
 
     loop = asyncio.get_event_loop()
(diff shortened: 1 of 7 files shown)


tests/test_guarantee_msg_order.py
```diff
@@ -9,7 +9,7 @@ def test_mismatch(server_guarantee_msg_order):
 
     # We use an exception handle to catch errors raised by the server
     def handle_exception(loop, context):
-        msg = context.get("exception", context["message"])
+        msg = str(context.get("exception", context["message"]))
```
jakirkham, line 12: Should we change how `context["exception"]` and `context["message"]` are set to ensure they are `stress?

madsbk: I am not sure if that is possible. It is defined here: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_exception_handler

jakirkham: Oh in that case wouldn't they already be `str`s?

madsbk: `context["exception"]` is a `ValueError`. I use `str()` to assert the message of the error.

pentschev:
@rapidsai/ops-codeowners I think you may need to approve this due to the change in `.flake8`. Could you check?

PR #493, opened 2020-04-16
Remove asyncio exception handler

madsbk:
We don't need the asyncio error handler anymore. Now it only seems to hide debug information.


quasiben:
@madsbk, FYI https://github.com/rapidsai/ucx-py/pull/488 has been merged in now

jakirkham:
Probably was some edge case with Cython.

PR #494, opened 2020-04-16
Fix error logging when closing peer endpoint

### LINKED ISSUE #491: Repeated and frequent shutdown and creation of UCX cluster causes IO errors
When I shutdown/create a dask cluster with UCX within 5-15 seconds of another one I seem to error quite a bit more than I used to in the past. This makes it hard to run consecutive workflows that include bringing up / spinning down a LocalCUDACluster.

Routinely the error is something like:
```
File "ucp/_libs/core.pyx", line 426, in ucp._libs.core._Endpoint.close
ucp.exceptions.UCXError: Comm Error "[Send shutdown] ep: 0x7fc1b9bc3728, tag: 0x17fc61825a8b67ec, close_after_n_recv: 4": Input/output error
```

which causes the following through dask’s dispatching:
```
  File "/raid/nicholasb/miniconda3/envs/rapids--20200415-cuda101-1112/lib/python3.7/site-packages/ucp/public_api.py", line 270, in close
    await self._ep.close()
  File "ucp/_libs/core.pyx", line 435, in close
ValueError: Unknown format code 'd' for object of type 'str'
```

This may be related to an issue @pentschev @quasiben and @madsbk were discussing earlier. Will provide a full traceback in a comment.

madsbk:
This PR fixes #491. I am working on testing this case.

@beckernick can I get you to see if this helps?

### DIFF
--- a/tests/test_disconnect.py
+++ b/tests/test_disconnect.py
@@ -1,0 +1,71 @@
+import asyncio
+import logging
+import multiprocessing as mp
+from io import StringIO
+from queue import Empty
+
+import numpy as np
+import ucp
+
+mp = mp.get_context("spawn")
+
+
+async def mp_queue_get_nowait(queue):
+    while True:
+        try:
+            return queue.get_nowait()
+        except Empty:
+            pass
+        await asyncio.sleep(0.01)
+
+
+def _test_shutdown_closed_peer_server(client_queue, server_queue):
+    async def run():
+        async def server_node(ep):
+            try:
+                await ep.send(np.arange(100, dtype=np.int64))
+                # Waiting for signal to close the endpoint
+                await mp_queue_get_nowait(server_queue)
+                await ep.close()
+            finally:
+                listener.close()
+
+        listener = ucp.create_listener(server_node)
+        client_queue.put(listener.port)
+        while not listener.closed():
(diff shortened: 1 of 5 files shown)


pentschev:
I'm not totally sure, but I think the error in #491 is more likely related to the shutdown behavior we discussed offline yesterday, or is the logger change fixing that?

madsbk:
> I'm not totally sure, but I think the error in #491 is more likely related to the shutdown behavior we discussed offline yesterday, or is the logger change fixing that?

This PR fixes the error:
``` 
  File "/raid/nicholasb/miniconda3/envs/rapids--20200415-cuda101-1112/lib/python3.7/site-packages/ucp/public_api.py", line 270, in close
    await self._ep.close()
  File "ucp/_libs/core.pyx", line 435, in close
ValueError: Unknown format code 'd' for object of type 'str'
```
And downgrading the log level to warning.

jakirkham:
That's right. This fixes the log message. It may not fix why we see the log message, but it does let us see it, which should give us more insight 🙂

jakirkham:
Thanks Mads! 😄 Was just eyeing that green button 😉

pentschev:
Sorry, I'm late here. LGTM, also that was a very clever test you added!

PR #496, opened 2020-04-20
Disabling guarantee_msg_order by default

madsbk:
This PR disable `guarantee_msg_order` by default.
The message order is guaranteed by UCX itself so is shouldn't be needed in UCX-Py.

Before we merge, let's make sure this doesn't break downstream libraries.

cc. @pentschev, @quasiben

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -745,11 +745,11 @@ def get_config():
         return _get_ctx().get_config()
 
 
-def create_listener(callback_func, port=None, guarantee_msg_order=True):
+def create_listener(callback_func, port=None, guarantee_msg_order=False):
     return _get_ctx().create_listener(callback_func, port, guarantee_msg_order)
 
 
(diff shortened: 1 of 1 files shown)


pentschev:
> Before we merge, let's make sure this doesn't break downstream libraries.

I've recently tested with both `True` and `False` and I haven't noticed any differences. However, I don't really know how we can ensure this doesn't break downstream libraries.

jakirkham:
> Before we merge, let's make sure this doesn't break downstream libraries.

@cjnolet @MattBBaker, would you be able to take a look?

MattBBaker:
All of my work was passing in `false` for this option, so 👍

cjnolet:
My apologies, I didn’t see this come through my notifications for some reason. Thank you for making me aware of this change! I dont believe this will have an effect on cuml.

PR #498, opened 2020-04-20
Streamlining the destructors in the UCX API

madsbk:
This PR make sure that all calls to the UCX is using valid UCX handles and streamlines the destruction of UCX handles.

### DIFF
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -381,18 +381,21 @@ cdef void _listener_callback(ucp_ep_h ep, void *args):
         )
 
 
-cdef class UCXListener:
+def _ucx_listener_handle_finalizer(handle_as_int):
+    cdef ucp_listener_h handle = <ucp_listener_h><uintptr_t> handle_as_int
+    ucp_listener_destroy(handle)
+
+
+cdef class UCXListener(UCXObject):
     """Python representation of `ucp_listener_h`"""
     cdef:
-        object __weakref__
-        ucp_listener_h _ucp_listener
-        object _ctx
+        ucp_listener_h _handle
+        dict cb_data
 
     cdef public:
         int port
-        dict cb_data
 
-    def __init__(self, port, ctx, cb_data):
+    def __init__(self, UCXWorker worker, port, cb_data):
         cdef ucp_listener_params_t params
         cdef ucp_listener_accept_callback_t _listener_cb = (
             <ucp_listener_accept_callback_t>_listener_callback
@@ -406,19 +409,21 @@ cdef class UCXListener:
             raise MemoryError("Failed allocation of ucp_ep_params_t")
 
         cdef ucs_status_t status = ucp_listener_create(
-            <ucp_worker_h><uintptr_t>ctx.worker.handle, &params, &self._ucp_listener
+            worker._handle, &params, &self._handle
         )
         c_util_get_ucp_listener_params_free(&params)
         assert_ucs_status(status)
-        Py_INCREF(self.cb_data)
-        self._ctx = ctx
 
-    def abort(self):
-        if self._ctx is not None:
-            ucp_listener_destroy(self._ucp_listener)
-            Py_DECREF(self.cb_data)
-            self._ctx = None
-            self.cb_data = None
+        self.add_handle_finalizer(
+            _ucx_listener_handle_finalizer,
+            int(<uintptr_t><void*>self._handle)
+        )
+        worker.add_child(self)
+
+    @property
+    def handle(self):
+        assert self.initialized
+        return int(<uintptr_t><void*>self._handle)
 
 
 cdef create_future_from_comm_status(ucs_status_ptr_t status,
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -218,7 +218,8 @@ cdef class UCXContext(UCXObject):
         return int(<uintptr_t><void*>self._handle)
 
 
-def _ucx_worker_handle_finalizer(uintptr_t handle_as_int):
+def _ucx_worker_handle_finalizer(uintptr_t handle_as_int, UCXContext ctx):
+    assert ctx.initialized
     cdef ucp_worker_h handle = <ucp_worker_h>handle_as_int
     ucp_worker_destroy(handle)
 
@@ -243,7 +244,8 @@ cdef class UCXWorker(UCXObject):
 
         self.add_handle_finalizer(
             _ucx_worker_handle_finalizer,
-            int(<uintptr_t><void*>self._handle)
+            int(<uintptr_t><void*>self._handle),
+            self._context
         )
         context.add_child(self)
 
@@ -314,6 +316,7 @@ cdef class UCXWorker(UCXObject):
 
 
 def _ucx_endpoint_finalizer(uintptr_t handle_as_int, worker, inflight_msgs):
+    assert worker.initialized
     cdef ucp_ep_h handle = <ucp_ep_h>handle_as_int
     cdef ucs_status_ptr_t status
 
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -230,6 +230,7 @@ cdef class UCXWorker(UCXObject):
         cdef ucp_params_t ucp_params
         cdef ucp_worker_params_t worker_params
         cdef ucs_status_t status
+        assert context.initialized
         self._context = context
         memset(&worker_params, 0, sizeof(worker_params))
         worker_params.field_mask = UCP_WORKER_PARAM_FIELD_THREAD_MODE
@@ -333,6 +334,8 @@ cdef class UCXEndpoint(UCXObject):
 
     cdef _init(self, UCXWorker worker, ucp_ep_h handle):
         """The Constructor"""
+
+        assert worker.initialized
         self.worker = worker
         self._handle = handle
         self.add_handle_finalizer(
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -103,12 +103,12 @@ def get_ucx_version():
     return (a, b, c)
 
 
-def _handle_finalizer_wrapper(children, handle_finalizer, handle_as_int):
+def _handle_finalizer_wrapper(children, handle_finalizer, handle_as_int, *extra_args):
     for weakref_to_child in children:
         child = weakref_to_child()
         if child is not None:
             child.close()
-    handle_finalizer(handle_as_int)
+    handle_finalizer(handle_as_int, *extra_args)
 
 
 cdef class UCXObject:
@@ -147,14 +147,15 @@ cdef class UCXObject:
         """
         self._children.append(weakref.ref(child))
 
-    def add_handle_finalizer(self, handle_finalizer, handle_as_int):
+    def add_handle_finalizer(self, handle_finalizer, handle_as_int, *extra_args):
         """Add a finalizer of `handle_as_int`"""
         self._finalizer = weakref.finalize(
             self,
             _handle_finalizer_wrapper,
             self._children,
             handle_finalizer,
-            handle_as_int
+            handle_as_int,
+            *extra_args
         )
 
 
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -199,7 +199,7 @@ cdef class UCXContext(UCXObject):
 
         self.add_handle_finalizer(
             _ucx_context_handle_finalizer,
-            int(<uintptr_t><void*>self._handle)
+            int(<uintptr_t>self._handle)
         )
 
         self._config = ucx_config_to_dict(config)
@@ -215,7 +215,7 @@ cdef class UCXContext(UCXObject):
     @property
     def handle(self):
         assert self.initialized
-        return int(<uintptr_t><void*>self._handle)
+        return int(<uintptr_t>self._handle)
 
 
 def _ucx_worker_handle_finalizer(uintptr_t handle_as_int, UCXContext ctx):
@@ -244,7 +244,7 @@ cdef class UCXWorker(UCXObject):
 
         self.add_handle_finalizer(
             _ucx_worker_handle_finalizer,
-            int(<uintptr_t><void*>self._handle),
+            int(<uintptr_t>self._handle),
             self._context
         )
         context.add_child(self)
@@ -291,7 +291,7 @@ cdef class UCXWorker(UCXObject):
     @property
     def handle(self):
         assert self.initialized
-        return int(<uintptr_t><void*>self._handle)
+        return int(<uintptr_t>self._handle)
 
     def request_cancel(self, ucp_request_as_int):
         assert self.initialized
@@ -312,7 +312,7 @@ cdef class UCXWorker(UCXObject):
         cdef ucs_status_t status = ucp_ep_create(self._handle, &params, &ucp_ep)
         c_util_get_ucp_ep_params_free(&params)
         assert_ucs_status(status)
-        return UCXEndpoint(self, <uintptr_t><void*> ucp_ep)
+        return UCXEndpoint(self, <uintptr_t>ucp_ep)
 
 
 def _ucx_endpoint_finalizer(uintptr_t handle_as_int, worker, inflight_msgs):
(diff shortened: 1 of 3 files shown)


madsbk:
cc. @beckernick @VibhuJawa

quasiben:
Running with this PR i am seeing the following error:

```
Task exception was never retrieved
future: <Task finished coro=<_listener_handler() done, defined at /gpfs/fs1/bzaitlen/miniconda3/envs/20200417/lib/python3.7/site-packages/ucp/core.py:116> exception=ValueError('Both peers must set guarantee_msg_order identically')>
  File "/gpfs/fs1/bzaitlen/miniconda3/envs/20200417/lib/python3.7/site-packages/ucp/core.py", line 127, in _listener_handler
    guarantee_msg_order=guarantee_msg_order,
  File "/gpfs/fs1/bzaitlen/miniconda3/envs/20200417/lib/python3.7/site-packages/ucp/core.py", line 56, in exchange_peer_info
    raise ValueError("Both peers must set guarantee_msg_order identically")
ValueError: Both peers must set guarantee_msg_order identically
```

This is in the middle of a workflow

quasiben:
Apologies, this is with an IB stress test with 4 nodes: https://gist.github.com/quasiben/73cf6d7c2131fe41370014ddc21ecb56

Perhaps this is still the bug @pentschev, myself, and others have been tracking

pentschev:
I've done some testing with this and I have some carefully optimistic news:

1. TCP: not a single hang or crash in about 3 runs;
1. NVLink: not a single hang or crash in about 3 runs;
1. IB: not a single hang or crash in 10+ runs;
1. IB+NVLink: in some 7-8 runs, it hanged all times except for two consecutive ones.

All tests above were on a 4 node/32 GPU cluster.

I'm carefully optimistic because IB was hanging and crashing most of the time and I haven't been able to reproduce that in any runs. The amount of observation is still not enough to ensure our bugs have been fixed. The hangs with IB+NVLink also point to some remaining issues.

Finally, just as the merge operation starts I still the following error several times when new endpoints are getting created to do worker-worker connection:

```python
future: <Task finished name='Task-1578' coro=<_listener_handler() done, defined at /datasets/pentschev/miniconda3/envs/pydbg/lib/python3.8/site-packages/ucp/core.py:116> exception=ValueError('Both peers must set guarantee_msg_order identically')>
    raise ValueError("Both peers must set guarantee_msg_order identically")
ValueError: Both peers must set guarantee_msg_order identically
```

For @VibhuJawa and @beckernick , please mind the NVLink+IB hang if you're testing. If you have the chance a test with IB only (NVLink disabled) would be useful as well to see whether you experience any hangs or segfaults.

pentschev:
I intend to do some more testing tomorrow to confirm whether I can't reproduce IB errors anymore.

ucp/_libs/ucx_api.pyx
```diff
@@ -102,21 +103,77 @@ def get_ucx_version():
...
 
 
-cdef class UCXContext:
-    """Python representation of `ucp_context_h`"""
+def _handle_finalizer_wrapper(children, handle_finalizer, handle_as_int, *extra_args):
+    for weakref_to_child in children:
+        child = weakref_to_child()
+        if child is not None:
+            child.close()
+    handle_finalizer(handle_as_int, *extra_args)
```
jakirkham, line 111: Maybe support other keyword arguments as well?

```suggestion
def _handle_finalizer_wrapper(children, handle_finalizer, handle_as_int, *extra_args, **extra_kargs):
    for weakref_to_child in children:
        child = weakref_to_child()
        if child is not None:
            child.close()
    handle_finalizer(handle_as_int, *extra_args, **extra_kargs)
```

ucp/_libs/ucx_api.pyx
```diff
@@ -102,21 +103,77 @@ def get_ucx_version():
...
+        self._finalizer = None
+        # List of weak references of UCX objects that make use of this object
+        self._children = []
+
+    def close(self):
+        """Close the object and free the underlying UCX handle.
+        Does nothing if the object is already closed
+        """
+        if self.initialized:
+            self._finalizer()
```
jakirkham, line 137: Maybe `_finalizer` should just be expected to be idempotent. That way if we call it more than once nothing happens. WDYT?

ucp/_libs/ucx_api.pyx
```diff
@@ -160,28 +216,34 @@ cdef class UCXContext:
         return int(<uintptr_t><void*>self._handle)
 
 
-cdef class UCXWorker:
+def _ucx_worker_handle_finalizer(handle_as_int):
+    cdef ucp_worker_h handle = <ucp_worker_h><uintptr_t> handle_as_int
```
jakirkham, line 220: We can just require `uintptr_t` typed value (even when interacting with Python) and it will add proper error handling for us at the outset. Plus we can just use the variable knowing it is `uintptr_t` type.

```suggestion
def _ucx_worker_handle_finalizer(uintptr_t handle_as_int):
    cdef ucp_worker_h handle = <ucp_worker_h>handle_as_int
```

ucp/_libs/ucx_api.pyx
```diff
@@ -254,39 +311,51 @@ cdef class UCXWorker:
         return ucx_ep_create(ucp_ep, self)
 
 
-cdef class UCXEndpoint:
+def _ucx_endpoint_finalizer(handle_as_int, worker, inflight_msgs):
+    cdef ucp_ep_h handle = <ucp_ep_h><uintptr_t> handle_as_int
```
jakirkham, line 315:
```suggestion
def _ucx_endpoint_finalizer(uintptr_t handle_as_int, worker, inflight_msgs):
    cdef ucp_ep_h handle = <ucp_ep_h>handle_as_int
```

ucp/_libs/ucx_api.pyx
```diff
@@ -254,39 +311,51 @@ cdef class UCXWorker:
...
-        bint initialized
-
-    cdef public:
         UCXWorker worker
 
-    def __cinit__(self, worker):
-        self._handle = NULL
-        self.worker = worker
-        self.initialized = False
+    cdef _init(self, UCXWorker worker, ucp_ep_h handle):
```
jakirkham, line 345: Why not use `__cinit__` or perhaps `__init__` for this?

ucp/_libs/ucx_api.pyx
```diff
@@ -314,16 +383,14 @@ cdef class UCXEndpoint:
...
 cdef UCXEndpoint ucx_ep_create(ucp_ep_h ep, UCXWorker worker):
-    ret = UCXEndpoint(worker)
-    ret._handle = ep
-    ret.initialized = True
+    ret = UCXEndpoint()
+    ret._init(worker, ep)
     return ret
 
 
 def ucx_ep_create_from_uintptr(ep, worker):
```
jakirkham, line 391:
```suggestion
def ucx_ep_create_from_uintptr(uintptr_t ep, worker):
```

ucp/_libs/ucx_api.pyx
```diff
@@ -314,16 +383,14 @@ cdef class UCXEndpoint:
...
-    ret = UCXEndpoint(worker)
-    ret._handle = <ucp_ep_h><uintptr_t>ep
-    ret.initialized = True
+    ret = UCXEndpoint()
+    ret._init(worker, <ucp_ep_h><uintptr_t>ep)
```
jakirkham, line 393:
```suggestion
    ret._init(worker, <ucp_ep_h>ep)
```

ucp/_libs/ucx_api.pyx
```diff
@@ -342,18 +409,21 @@ cdef void _listener_callback(ucp_ep_h ep, void *args):
         )
 
 
-cdef class UCXListener:
+def _ucx_listener_handle_finalizer(handle_as_int):
+    cdef ucp_listener_h handle = <ucp_listener_h><uintptr_t> handle_as_int
```
jakirkham, line 413:
```suggestion
def _ucx_listener_handle_finalizer(uintptr_t handle_as_int):
    cdef ucp_listener_h handle = <ucp_listener_h>handle_as_int
```

ucp/core.py
```diff
@@ -405,10 +384,11 @@ def port(self):
...
         """Closing the listener"""
-        if not self._closed:
-            self._b.abort()
-            self._closed = True
-            self._b = None
+        self._b.close()
+
+    def abort(self):
+        """Closing the listener"""
+        self._b.close()
```
jakirkham, line 391: Why do we need both `abort` and `close` if they do the same thing?

madsbk: Removed :)
(Used `abort()` in a intermediate develop step)

ucp/_libs/ucx_api.pyx
madsbk: Cython doesn't support a `cdef` constructor so it complains that `ucp_ep_h handle` isn't convertible to a `PyObject`.

ucp/_libs/ucx_api.pyx
madsbk: I did that initially but it requires a dummy finalizer class for the case when `add_handle_finalizer()` hasn't been called yet. 
I concluded that this solution was less code and more readable.

ucp/_libs/ucx_api.pyx
jakirkham: One way to workaround this limitation is take a `uintptr_t` instead (Python will allow this and perform type checking for it) and then cast to your specific pointer type.

ucp/_libs/ucx_api.pyx
```diff
@@ -254,39 +315,52 @@ cdef class UCXWorker:
...
-        bint initialized
-
-    cdef public:
         UCXWorker worker
 
-    def __cinit__(self, worker):
-        self._handle = NULL
-        self.worker = worker
-        self.initialized = False
+    cdef _init(self, UCXWorker worker, ucp_ep_h handle):
```
jakirkham, line 350:
```suggestion
    def __cinit__(self, UCXWorker worker, uintptr_t handle):
```

ucp/_libs/ucx_api.pyx
```diff
@@ -254,39 +315,52 @@ cdef class UCXWorker:
...
-                    worker.progress()
-                assert not UCS_PTR_IS_ERR(status)
-                ucp_request_free(status)
+        assert worker.initialized
+        self.worker = worker
+        self._handle = handle
+        self._inflight_msgs = dict()
+        self.add_handle_finalizer(
+            _ucx_endpoint_finalizer,
+            int(<uintptr_t><void*>handle),
```
jakirkham, line 359:
```suggestion
            int(handle),
```

ucp/_libs/ucx_api.pyx
```diff
@@ -254,39 +315,52 @@ cdef class UCXWorker:
...
-            if UCS_PTR_STATUS(status) != UCS_OK:
-                assert not UCS_PTR_IS_ERR(status)
-                # We spinlock here until `status` has finished
-                while ucp_request_check_status(status) != UCS_INPROGRESS:
-                    worker.progress()
-                assert not UCS_PTR_IS_ERR(status)
-                ucp_request_free(status)
+        assert worker.initialized
+        self.worker = worker
+        self._handle = handle
```
jakirkham, line 355:
```suggestion
        self._handle = <ucp_ep_h>handle
```

ucp/_libs/ucx_api.pyx
```diff
@@ -314,16 +388,14 @@ cdef class UCXEndpoint:
 
 
 cdef UCXEndpoint ucx_ep_create(ucp_ep_h ep, UCXWorker worker):
-    ret = UCXEndpoint(worker)
-    ret._handle = ep
-    ret.initialized = True
+    ret = UCXEndpoint()
+    ret._init(worker, ep)
```
jakirkham, line 392:
```suggestion
    ret = UCXEndpoint(worker, <uintptr_t>ep)
```

ucp/_libs/ucx_api.pyx
```diff
@@ -314,16 +388,14 @@ cdef class UCXEndpoint:
...
     return ret
 
 
-def ucx_ep_create_from_uintptr(ep, worker):
-    ret = UCXEndpoint(worker)
-    ret._handle = <ucp_ep_h><uintptr_t>ep
-    ret.initialized = True
+def ucx_ep_create_from_uintptr(uintptr_t ep, worker):
+    ret = UCXEndpoint()
+    ret._init(worker, <ucp_ep_h>ep)
```
jakirkham, line 398:
```suggestion
    ret = UCXEndpoint(worker, ep)
```

pentschev:
I did more testing today and results were very similar to what I observed yesterday. Having NVLink enabled makes the cluster more likely to hang. IB was (almost) hang-free, it did hang once in some 30 runs. I also did various runs with larger data -- multiplying the sizes of the workflow in https://github.com/rapidsai/ucx-py/issues/402#issue-556594147 by 8 -- which takes roughly 4 minutes to create the data and another 11-12 minutes for the merge to complete, and they all seemed to behave the same as the original size, hanging sometimes with NVLink and passing with IB. Due to the long wait between runs, I did 3 successful runs of each only, where I observed 2 or 3 hangs for NVLink (totalling 5-6 runs in that case).

I am very suspicious still of what happens during `exchange_peer_info`, it seems that dask-cuda-workers listener's will receive corrupted responses or not receive a response at all (not sure which one of the two is really happening), causing tags and `guarantee_msg_order` to be invalid. Perhaps this will eventually be fixed by #503, but it isn't the case at the time of writing. To be more specific, the reason I believe the hangs are connected to `exchange_peer_info` is that hangs happen just a little while after many of the workers raise https://github.com/rapidsai/ucx-py/blob/37e444587709b7c2d26c19ebcdd49789dc1c18b5/ucp/core.py#L56 , something that happens at the beginning of the merge task stream when workers are establishing connections to each other.

quasiben:
How do folks generally feel about merging this PR in given that other work is now dependent on this ?

ucp/_libs/ucx_api.pyx
jakirkham: Proposed some changes in a more recent review to address this.

jakirkham:
I made a few minor suggestions above that would be nice to see addressed first.

quasiben:
> I made a few minor suggestions above that would be nice to see addressed first.

Definitely agreed -- I wanted to get a sense where folks were at.

jakirkham:
Yeah after that's addressed I'm +1 on merging.

ucp/_libs/ucx_api.pyx
```diff
@@ -254,39 +311,51 @@ cdef class UCXWorker:
...
+
+    # Cancel all inflight messages
+    for msg in list(inflight_msgs.values()):
+        logger.debug("Future cancelling: %s" % msg["log_msg"])
+        # Notice, `request_cancel()` evoke the send/recv callback functions
+        worker.request_cancel(msg["ucp_request"])
+
+    # Close the endpoint
+    # TODO: Support UCP_EP_CLOSE_MODE_FORCE
+    status = ucp_ep_close_nb(handle, UCP_EP_CLOSE_MODE_FLUSH)
```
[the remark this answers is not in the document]
madsbk: Added https://github.com/rapidsai/ucx-py/issues/505

pentschev:
I'm +1 on merging this as well.

ucp/_libs/ucx_api.pyx
```diff
@@ -312,7 +312,7 @@ cdef class UCXWorker(UCXObject):
         cdef ucs_status_t status = ucp_ep_create(self._handle, &params, &ucp_ep)
         c_util_get_ucp_ep_params_free(&params)
         assert_ucs_status(status)
-        return ucx_ep_create(ucp_ep, self)
+        return UCXEndpoint(self, <uintptr_t><void*> ucp_ep)
```
pentschev, line 315: Is this necessary? Do you need to cast to `void*` and then cast it again to `uintptr_t` ?

pentschev: IOW, isn't it possible to directly cast once as `<uintptr_t> ucp_ep`?

madsbk:
>  am very suspicious still of what happens during `exchange_peer_info`

@pentschev I think you are on to something: https://github.com/rapidsai/ucx-py/pull/506 !

ucp/_libs/ucx_api.pyx
madsbk: Fixed

quasiben:
Thank you @mads and @jakirkham for the reviews

pentschev:
LGTM, any final comments before we merge @jakirkham ?

jakirkham:
Thanks Mads for the PR! Also thanks Peter and Ben for the reviews!

PR #503, opened 2020-04-22
Remove ucx stream

### LINKED ISSUE #497: Move away from UCX Stream API
In `exchange_peer_info` https://github.com/rapidsai/ucx-py/blob/e9a2c92a21551dfece11591dcec0a4ce24ab0a4f/ucp/core.py#L47-L50 :

We use the UCX Stream API for setting up uniquely identifiable matching tags between endpoints. It was recently suggested that we switch to the more mature ucp_tag_send/recv set up endpoints. 

This may be related to errors we first started exploring in https://github.com/rapidsai/ucx-py/issues/495

madsbk:
This PR fixes #497 by using tag send/recv to initialize the endpoints instead of steam send/recv.

cc. @pentschev @quasiben

### DIFF
--- a/tests/test_guarantee_msg_order.py
+++ b/tests/test_guarantee_msg_order.py
@@ -17,17 +17,17 @@ def test_mismatch(server_guarantee_msg_order):
     loop.test_failed = False
     loop.error_msg_expected = "Both peers must set guarantee_msg_order identically"
 
-    with pytest.raises(ValueError, match=loop.error_msg_expected):
-        lt = ucp.create_listener(
-            lambda x: x, guarantee_msg_order=server_guarantee_msg_order
+    # with pytest.raises(ValueError, match=loop.error_msg_expected):
+    lt = ucp.create_listener(
+        lambda x: x, guarantee_msg_order=server_guarantee_msg_order
+    )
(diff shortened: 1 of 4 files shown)


pentschev:
I did some testing on this PR and I see it consistently hanging already during the first task stream (creating random data). I don't see this if I'm using a single DGX-1 (8 workers), but I see it if I have at least a total of 8 workers in two DGX-1s (4 workers per node), and it also happens independent of transports used (just TCP, or also NVLink/IB as well). I'm thinking this is perhaps due to tag collision, but I'm not sure.

pentschev:
I'm not sure why but with this branch now workers connect to the scheduler briefly and disconnect almost immediately. Is this something you're still looking at @madsbk ?

quasiben:
@pentschev are you saying this works both IB+NVLink ?

pentschev:
> @pentschev are you saying this works both IB+NVLink ?

I don't understand how you reached that conclusion from what I wrote. But in the latest commits workers connect to the scheduler and almost immediately disconnect, which means we can't dispatch any work to them.

quasiben:
Yeah, I don't know how I got that so wrong.  Thanks for the clarification.

PR #504, opened 2020-04-22
Progress after endpoint create

madsbk:
UCX hangs when many clients (~100) call `ep_create()` without any call to `progress()` in between.

This PR adds a test of the case and fix the hang by calling `progress()` right after the call to `ep_create()`, which is probably a good idea in any case.


quasiben:
This looks great!  And thank you for adding a stress test here as well.  @pentschev recently brought up a note from UCX

https://github.com/openucx/ucx/blob/5c71925431166d43357a3ed1aee4fc378cb81a16/test/examples/ucp_client_server.c#L679-L685

Do you think in the context of using ucx-py in dask we are in danger of several clients are trying to connect to the listener parallel in running into issues ?

PR #506, opened 2020-04-23
Endpoint sanity check

madsbk:
Added checksum of the message exchange in `exchange_peer_info()`

cc. @pentschev

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -35,16 +35,24 @@ def _get_ctx():
 
 
 async def exchange_peer_info(
-    endpoint, msg_tag, ctrl_tag, guarantee_msg_order, listener
+    endpoint, msg_tag, ctrl_tag, guarantee_msg_order, port, listener
 ):
     """Help function that exchange endpoint information"""
 
-    msg_tag = int(msg_tag)
-    ctrl_tag = int(ctrl_tag)
-    guarantee_msg_order = bool(guarantee_msg_order)
-    my_info = struct.pack("QQ?", msg_tag, ctrl_tag, guarantee_msg_order)
+    # Pack peer information incl. a checksum
+    fmt = "QQ?QQ"
+    my_info = struct.pack(
+        fmt,
+        msg_tag,
+        ctrl_tag,
+        guarantee_msg_order,
+        port,
+        hash64bits(msg_tag, ctrl_tag, guarantee_msg_order, port),
+    )
     peer_info = bytearray(len(my_info))
 
+    # Send/recv peer information. Notice, we force an `await` between the two
+    # streaming calls (see <https://github.com/rapidsai/ucx-py/pull/509>)
     if listener is True:
         await ucx_api.stream_send(endpoint, my_info, len(my_info))
         await ucx_api.stream_recv(endpoint, peer_info, len(peer_info))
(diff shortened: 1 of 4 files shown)


pentschev:
I confirm I see many errors in Dask with this branch in all workers:

```
future: <Task finished coro=<_listener_handler() done, defined at /datasets/pentschev/miniconda3/envs/ucx-src-102-0.14.0b200423/lib/python3.7/site-packages/ucp/core.py:145> exception=UCXError('Checksum invalid! 0xde459acad26e563 != 0x7f4e28bcb0e0')>
  File "/datasets/pentschev/miniconda3/envs/ucx-src-102-0.14.0b200423/lib/python3.7/site-packages/ucp/core.py", line 78, in exchange_peer_info
ucp.exceptions.UCXError: Checksum invalid! 0xde459acad26e563 != 0x7f4e28bcb0e0
future: <Task finished coro=<_listener_handler() done, defined at /datasets/pentschev/miniconda3/envs/ucx-src-102-0.14.0b200423/lib/python3.7/site-packages/ucp/core.py:145> exception=UCXError('Checksum invalid! 0x305d5bf91953207b != 0x7f6a70bf4870')>
  File "/datasets/pentschev/miniconda3/envs/ucx-src-102-0.14.0b200423/lib/python3.7/site-packages/ucp/core.py", line 78, in exchange_peer_info
ucp.exceptions.UCXError: Checksum invalid! 0x305d5bf91953207b != 0x7f6a70bf4870
```

quasiben:
Wanted to confirm that I am seeing similar issues on a 4 node ib-stress test

pentschev:
I think #509 possibly fixes the issue that motivated this PR, regardless I think this is useful but we could move this change to debug logging only and avoid some extra computation. Thoughts?

jakirkham:
We are just doing the check at startup right? Not every time a message is sent? Assuming I understand that correctly, it seems like a good idea to add this. Yeah there's a little setup cost, but it does provide some reassurance that things are working (or notifies us immediately that they are not). Seems like an easy win to avoid regressions and save us from having to debug this again. WDYT? 🙂

pentschev:
> We are just doing the check at startup right?

Actually, this is done every time a new endpoint is created, some of the causes that may create new endpoints:

1. Client or workers connecting to scheduler;
1. Workers connecting to workers;
1. Client uploading files/running functions on workers.

I don't think this is a very expensive process, so I'm also fine having that at runtime.

madsbk:
rerun tests

ucp/core.py
```diff
@@ -36,35 +35,57 @@ def _get_ctx():
...
+    if expected_checksum != ret["checksum"]:
+        raise RuntimeError(
+            f'Checksum invalid! {hex(expected_checksum)} != {hex(ret["checksum"])}'
+        )
+
+    if port != ret["port"]:
+        raise RuntimeError(f'Port mismatch! {port} != {ret["port"]}')
+
+    if ret["guarantee_msg_order"] != guarantee_msg_order:
         raise ValueError("Both peers must set guarantee_msg_order identically")
```
pentschev, line 86: I don't think either of those would ever be raised, given the checksum above will raise an exception before. I suggest removing them.

madsbk: Notice, the checksum is for the message itself thus both port and guarantee_msg_order can disagree between the two endpoints, which is also tests: https://github.com/rapidsai/ucx-py/blob/f38645426e48e0f9f4731b6bd1bf8e84b097c169/tests/test_guarantee_msg_order.py#L8

pentschev: Oh yes, you're right about `guarantee_msg_order`, my mistake. I think the port case will never happen though as that would require a client to connect to the wrong listener somehow _and_ be lucky that there's a UCX listener on that port exactly, but that's another small check so that's probably fine to stay as well.

madsbk: >  I think the port case will never happen though as that would require a client to connect to the wrong listener somehow

Agree, I added it while debugging and thought it was useful to have as an extra sanity check :)

PR #507, opened 2020-04-23
Fix two worker test

quasiben:
PR fixes the two worker test for NVLink testing.  This is necessary due to updates in cuDF


tests/test_send_recv_two_workers.py
```diff
@@ -217,16 +213,14 @@ def cupy():
...
-    "cuda_obj_generator", [dataframe, empty_dataframe, series, cupy]
+    "cuda_obj_generator", [dataframe, empty_dataframe, series, cupy_obj]
 )
 def test_send_recv_cu(cuda_obj_generator):
     base_env = os.environ
-    env1 = base_env.copy()
-    env2 = base_env.copy()
-    cvd = get_cuda_visible_devices()
+    env_client = base_env.copy()
+    cvd = get_cuda_visible_devices()[:3]
```
pentschev, line 221: Why `[:3]`?

quasiben: Yeah, that's not great.  The `[:3]` is because we are getting a string back from `get_cuda_visible_devices` function: `1,2,3,4,5,6,7`.  We only need the first device `1,2`

quasiben:
@pentschev can you give this another review before merging in ?

tests/utils.py
```diff
@@ -13,17 +13,16 @@ def set_env():
 
 
 def more_than_two_gpus():
-    import pynvml
+    import numba.cuda
```
jakirkham, line 19: Were we running into some issues using pynvml?

quasiben: I was just a bit awkward to use pynvml just to get a list of GPUs

pentschev: I think it's still better to use pynvml, even if the code is more cluttered. Using `numba.cuda` will enforce a context creation in all devices and pynvml won't.

tests/test_send_recv_two_workers.py
pentschev: I see you changed that to `[:2]` now, which seems to be what you wanted.

tests/utils.py
jakirkham: Thanks for the context. Yeah no strong feelings from me either way. It would be good to figure out how to make pynvml usage less awkward.

cc @rjzamora (in case you have thoughts on how we could improve this)

tests/test_send_recv_two_workers.py
```diff
@@ -251,9 +251,8 @@ def test_send_recv_cu(cuda_obj_generator):
 
 
 def total_nvlink_transfer():
-    pynvml.nvmlShutdown()
     pynvml.nvmlInit()
```
jakirkham, line 254: This is certainly an oddity. Makes sense to drop. Though it is curious why it was there.

quasiben: This was just poor coding because I had previously initialized in another function

quasiben:
Thanks all for the reviews.  Merging in

PR #508, opened 2020-04-24
Cleanup setup.py

jakirkham:
Makes a few adjustments to our `setup.py`. Most notably this marks our package as **not** zip-safe. So hopefully Python won't try to bundle it in a zip file.


quasiben:
Thanks @jakirkham

PR #509, opened 2020-04-24
Reorder `exchange_peer_info` send/recv operations

pentschev:
Reorders `exchange_peer_info` operations based on whether the endpoint is a listener or client endpoint.

This seems to fix the `ValueError("Both peers must set guarantee_msg_order identically")` exception being raised when workers start creating endpoints to each other. In 10 IB+NVLink runs on 4 nodes (32 GPUs) that exception wasn't raised a single time, nor did I experience hangs. Without changes here I would see the exception every run and it would hang about 50% of the time.

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -35,7 +35,9 @@ def _get_ctx():
     return _ctx
 
 
-async def exchange_peer_info(endpoint, msg_tag, ctrl_tag, guarantee_msg_order, listener):
+async def exchange_peer_info(
+    endpoint, msg_tag, ctrl_tag, guarantee_msg_order, listener
+):
     """Help function that exchange endpoint information"""
 
     msg_tag = int(msg_tag)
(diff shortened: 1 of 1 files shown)


pentschev:
cc @madsbk @quasiben

quasiben:
Small update, 5/5 runs have passed, albeit with varying merge times:

```
Merge time: 72.20977282524109
Merge time: 28.06112241744995
Merge time: 29.551748514175415
Merge time: 70.80693554878235
Merge time: 104.15069055557251
```

quasiben:
A glance at dask performance reports shows that there was some pickling happening.  I don't think this a big cause for concern -- it may be related to an out of date cuDF and/or distributed package

pentschev:
> If it is a bug in UCX-Py it could be specific to UCX streams or maybe a general race condition that may trigger when an endpoint has multiple outstanding requests?

It is probably some sort of race condition. I don't think there are multiple outstanding requests though, as far as I could see an endpoint will not be used to communicate with more than one endpoint, plus they only send/recv information with streams once, so I can't imagine how there would be multiple outstanding requests.

To be honest, I'm not 100% sure if there's a bug in UCX or UCX-Py or if streams are meant to be used that way, i.e., we should not have transfers in both directions at the same time.

pentschev:
I was also unable to reproduce the large variations Ben has mentioned, running multiple times I see no variations larger than 5-10%.

quasiben:
After upgrading to latest distributed I no longer see the variations

PR #511, opened 2020-04-24
Update to UCX 1.8.0

pentschev:

PR #512, opened 2020-04-24
fix/rndv_scheme

quasiben:
PR fixes a segfault we are seeing in CI.   Still need to understand the difference between the two schemes


pentschev:
@Akshay-Venkatesh this is necessary for UCX 1.8 it seems, otherwise we get segfaults, as in https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/1830/pipeline#log-2406 . Any ideas why this used to work before we moved to 1.8?

pentschev:
I added this change to #511 to avoid many CI rebuilds.

PR #513, opened 2020-04-25
[WIP] Upload_File Fix

quasiben:
This PR seems to fix an issue when using dask's `c.upload_file`.  We thought we had resolved with https://github.com/rapidsai/ucx-py/pull/473 and https://github.com/rapidsai/ucx-py/pull/477 .  I am still a bit confused as to what is happening and why this PR "fixes" things when really the code in the `finally` block should have no issue

cc @VibhuJawa @pentschev

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -478,7 +478,11 @@ class Endpoint:
                 )
         finally:
             # Give all current outstanding send() calls a chance to return
-            self._ctx.worker.progress()
+            worker = getattr(self._ctx, 'worker', None)
+            if worker is None or not worker.initialized:
+                logger.info("UCX Worker is None -- mostly it has already been remove: " + str(self))
+            else:
+                self._ctx.worker.progress()
             await asyncio.sleep(0)
             self.abort()
 
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -478,9 +478,12 @@ class Endpoint:
                 )
         finally:
             # Give all current outstanding send() calls a chance to return
-            worker = getattr(self._ctx, 'worker', None)
+            worker = getattr(self._ctx, "worker", None)
             if worker is None or not worker.initialized:
-                logger.info("UCX Worker is None -- mostly it has already been remove: " + str(self))
+                logger.info(
+                    "UCX Worker is None -- mostly it has already been removed: "
+                    + str(self)
+                )
             else:
                 self._ctx.worker.progress()
             await asyncio.sleep(0)
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -477,17 +477,11 @@ class Endpoint:
                     % (hex(self.uid), repr(e))
                 )
         finally:
-            # Give all current outstanding send() calls a chance to return
-            worker = getattr(self._ctx, "worker", None)
-            if worker is None or not worker.initialized:
-                logger.info(
-                    "UCX Worker is None -- mostly it has already been removed: "
-                    + str(self)
-                )
-            else:
+            if not self.closed():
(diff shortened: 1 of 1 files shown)


ucp/core.py
```diff
@@ -478,7 +478,11 @@ async def close(self):
                 )
         finally:
             # Give all current outstanding send() calls a chance to return
-            self._ctx.worker.progress()
+            worker = getattr(self._ctx, 'worker', None)
+            if worker is None or not worker.initialized:
+                logger.info("UCX Worker is None -- mostly it has already been remove: " + str(self))
```
pentschev, line 483:
```suggestion
                logger.info("UCX Worker is None -- probably it has already been removed: " + str(self))
```

pentschev:
Seems also there's `black` formatting issues. Apart from that, the other build error is the same one that happened yesterday in #511, I'm not sure what that's about.

pentschev:
> Apart from that, the other build error is the same one that happened yesterday in #511, I'm not sure what that's about.

And rerunning tests in #511 made it pass yesterday.

quasiben:
Thanks @pentschev.  Want I find some what troubling is that I never seen the logout output when the work is `None` but I also don't experience the errors reported by @VibhuJawa .

I was using this to test on with 256 workers on 4 nodes:

```python
from distributed import Client

client = Client(scheduler_file='/gpfs/fs1/bzaitlen/dask-local-directory/dask-scheduler.json')

client.upload_file('profiling.py')

def has_file(dask_worker):
    import os
    path = os.path.join(dask_worker.local_directory, 'profiling.py')
    return os.path.exists(path)

print(all(client.run(has_file).values()))
```

quasiben:
Before merging in, it would be good to check in with @madsbk as it's possible the PR works around an underlying problem with closing (hopefully not)

madsbk:
@quasiben, I have push a change to the PR -- now checking ` if not self.closed():` before calling progress, which should be enough. Could you please test and see if also fixes the bug?

The race condition happens if [ `await ucx_api.tag_send()`](https://github.com/quasiben/ucx-py/blob/0b3e41e566dde3db4ba8ef60cf74a905206aa86d/ucp/core.py#L470) doesn't return immediately and another worker aborts the endpoint.

quasiben:
Thanks @madsbk !  I tested and it works for me. @VibhuJawa if you have time can you test as well ?  

@madsbk do you think it's worth adding a test for ep.close() then ep.abort ?

VibhuJawa:
> Thanks @madsbk ! I tested and it works for me. @VibhuJawa if you have time can you test as well ?
> 
> @madsbk do you think it's worth adding a test for ep.close() then ep.abort ?

Will test it in a bit to confirm.

madsbk:
> @madsbk do you think it's worth adding a test for ep.close() then ep.abort ?

The problem is that it only happens if [ `await ucx_api.tag_send()`](https://github.com/quasiben/ucx-py/blob/0b3e41e566dde3db4ba8ef60cf74a905206aa86d/ucp/core.py#L470) returns _after_ the abort by another coroutine or worker. 

Maybe we can try to use the rendezvous protocol to force this?

VibhuJawa:
> Thanks @madsbk ! I tested and it works for me. @VibhuJawa if you have time can you test as well ?

Tested on server and the upload issue seems to be gone.

quasiben:
I filed https://github.com/rapidsai/ucx-py/pull/513 to track whether we can write an internal test for this.  For now, I'll merge in.  Thanks so much for the work @madsbk and @VibhuJawa for confirmation

PR #515, opened 2020-04-27
User tags

madsbk:
This PR implements user tags to send/recv. 

This doesn't implement  "any tag", which would require the use of `ucp_tag_recv_nb()`'s `tag_mask`. Let's implement "any tag" in a future PR if there is any interest.

- [x] Implement tags
- [x] Add tests
- [x] CI pas

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -18,7 +18,12 @@ async def test_tag_match():
     lf = ucp.create_listener(server_node)
     ep = await ucp.create_endpoint(ucp.get_address(), lf.port)
     m1, m2 = (bytearray(len(msg1)), bytearray(len(msg2)))
-    f2 = asyncio.create_task(ep.recv(m2, tag="msg2"))
+    # May be dropped in favor of `asyncio.create_task` only
+    # once Python 3.6 is dropped.
+    if hasattr(asyncio, "create_future"):
+        f2 = asyncio.create_task(ep.recv(m2, tag="msg2"))
+    else:
+        f2 = asyncio.ensure_future(ep.recv(m2, tag="msg2"))
 
     # At this point f2 shouldn't be able to finish because its
     # tag "msg2" doesn't match the servers send tag "msg1"
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -1,0 +1,31 @@
+import asyncio
+
+import pytest
+import ucp
+
+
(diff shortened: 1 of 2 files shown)


tests/test_tags.py
```diff
@@ -0,0 +1,31 @@
...
+@pytest.mark.asyncio
+async def test_tag_match():
+    msg1 = bytes("msg1", "utf-8")
+    msg2 = bytes("msg2", "utf-8")
+
+    async def server_node(ep):
+        f1 = ep.send(msg1, tag="msg1")
+        await asyncio.sleep(1)  # Let msg1 finish
+        f2 = ep.send(msg2, tag="msg2")
+        await asyncio.gather(f1, f2)
```
pentschev, line 16: Why do we have to `await asyncio.sleep(1)` above and then gather the same message here? Wouldn't the gather alone suffice?

pentschev: Looking below at the code, it seems like this is to enforce ordering, is that right?

pentschev:
Seems like you're using `asyncio.create_task` that was only introduced in Python 3.7, apparently you can use `asyncio.ensure_future` for Python 3.6: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task .

tests/test_tags.py
```diff
@@ -0,0 +1,31 @@
...
+
+    lf = ucp.create_listener(server_node)
+    ep = await ucp.create_endpoint(ucp.get_address(), lf.port)
+    m1, m2 = (bytearray(len(msg1)), bytearray(len(msg2)))
+    f2 = asyncio.create_task(ep.recv(m2, tag="msg2"))
```
pentschev, line 21:
```suggestion
    # May be dropped in favor of `asyncio.create_task` only
    # once Python 3.6 is dropped.
    if hasattr(asyncio, "create_future"):
        f2 = asyncio.create_task(ep.recv(m2, tag="msg2"))
    else:
        f2 = asyncio.ensure_future(ep.recv(m2, tag="msg2"))
```

pentschev: Perhaps doing like this to be easy to spot/deprecate in the future?

tests/test_tags.py
madsbk: Yes, enforcing an order with calling `await` :)

PR #521, opened 2020-04-30
Test concatenating trivial CuPy array to message

jakirkham:
Adding this test to try to reproduce an error seen when consolidating messages ( https://github.com/dask/distributed/pull/3732#discussion_r416329022 ).

### DIFF
--- a/tests/test_send_recv.py
+++ b/tests/test_send_recv.py
@@ -112,6 +112,17 @@ async def test_send_recv_cupy(size, dtype, blocking_progress_mode):
     await client.recv(resp)
     np.testing.assert_array_equal(cupy.asnumpy(resp), cupy.asnumpy(msg))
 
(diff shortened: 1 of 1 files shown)


quasiben:
Looks like there was a build timeout:
> tests/test_send_recv.py::test_send_recv_cupy[True-|u1-1] Build timed out (after 20 minutes). Marking the build as failed.

jakirkham:
The failure is expected. This is demonstrating a problem I saw elsewhere (though in that case it aborted).

pentschev:
I'm not sure what's the other use case you're mentioning, but here it's definitely failing because you're not anymore following the expected pattern, as documented in https://github.com/rapidsai/ucx-py/blob/branch-0.14/tests/test_send_recv.py#L36-L44 .

PR #529, opened 2020-05-20
UCXRequest

madsbk:
This PR implements a Python wrapper of the C handle returned by UCX send/recv.

The goal of this PR is to cleanup the ucx-api code and make it easier to split ucx-py into two packages.


ucp/_libs/ucx_api.pyx
```diff
@@ -446,11 +446,75 @@ cdef class UCXListener(UCXObject):
...
+        assert req != NULL
+        self._handle = req
+
+        cdef dict info = {"status": "pending"}
+        if self._handle.info == NULL:  # First time we are wrapping this UCX request
+            Py_INCREF(info)
+            self._handle.info = <PyObject*> info
+            _ucx_request_counter += 1
+            self._uid = _ucx_request_counter
+            assert self._handle.uid == 0
```
pentschev, line 476: According to the class doc, this object doesn't own `self._handle`, is it safe to assume it's not being used elsewhere and cause a failure here?

madsbk: This is what `self._uid` is for. By checking `self._uid == self._handle.uid`, we know that the underlying UCX handle hasn't been modified.

pentschev: I understand that, but is that allowed to happen in some circumstance at runtime? IOW, if that happens is it due to a bug in user code?

pentschev: And by runtime I mean production code, rather than some debug case.

madsbk: Sorry misunderstood your question, yes this can happen. 

Consider a `ucp_tag_send_nb()` call, it can return a UCX handle immediately **and** when the communicate finishes, call `_send_callback(void *request)` where `request` is the same handle .

jakirkham: Should we have a class that does own the handle and this refers to? Wondering how we can make sure the handle gets cleaned up correctly.

madsbk: As far as I can see, it is not possible to make UCX create/free a custom request handle. We can set a [`request_init()` and `request_cleanup()`](http://openucx.github.io/ucx/api/latest/html/group___u_c_p___c_o_n_f_i_g.html#ac1f0a25d794111217af5e0e6b9865458) but "_This function will be called only on the very first time a request memory is initialized, and may not be called again if a request is reused_".

Currently, this PR guarantee that `ucp_request_free()` is called _at most_ once and since our send/recv makes sure to free inflight requests, we shouldn't see any leaks.

pentschev: From the last few comments, it seems that `if self._handle.info == NULL:` should normally ensure that we don't go into this block more than once, and if it happens that we do and the assertion fails, something unexpected did happen. Is that correct?

madsbk: Yes, that is correct.

madsbk:
@pentschev @jakirkham,I have updated the PR so that `request_cancel()` now also takes a `UCXRequest` and checks that the request hasn't been closed prematurely.

ucp/_libs/ucx_api.pyx
pentschev: Ok, that seems reasonable to me then. I'll defer any final comments to @jakirkham , just in case he sees issues with it.

ucp/_libs/ucx_api.pyx
```diff
@@ -446,11 +446,88 @@ cdef class UCXListener(UCXObject):
...
+        cdef ucp_request *req = <ucp_request*>req_as_int
+        assert req != NULL
+        self._handle = req
+
+        cdef dict info = {"status": "pending"}
+        if self._handle.info == NULL:  # First time we are wrapping this UCX request
+            Py_INCREF(info)
+            self._handle.info = <PyObject*> info
+            _ucx_request_counter += 1
+            self._uid = _ucx_request_counter
```
jakirkham, line 475: Would it be possible to use [`uuid.uuid4()`]( https://docs.python.org/3.8/library/uuid.html#uuid.uuid4 ) or perhaps [`secrets.randbits(32)`]( https://docs.python.org/3/library/secrets.html#secrets.randbits )? This would avoid needing a global variable.

ucp/_libs/ucx_api.pyx
```diff
@@ -446,11 +446,88 @@ cdef class UCXListener(UCXObject):
...
+            assert self._handle.uid == 0
+            self._handle.uid = _ucx_request_counter
+        else:
+            self._uid = self._handle.uid
```
jakirkham, line 479: Could we add `weakref.finalize(self.close)` here somewhere to make sure this does get closed at some point?

ucp/_libs/ucx_api.pyx
```diff
@@ -446,11 +446,88 @@ cdef class UCXListener(UCXObject):
...
+
+    @property
+    def handle(self):
+        assert not self.closed()
+        return int(<uintptr_t>self._handle)
+
+    def __hash__(self):
+        if self.closed():
+            assert False
+            return id(self)
```
jakirkham, line 510: Do we need to `return` if we already `assert False`? Also if the intent is to error when this closed, could we add a more descriptive error (perhaps a `ValueError`)?

ucp/_libs/ucx_api_dep.pxd
```diff
@@ -240,21 +241,13 @@ cdef extern from "sys/epoll.h":
...
-    bint finished
-    PyObject *future
-    PyObject *event_loop
-    PyObject *log_msg
-    PyObject *inflight_msgs
-    size_t expected_receive
-    int64_t received
+    bint finished  # Used by downstream projects such as cuML
+    int uid
+    PyObject *info
```
jakirkham, line 246: Should these match [these]( https://github.com/openucx/ucx/blob/a21bdf4d06d7b645ad37844fcac2ab047df9071e/src/ucp/core/ucp_request.h#L102-L104 )?

ucp/_libs/ucx_api.pyx
```diff
@@ -446,11 +446,88 @@ cdef class UCXListener(UCXObject):
...
+    thus we use `_uid` to make sure the handle hasn't been modified.
+    """
+    cdef:
+        ucp_request *_handle
+        int _uid
+
+    def __init__(self, uintptr_t req_as_int):
+        global _ucx_request_counter
+        cdef ucp_request *req = <ucp_request*>req_as_int
+        assert req != NULL
```
jakirkham, line 467: Should we raise an error here (perhaps `ValueError`)?

ucp/_libs/ucx_api.pyx
madsbk: Sorry, the `assert False` is leftovers from debugging :)

ucp/_libs/ucx_api_dep.pxd
madsbk: No, but that is a potential name clash! 
Renamed to `ucp_request => ucx_py_request` and moved its definition into the `ucx_api.pyx` file.

ucp/_libs/ucx_api.pyx
madsbk: I am not sure, this is really an internal check. 
This only fails if UCX calls a callback function with a NULL argument, which mean something has gone seriously wrong :)

ucp/_libs/ucx_api.pyx
madsbk: I must admit that I prefer a global counter in this case -- it is both faster and provide good debug info. But having said that, we could use `uuid.uuid4()` if avoiding global variables is important?

ucp/_libs/ucx_api.pyx
madsbk: It is not easy to do; the problem is that multiple instances of `UCXRequest` can point to the same the underlying UCX request. I think the current implementation is reasonable safe since endpoints always makes sure to closes all `_inflight_msgs`.

ucp/_libs/ucx_api.pyx
jakirkham: No worries. Thanks for clarifying/fixing! :)

ucp/_libs/ucx_api.pyx
jakirkham: Though `self.close()` already protects against this, right?

ucp/_libs/ucx_api.pyx
jakirkham: That's fair. Leaving it makes sense then.

jakirkham:
How does a user create or get a `UCXRequest` object?

ucp/_libs/ucx_api.pyx
madsbk: `self.close()` protects against double calls to `ucp_request_free()` not that we are calling `ucp_request_free` too early. 
Consider a `ucp_tag_send_nb()` call that **before returning** evoke the callback function with a UCX handle wrapped in `UCXRequest`. Now, this handles will goes out of scope and run the finalizer before  `ucp_tag_send_nb()` returns with the second handle.

I have considered using a *close counter* that could count to two but I think that is too much complexity for little gain.

madsbk:
> How does a user create or get a `UCXRequest` object?

They shouldn't :)
Add a class comment:
```
    Don't use this class directly, the send/recv functions and their
    callback functions will return UCXRequest objects.
```

ucp/_libs/ucx_api.pyx
jakirkham: That makes sense. Thanks for the example. Appreciate this is not trivial.

Would [`axtexit.register`]( https://docs.python.org/3.8/library/atexit.html#atexit.register ) make sense or could things still be messed up by order of cleanup?

What happens if no one calls `close`?

jakirkham:
Outside of the remaining question above this seems reasonable to merge to me. Is it ok if we move this to `branch-0.15` or do we need this in `branch-0.14`?

ucp/_libs/ucx_api.pyx
madsbk: > Would [`axtexit.register`](https://docs.python.org/3.8/library/atexit.html#atexit.register) make sense or could things still be messed up by order of cleanup?

At exist, UCX might have freed the memory of the underlying UCX handler and result in a segfault. But remember, at exist all `UCXEndpoints` will cancel all pending `UCXRequest`.

> What happens if no one calls `close`?

Nothing too dangerous, UCX will print a warning saying that a `ucp_request` wasn't freed at exist.

madsbk:
> Outside of the remaining question above this seems reasonable to merge to me. Is it ok if we move this to `branch-0.15` or do we need this in `branch-0.14`?

Agree, let's move it to `branch-0.15`.

ucp/_libs/ucx_api.pyx
jakirkham: If it's already handled by some other cleanup mechanism, that sounds good enough to me.

Good to know. FWIW warnings are not totally free especially at scale as they can grind things to a halt. Leaking semaphores have bit us recently just as one example ( https://github.com/dask/distributed/issues/1078 )

That said, I'm ok leaving this as-is and just keeping an eye open for issues that may crop up.

jakirkham:
Thanks @madsbk! Also thanks @pentschev for the review! 😄

PR #538, opened 2020-05-29
Add endpoint error handler

### LINKED ISSUE #500: Improve resiliency by handling endpoint errors
We are currently not handling endpoint errors, causing even timeouts to raise an error and terminate a UCX-Py process. As per the discussion in https://github.com/openucx/ucx/issues/5027, we need a callback function for that, which I'm working on now.

pentschev:
This PR uses the `UCP_ERR_HANDLING_MODE_PEER` mode by default for InfiniBand endpoints, preventing a UCX process from exiting with a fatal error should an issue occur. This mode isn't currently supported by UCX with the TCP transport, therefore no callback is added should the `UCX_TLS` not contain `dc`, `ib` or `rc`. It's also possible to prevent endpoints and listeners from using this mode by passing `endpoint_error_handling=False`.

As future work, we may want to add some mechanism to enforce `UCXEndpoint` to be destroyed once an error occurs and the callback is executed. This seems currently not necessary in Dask, once can test this only with IB enabled and kill a worker in the middle of a task and a new one will spawn without any other processes exiting with a fatal error. I'm not yet sure on a pytest to add here, and such test would _not_ be executed by CI anyway given its lack of IB support.

Fixes #500 .

### DIFF
--- a/tests/debug-tests/test_endpoint_error_callback.py
+++ b/tests/debug-tests/test_endpoint_error_callback.py
@@ -1,5 +1,7 @@
 # This test requires InfiniBand, to run:
-# UCXPY_IFNAME=ib0 UCX_NET_DEVICES=mlx5_0:1 UCX_TLS=rc,tcp,sockcm,cuda_copy UCX_SOCKADDR_TLS_PRIORITY=sockcm py.test --cache-clear tests/debug-tests/test_endpoint_error_callback.py
+# UCXPY_IFNAME=ib0 UCX_NET_DEVICES=mlx5_0:1 \
+# UCX_TLS=rc,tcp,sockcm,cuda_copy UCX_SOCKADDR_TLS_PRIORITY=sockcm \
+# py.test --cache-clear tests/debug-tests/test_endpoint_error_callback.py
 
 import asyncio
 import logging
--- a/tests/debug-tests/test_endpoint_error_callback.py
+++ b/tests/debug-tests/test_endpoint_error_callback.py
@@ -1,0 +1,173 @@
+# This test requires InfiniBand, to run:
+# UCXPY_IFNAME=ib0 UCX_NET_DEVICES=mlx5_0:1 UCX_TLS=rc,tcp,sockcm,cuda_copy UCX_SOCKADDR_TLS_PRIORITY=sockcm py.test --cache-clear tests/debug-tests/test_endpoint_error_callback.py
+
+import asyncio
+import logging
+import multiprocessing
+import os
+import queue
+import random
+import signal
+import sys
+
+from distributed.comm.utils import to_frames
+from distributed.protocol import to_serialize
+
+import cloudpickle
+import pytest
+import ucp
+from ucp.utils import get_ucxpy_logger
+from utils import get_cuda_devices, get_num_gpus, recv, send
+
+cupy = pytest.importorskip("cupy")
+
+
+def get_log_queue_handler():
+    handler = logging.StreamHandler()
+
+    log_queue = queue.Queue(-1)  # no limit on size
+    queue_handler = logging.handlers.QueueHandler(log_queue)
+    listener = logging.handlers.QueueListener(log_queue, handler)
+
+    ucxpy_logger = get_ucxpy_logger()
+    ucxpy_logger.addHandler(queue_handler)
+
+    formatter = logging.Formatter("%(message)s")
+    handler.setFormatter(formatter)
+
+    return log_queue, listener
+
+
+async def get_ep(name, port):
+    addr = ucp.get_address()
+    ep = await ucp.create_endpoint(addr, port)
+    return ep
+
+
+def client(port, func):
+    # wait for server to come up
+    # receive object
+    # process suicides
+
+    ucp.init()
+
+    # must create context before importing
+    # cudf/cupy/etc
+
+    async def read():
+        await asyncio.sleep(1)
+        ep = await get_ep("client", port)
+        msg = None
+        import cupy
+
+        cupy.cuda.set_allocator(None)
+
+        frames, msg = recv(ep)
+
+        # Client process suicides to force an "Endpoint timeout"
+        # on the server
+        os.kill(os.getpid(), signal.SIGKILL)
+
+    asyncio.get_event_loop().run_until_complete(read())
+
+
+def server(port, func):
+    # create listener receiver
+    # add queue logger
+    # write cudf object
+    # terminates ep/listener
+    # checks that "Endpoint timeout" was logged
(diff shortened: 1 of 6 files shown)


quasiben:
Thank you so much for this work @pentschev.  I don't suppose you have an idea for a test here to trigger an endpoint error ?  Or outline how we might go about testing this in the future

pentschev:
I've been thinking about this, but unfortunately I don't have an idea yet. Suggestions are welcome too. :)

pentschev:
It took me a lot of time, but I managed to get a test right. We can't test this without InfiniBand as TCP doesn't support error callbacks currently, for that reason CI won't run it but it can be useful for testing on other setups, as well as TCP testing when support for endpoint error callbacks arrive.

pentschev:
rerun tests

pentschev:
The failing test is in distributed, not UCX-Py. It seems to be a serialization test but it doesn't fail on all systems, I tested on a DGX-1 and it passes. Any idea what could be the issue @jakirkham ?

tests/debug-tests/test_endpoint_error_callback.py
```diff
@@ -0,0 +1,179 @@
...
+            while not lf.closed():
+                await asyncio.sleep(0.1)
+        except ucp.UCXCloseError:
+            pass
+
+    log_listener.stop()
+
+    asyncio.get_event_loop().run_until_complete(f(port))
+
+    # Check log for the expected "Endpoint timeout" and exits with
```
quasiben, line 126: Thank you for adding this test Peter!  It might be worth mentioning that without `UCP_ERR_HANDLING_MODE_PEER` we would not see an endpoint timeout and instead would see fatal error.

pentschev: I honestly don't see the point in writing that. The test is specifically testing the endpoint error callback, seems redundant to me.

quasiben: That's true.  I suppose the name of the test `test_endpoint_error_callback` is a large enough clue.

pentschev: On the other hand, I realized we didn't test without the endpoint error callback, which we can do and make it clear. Let me try to add that.

pentschev: Ok, I added the test, hopefully it makes more sense now.

tests/debug-tests/test_endpoint_error_callback.py
```diff
@@ -127,10 +131,9 @@ async def write(ep):
     # status 0 if encountered, -1 otherwise.
     while not log_queue.empty():
         log = log_queue.get()
-        print(log)
         if "Endpoint timeout" in log.getMessage():
-            sys.exit(0)
-    sys.exit(-1)
+            sys.exit(-80)
```
quasiben, line 136: What is a sys.exit with `-80` ?

pentschev: It's just the error code for endpoint timeout in UCX: https://github.com/openucx/ucx/blob/c7a22c915c4722135af1366922c702389dbdd320/src/ucs/type/status.h#L83 . I didn't have any better ideas, so I thought this made sense.

jakirkham:
> The failing test is in distributed, not UCX-Py. It seems to be a serialization test but it doesn't fail on all systems, I tested on a DGX-1 and it passes. Any idea what could be the issue @jakirkham ?

Which test? Do you have the output?

pentschev:
Yeah, the one from centos7-cuda10.0 build: https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/1975/pipeline#log-1877

jakirkham:
Thanks. Will take a look.

jakirkham:
Added PR ( https://github.com/rapidsai/cudf/pull/5379 ) to cuDF to fix the Distributed test failure. Have confirmed this fixes the issue locally as well.

quasiben:
rerun tests

pentschev:
Seems to still be failing. :(

@jakirkham were you able to reproduce that issue locally before https://github.com/rapidsai/cudf/pull/5379 ? I'm asking because I wasn't.

quasiben:
I was able to reproduce with latest cuDF from nightly builds

pentschev:
I still can't on an environment from yesterday. What version of CUDA toolkit did you use @quasiben ?

quasiben:
```
# packages in environment at /datasets/bzaitlen/miniconda3/envs/20200602:
#
# Name                    Version                   Build  Channel
cudatoolkit               10.2.89              h6bb024c_0    nvidia
(20200602) bzaitlen@dgx12:/datasets/bzaitlen/GitRepos/distributed$ conda list cudf
# packages in environment at /datasets/bzaitlen/miniconda3/envs/20200602:
#
# Name                    Version                   Build  Channel
cudf                      0.15.0a200603   py37_gb32e777_731    rapidsai-nightly
dask-cudf                 0.15.0a200603   py37_gb32e777_731    rapidsai-nightly
libcudf                   0.15.0a200603   cuda10.2_gb32e777_731    rapidsai-nightly
```

pentschev:
Oh, that was my mistake, I was running 0.14 nightly. I've now created a new environment with 0.15 and I can reproduce that even after https://github.com/rapidsai/cudf/pull/5379 was merged.

pentschev:
It seems like rapidsai-nightly conda packages had not been updated, but they were about 5 minutes ago. I tested the new packages and now the test passes locally.

pentschev:
rerun tests

pentschev:
Woohoo! It passes now! Thanks @jakirkham for fixing the issue!

quasiben:
@madsbk can I ask you to take a look at this PR ?

madsbk:
> Unfortunately that error is raised by UCX _before_ `exchange_peer_info` gets executed so I don't think we can check for that currently. I don't know of a way to handle that currently.

Fair enough, at least it throws an error. It would have been way worse if the result was an unstable connection.

pentschev:
> Fair enough, at least it throws an error.

It does, it's not user-friendly though as it seems to cause some sort of hang in UCX, so it's definitely something that we could work on improving in the future, even though I'd consider it low-priority.

pentschev:
Any more comments here @quasiben ?

quasiben:
Nope!  Thank you again @pentschev ! Merging in

pentschev:
Thanks everyone for reviews!

Issue #531, opened 2020-05-21
Running IB multi-GPU test on `cudf.Series` crashes


pentschev:
The new multi-GPU test introduced in #530 crashes when transferring `cudf.Series`. That can be reproduced on a DGX-1 with the following command:

```
UCX_TLS=rc,tcp,sockcm,cuda_copy UCX_SOCKADDR_TLS_PRIORITY=sockcm pytest tests/debug-tests/test_send_recv_many_workers.py -vs
```

The crash output is shown below:

```
mlx5: dgx13: got completion with error:
00000000 00000000 00000000 00000000
00000000 00000000 00000000 00000000
00000002 00000000 00000000 00000000
00000000 93003204 10003878 0002ded2
[dgx13:46740:0:46740] rc_verbs_iface.c:65   send completion with error: local protection error qpn 0x3878 wrid 0x1 vendor_err 0x32
==== backtrace (tid:  46740) ====
 0  /datasets/pentschev/miniconda3/envs/rn-102-0.14.0b200518/lib/libucs.so.0(ucs_handle_error+0x12d) [0x7f8e15a7047d]
 1  /datasets/pentschev/miniconda3/envs/rn-102-0.14.0b200518/lib/libucs.so.0(ucs_fatal_error_message+0x51) [0x7f8e15a6d291]
 2  /datasets/pentschev/miniconda3/envs/rn-102-0.14.0b200518/lib/libucs.so.0(+0x241c7) [0x7f8e15a711c7]
 3  /datasets/pentschev/miniconda3/envs/rn-102-0.14.0b200518/lib/libucs.so.0(ucs_log_dispatch+0xd9) [0x7f8e15a712f9]
 4  /datasets/pentschev/miniconda3/envs/rn-102-0.14.0b200518/lib/ucx/libuct_ib.so.0(+0x1d07e) [0x7f8e10c2507e]
 5  /datasets/pentschev/miniconda3/envs/rn-102-0.14.0b200518/lib/ucx/libuct_ib.so.0(+0x1d6be) [0x7f8e10c256be]
 6  /datasets/pentschev/miniconda3/envs/rn-102-0.14.0b200518/lib/libucp.so.0(ucp_worker_progress+0x6a) [0x7f8e15b1109a]
[...]

PR #540, opened 2020-06-04
Mark many worker send/recv as xfail without RMM

jakirkham:
This test currently fails without the RMM pool enabled. So mark it as an xfailing test. We can always selectively enable it if we wish to explore/debug it further.

xref: https://github.com/rapidsai/ucx-py/issues/531

### DIFF
--- a/tests/debug-tests/test_send_recv_many_workers.py
+++ b/tests/debug-tests/test_send_recv_many_workers.py
@@ -195,6 +195,13 @@ def cupy_obj():
 )
 @pytest.mark.parametrize("enable_rmm", [True, False])
 def test_send_recv_cu(cuda_obj_generator, enable_rmm):
+    if not enable_rmm:
+        pytest.xfail(
+            "This test currently does not work without RMM enabled."
+            " Please see issue"
+            " ( https://github.com/rapidsai/ucx-py/issues/531 ) for details."
+        )
+
     cuda_visible_devices_base = os.environ.get("CUDA_VISIBLE_DEVICES")
     ucx_net_devices_base = os.environ.get("UCX_NET_DEVICES")
 


pentschev:
These are debug tests, so IMO we should not xfail them as their whole purpose is to be explicit of issues.

jakirkham:
It's always possible to [run pytest with xfailed tests]( https://docs.pytest.org/en/stable/skipping.html#ignoring-xfail ).

jakirkham:
rerun tests

pentschev:
I would still prefer to be as verbose as possible in debug tests. Note that this test is not even running in CI, and an xfailed test is much easier to overlook than a crash in your face. :)

jakirkham:
Well as a user of the test suite, I think this is a useful feature as it allows one to run through the test suite without getting a crash from a known issue. Not to mention time spent trying to figure out whether that failure is caused by one's own code, if others know about, etc.

pentschev:
That's a fair point. I think the best is to separate tests and debug-tests to avoid such issues. The debug tests can still cause hangs and other more severe issues that xfail would not be helpful. Let's merge this for now and I'll later work on ensuring debug-tests are not in the same place to avoid people from running them by mistake.

pentschev:
Argh, I just realized we're not checking the transports enabled here. The test does not work with RMM _only_ when `rc` is specified, so it should not xfail for all cases. I'll fix that in a follow-up PR.

jakirkham:
Oh sorry I misunderstood. I thought this always failed when RMM is disabled.

pentschev:
No worries, it was my mistake for reviewing as I was half asleep. :(

jakirkham:
How should we check that do you think? Inspecting the environment variable? Or do we have a better way to check this?

jakirkham:
NVM it seems PR ( https://github.com/rapidsai/ucx-py/pull/541 ) should address this.

pentschev:
Yeah, #541 should prevent issues. But just to respond to your question, currently we would need to check the environment variable as we need to explicitly set them, if we were using `UCX_TLS=all` then it would be even more complicated, I'm not sure if we would have a way of doing that today.

PR #545, opened 2020-06-10
Disable endpoint error handling by default

pentschev:
When the error handler is enabled, there's no NVLink transfers happening even with `cuda_ipc`. This PR disables the error handler by default until we can sort out why that happens.


jakirkham:
rerun tests

jakirkham:
Thanks Peter! 😄

PR #546, opened 2020-06-12
Optimize buffer utility functions

jakirkham:
Rewrites buffer utility functions to get more optimizations out of Cython. This includes typing variables where possible, leveraging `for`-loops that are easier for Cython to lower to C, making use of Cython `memoryview`s, and some code simplification (where possible).

cc @pentschev @madsbk @quasiben

### DIFF
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -1,4 +1,4 @@
-# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
 # cython: language_level=3
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -63,17 +63,18 @@ def get_buffer_nbytes(buffer, check_min_size, cuda_support):
         itemsize = int(numpy.dtype(iface["typestr"]).itemsize)
         # Making sure that the elements in shape is integers
         shape = [int(s) for s in iface["shape"]]
+        ndim = len(shape)
         nbytes = reduce(operator.mul, shape, itemsize)
         # Check that data is contiguous
         strides = iface.get("strides")
-        if len(shape) > 0 and strides is not None:
+        if ndim > 0 and strides is not None:
             strides = [int(s) for s in strides]
-            if len(strides) != len(shape):
+            if len(strides) != ndim:
                 raise ValueError(
                     "The length of shape and strides must be equal"
                 )
             s = itemsize
-            for i in reversed(range(len(shape))):
+            for i in reversed(range(ndim)):
                 if s != strides[i]:
                     raise ValueError("Array must be contiguous")
                 s *= shape[i]
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -82,8 +82,8 @@ cpdef Py_ssize_t get_buffer_nbytes(buffer, check_min_size, bint cuda_support) ex
     else:
         mview = memoryview(buffer)
         nbytes = mview.nbytes
-        if not mview.contiguous:
-            raise ValueError("buffer must be contiguous")
+        if not mview.c_contiguous:
+            raise ValueError("buffer must be C-contiguous")
 
     if check_min_size is not None:
         min_size = check_min_size
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -15,9 +15,7 @@ cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
     is read only and check_writable=True is set.
     """
 
-    cdef dict iface = getattr(buffer, "__array_interface__", None)
-    if iface is None:
-        iface = getattr(buffer, "__cuda_array_interface__", None)
+    cdef dict iface = getattr(buffer, "__cuda_array_interface__", None)
 
     cdef uintptr_t data_ptr
     cdef bint data_readonly
@@ -45,18 +43,16 @@ cpdef Py_ssize_t get_buffer_nbytes(buffer, check_min_size, bint cuda_support) ex
     if `check_min_size` is greater than the size of the buffer
     """
 
-    cdef dict iface = getattr(buffer, "__array_interface__", None)
-    if iface is None:
-        iface = getattr(buffer, "__cuda_array_interface__", None)
-        if not cuda_support and iface is not None:
-            raise ValueError(
-                "UCX is not configured with CUDA support, please add "
-                "`cuda_copy` and/or `cuda_ipc` to the UCX_TLS environment"
-                "variable and that the ucx-proc=*=gpu package is "
-                "installed. See "
-                "https://ucx-py.readthedocs.io/en/latest/install.html for "
-                "more information."
-            )
+    cdef dict iface = getattr(buffer, "__cuda_array_interface__", None)
+    if not cuda_support and iface is not None:
+        raise ValueError(
+            "UCX is not configured with CUDA support, please add "
+            "`cuda_copy` and/or `cuda_ipc` to the UCX_TLS environment"
+            "variable and that the ucx-proc=*=gpu package is "
+            "installed. See "
+            "https://ucx-py.readthedocs.io/en/latest/install.html for "
+            "more information."
+        )
 
     cdef tuple shape, strides
     cdef Py_ssize_t i, s, itemsize, ndim, nbytes, min_size
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -6,6 +6,7 @@
 
 from cpython.memoryview cimport PyMemoryView_GET_BUFFER
 from libc.stdint cimport uintptr_t
+cimport cython
 
 
 cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
@@ -36,6 +37,8 @@ cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
     return data_ptr
 
 
+@cython.boundscheck(False)
+@cython.wraparound(False)
 cpdef Py_ssize_t get_buffer_nbytes(buffer, check_min_size, bint cuda_support) except *:
     """
     Returns the size of the buffer in bytes. Returns ValueError
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -10,8 +10,6 @@ from functools import reduce
 from cpython.memoryview cimport PyMemoryView_GET_BUFFER
 from libc.stdint cimport uintptr_t
 
-from ..exceptions import UCXCloseError, UCXError
-
 
 def get_buffer_data(buffer, check_writable=False):
     """
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -86,6 +86,10 @@ def get_buffer_nbytes(buffer, check_min_size, cuda_support):
         if not mview.contiguous:
             raise ValueError("buffer must be contiguous")
 
-    if check_min_size is not None and nbytes < check_min_size:
-        raise ValueError("the nbytes is greater than the size of the buffer!")
+    if check_min_size is not None:
+        min_size = check_min_size
+        if nbytes < min_size:
+            raise ValueError(
+                "the nbytes is greater than the size of the buffer!"
+            )
     return nbytes
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -70,8 +70,9 @@ def get_buffer_nbytes(buffer, check_min_size, cuda_support):
         shape = [int(s) for s in iface["shape"]]
         nbytes = reduce(operator.mul, shape, 1) * itemsize
         # Check that data is contiguous
-        if len(shape) > 0 and iface.get("strides") is not None:
-            strides = [int(s) for s in iface['strides']]
+        strides = iface.get("strides")
+        if len(shape) > 0 and strides is not None:
+            strides = [int(s) for s in strides]
             if len(strides) != len(shape):
                 msg = "The length of shape and strides must be equal"
                 raise ValueError(msg)
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -31,11 +31,6 @@ def get_buffer_data(buffer, check_writable=False):
         data_ptr = int(<uintptr_t>PyMemoryView_GET_BUFFER(mview).buf)
         data_readonly = mview.readonly
 
-    # Workaround for numba giving None, rather than an 0.
-    # https://github.com/cupy/cupy/issues/2104 for more info.
-    if data_ptr is None:
-        data_ptr = 0
-
     if data_ptr == 0:
         raise NotImplementedError("zero-sized buffers isn't supported")
 
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -5,8 +5,8 @@
 
 
 from cpython.memoryview cimport PyMemoryView_GET_BUFFER
+from cython cimport boundscheck, wraparound
 from libc.stdint cimport uintptr_t
-cimport cython
 
 
 cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
@@ -37,8 +37,8 @@ cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
     return data_ptr
 
 
-@cython.boundscheck(False)
-@cython.wraparound(False)
+@boundscheck(False)
+@wraparound(False)
 cpdef Py_ssize_t get_buffer_nbytes(buffer, check_min_size, bint cuda_support) except *:
     """
     Returns the size of the buffer in bytes. Returns ValueError
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -68,7 +68,7 @@ def get_buffer_nbytes(buffer, check_min_size, cuda_support):
         itemsize = int(numpy.dtype(iface["typestr"]).itemsize)
         # Making sure that the elements in shape is integers
         shape = [int(s) for s in iface["shape"]]
-        nbytes = reduce(operator.mul, shape, 1) * itemsize
+        nbytes = reduce(operator.mul, shape, itemsize)
         # Check that data is contiguous
         strides = iface.get("strides")
         if len(shape) > 0 and strides is not None:
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -4,7 +4,6 @@
 # cython: language_level=3
 
 
-import asyncio
 import operator
 from functools import reduce
 
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -28,7 +28,7 @@ def get_buffer_data(buffer, check_writable=False):
         data_ptr, data_readonly = iface["data"]
     else:
         mview = memoryview(buffer)
-        data_ptr = int(<uintptr_t>PyMemoryView_GET_BUFFER(mview).buf)
+        data_ptr = <uintptr_t>PyMemoryView_GET_BUFFER(mview).buf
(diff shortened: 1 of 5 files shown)


pentschev:
@jakirkham I'm getting some errors when I try the send/recv debug test:

```
$ UCX_TLS=tcp,sockcm,cuda_copy UCX_SOC
KADDR_TLS_PRIORITY=sockcm python recv.py
'CUDA_VISIBLE_DEVICES'
ITER:  0
IndexError: Out of bounds on buffer access (axis 0)
Exception ignored in: 'ucp._libs.utils.get_buffer_nbytes'
IndexError: Out of bounds on buffer access (axis 0)
  File "recv.py", line 140, in <module>
    test_send_recv_cu(cupy)
  File "recv.py", line 113, in test_send_recv_cu
    client(env2, port, func)
  File "recv.py", line 58, in client
    asyncio.get_event_loop().run_until_complete(read())
  File "/datasets/pentschev/miniconda3/envs/rn-102-0.15.0b200615/lib/python3.7/asyncio/base_events.py", line 583, in run_until_complete
    return future.result()
  File "recv.py", line 47, in read
    frames, msg = await recv(ep)
  File "/datasets/pentschev/src/ucx-py/debug-tests/utils.py", line 104, in recv
    await ep.recv(frame)
  File "/datasets/pentschev/miniconda3/envs/rn-102-0.15.0b200615/lib/python3.7/site-packages/ucp/core.py", line 604, in recv
    ret = await ucx_api.tag_recv(self._ep, buffer, nbytes, tag, log_msg=log)
ucp.exceptions.UCXMsgTruncated: Comm Error "[Recv #005] ep: 0x7fcf28038048, tag: 0x68f5dfcc8a2d7c57, nbytes: 0, type: <class 'rmm._lib.device_buffer.DeviceBuffer'>": length mismatch: 472392 (got) != 0 (expected)
```

I'm guessing this has to do with some of the logic change in `get_buffer_nbytes`, but I'm not sure if this is ready to test or you're still working on it.

jakirkham:
Yeah there are not really tests of this functionality, which makes it hard to tell what would be expected of these functions. Think we should get some tests in first and then revisit.

jakirkham:
I've added some tests in PR ( https://github.com/rapidsai/ucx-py/pull/549 ).

ucp/_libs/utils.pyx
```diff
@@ -42,41 +42,52 @@ def get_buffer_data(buffer, check_writable=False):
...
+    cdef int64_t[::1] strides
+    cdef size_t nbytes
     if iface is not None:
         import numpy
-        itemsize = int(numpy.dtype(iface['typestr']).itemsize)
+        itemsize = numpy.dtype(iface["typestr"]).itemsize
         # Making sure that the elements in shape is integers
-        shape = [int(s) for s in iface['shape']]
-        nbytes = reduce(operator.mul, shape, 1) * itemsize
+        shape = numpy.array(iface["shape"], dtype=numpy.uint64)
```
kkraus14, line 73: What is `iface["shape"]` here? A tuple? In Numba previously @shwina found that allocating a numpy array to do a cumulative product of a tuple is much more expensive then just doing an `itertools.reduce` for example. We should likely be even more efficient in using the Python C-API here if we wanted and iterating over the tuple via that.

jakirkham:
I wouldn't worry about reviewing this yet. We seem to still be getting failures atm. Would be good to get some tests in first ( https://github.com/rapidsai/ucx-py/pull/549 ) so we can better understand the origin of the failures.

ucp/_libs/utils.pyx
jakirkham: Yeah wasn't planning to stick with NumPy. In any event there are a few ways of coercing something like a `tuple` to a C-style array. Of them just using the builtin `array` should be fine for small arrays (as seen below). This was run on a DGX-1.

```python
In [1]: from array import array 
   ...: import struct 
   ...:  
   ...: import numpy                                                            

In [2]: t = (2, 3, 4)                                                           

In [3]: %timeit numpy.array(t)                                                  
1.17 µs ± 2.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit array("l", t)                                                   
243 ns ± 0.105 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit struct.pack(f"{len(t)}l", *t)                                   
368 ns ± 1.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

ucp/_libs/utils.pyx
```diff
@@ -1,38 +1,40 @@
...
-from libc.stdint cimport uintptr_t
+from libc.stdint cimport int64_t, uint64_t, uintptr_t
+
+from array import array
 
 from ..exceptions import UCXCloseError, UCXError
 
 
-def get_buffer_data(buffer, check_writable=False):
+cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
```
kkraus14, line 13:
```suggestion
cpdef uintptr_t get_buffer_data(object buffer, bint check_writable=False) except *:
```

ucp/_libs/utils.pyx
jakirkham: This now uses `array`. So marking this as resolved.

ucp/_libs/utils.pyx
```diff
@@ -43,44 +45,55 @@ def get_buffer_data(buffer, check_writable=False):
     return data_ptr
 
 
-def get_buffer_nbytes(buffer, check_min_size, cuda_support):
+cpdef size_t get_buffer_nbytes(buffer, check_min_size, bint cuda_support) except *:
```
kkraus14, line 48: Not sure on the typing of `check_min_size` here:
```suggestion
cpdef size_t get_buffer_nbytes(object buffer, object check_min_size, bint cuda_support) except *:
```

ucp/_libs/utils.pyx
```diff
@@ -1,38 +1,40 @@
...
 
 
-def get_buffer_data(buffer, check_writable=False):
+cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
     """
     Returns data pointer of the buffer. Raising ValueError if the buffer
     is read only and check_writable=True is set.
     """
-    iface = None
+    cdef dict iface = None
```
kkraus14, line 18: I didn't know this is allowed, neat!

ucp/_libs/utils.pyx
```diff
@@ -43,44 +45,55 @@ def get_buffer_data(buffer, check_writable=False):
...
         import numpy
-        itemsize = int(numpy.dtype(iface['typestr']).itemsize)
+        itemsize = numpy.dtype(iface["typestr"]).itemsize
         # Making sure that the elements in shape is integers
-        shape = [int(s) for s in iface['shape']]
-        nbytes = reduce(operator.mul, shape, 1) * itemsize
+        shape = array("L", iface["shape"])
+        nbytes = itemsize
+        for i in range(len(shape)):
+            nbytes *= shape[i]
```
kkraus14, line 82: Is constructing an array from the tuple and then iterating over the array actually cheaper than just iterating over the tuple itself?

jakirkham:
We are seeing a test failure due to a recent change in Dask ( https://github.com/dask/distributed/pull/4019 ). Submitted PR ( https://github.com/dask/distributed/pull/4036 ) to fix that.

ucp/_libs/utils.pyx
jakirkham: Cython already defaults to `object` if a type is specified. So we are ok here.

ucp/_libs/utils.pyx
jakirkham: Cython already defaults to `object` if a type is specified. So we are ok here.

Yeah `check_min_size` could be several things. So `object` is probably as close as we can get.

jakirkham:
rerun tests

ucp/_libs/utils.pyx
```diff
@@ -42,41 +42,52 @@ def get_buffer_data(buffer, check_writable=False):
...
+            if len(shape) > 0:
+                if len(strides) != len(shape):
+                    raise ValueError(
+                        "The length of shape and strides must be equal"
+                    )
+                s = <int64_t>itemsize
+                for i from len(shape) >= i >= 0 by 1:
+                    if s != strides[i]:
+                        raise ValueError("Array must be contiguous")
+                    s *= <int64_t>shape[i]
```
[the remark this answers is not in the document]
jakirkham: This was done in recent commits.

ucp/_libs/utils.pyx
```diff
@@ -16,23 +18,17 @@ def get_buffer_data(buffer, check_writable=False):
...
     """
-    iface = None
-    if hasattr(buffer, "__cuda_array_interface__"):
-        iface = buffer.__cuda_array_interface__
-    elif hasattr(buffer, "__array_interface__"):
-        iface = buffer.__array_interface__
+
+    iface = getattr(buffer, "__array_interface__", None)
+    if iface is None:
+        iface = getattr(buffer, "__cuda_array_interface__", None)
```
kkraus14, line 22: Can we swap these to say the extra `getattr` call in the case of GPU memory? 😄

ucp/_libs/utils.pyx
```diff
@@ -16,23 +18,17 @@ def get_buffer_data(buffer, check_writable=False):
...
         mview = memoryview(buffer)
-        data_ptr = int(<uintptr_t>PyMemoryView_GET_BUFFER(mview).buf)
-        data_readonly = mview.readonly
-
-    # Workaround for numba giving None, rather than an 0.
-    # https://github.com/cupy/cupy/issues/2104 for more info.
-    if data_ptr is None:
-        data_ptr = 0
+        data_ptr = <uintptr_t>PyMemoryView_GET_BUFFER(mview).buf
+        data_readonly = <bint>PyMemoryView_GET_BUFFER(mview).readonly
```
kkraus14, line 29: Can store an intermediate of the result of `PyMemoryView_GET_BUFFER(mview)` here to save a call.

ucp/_libs/utils.pyx
```diff
@@ -49,38 +45,40 @@ def get_buffer_nbytes(buffer, check_min_size, cuda_support):
...
-                  "variable and that the ucx-proc=*=gpu package is " \
-                  "installed. See " \
-                  "https://ucx-py.readthedocs.io/en/latest/install.html for " \
-                  "more information."
-            raise ValueError(msg)
-    elif hasattr(buffer, "__array_interface__"):
-        iface = buffer.__array_interface__
+    iface = getattr(buffer, "__array_interface__", None)
+    if iface is None:
+        iface = getattr(buffer, "__cuda_array_interface__", None)
```
kkraus14, line 48: Similarly can we swap these?

jakirkham: We need to check whether we have a CUDA object and whether those are allowed in this case below.

ucp/_libs/utils.pyx
jakirkham: [`PyMemoryView_GET_BUFFER`]( https://github.com/python/cpython/blob/d905df766c367c350f20c46ccd99d4da19ed57d8/Include/memoryobject.h#L18 ) is just a macro that access the `Py_buffer` object owned by the `memoryview`.

ucp/_libs/utils.pyx
```diff
@@ -1,38 +1,31 @@
...
+        data_ptr, data_readonly = iface["data"]
     else:
         mview = memoryview(buffer)
-        data_ptr = int(<uintptr_t>PyMemoryView_GET_BUFFER(mview).buf)
-        data_readonly = mview.readonly
-
-    # Workaround for numba giving None, rather than an 0.
-    # https://github.com/cupy/cupy/issues/2104 for more info.
-    if data_ptr is None:
-        data_ptr = 0
```
jakirkham, line 35: This was fixed in Numba PR ( https://github.com/numba/numba/pull/4609 ), which was included in 0.46.0+ released last October. We require [`numba >=0.49.0`]( https://github.com/rapidsai/integration/blob/5a6b620e70ed774b202ad91868adcf325d23df11/conda/recipes/versions.yaml#L96-L97 ) currently and likely will need to bump that before RAPIDS 0.15 is released. So I think we are safe to drop this.

ucp/_libs/utils.pyx
```diff
@@ -43,51 +37,62 @@ def get_buffer_data(buffer, check_writable=False):
...
-                  "variable and that the ucx-proc=*=gpu package is " \
-                  "installed. See " \
-                  "https://ucx-py.readthedocs.io/en/latest/install.html for " \
-                  "more information."
-            raise ValueError(msg)
-    elif hasattr(buffer, "__array_interface__"):
-        iface = buffer.__array_interface__
+    cdef dict iface = getattr(buffer, "__array_interface__", None)
+    if iface is None:
+        iface = getattr(buffer, "__cuda_array_interface__", None)
```
madsbk, line 50: Ditto

ucp/_libs/utils.pyx
```diff
@@ -1,38 +1,32 @@
...
     """
-    iface = None
-    if hasattr(buffer, "__cuda_array_interface__"):
-        iface = buffer.__cuda_array_interface__
-    elif hasattr(buffer, "__array_interface__"):
-        iface = buffer.__array_interface__
 
+    cdef dict iface = getattr(buffer, "__array_interface__", None)
+    if iface is None:
+        iface = getattr(buffer, "__cuda_array_interface__", None)
```
madsbk, line 20: Don't we want `"__cuda_array_interface__"` to take precedence over `"__array_interface__"`?

ucp/_libs/utils.pyx
madsbk, line 50: Ditto

jakirkham: Please see [this comment]( https://github.com/rapidsai/ucx-py/pull/546#discussion_r468342638 ).

jakirkham:
Went back through and profiled things carefully. Noted that `hasattr` was costing us (more details in the OP of PR https://github.com/rapidsai/ucx-py/pull/573 since included here). So changed the code to use `getattr` instead, which helped quite a bit. Then made some other similar changes just focusing on the Python code.

After that went back and readded Cython types to the Python variables and converted comprehensions to `for`-loops. Though have skipped dealing with arrays. Arrays are a bit slower, but maybe not noticeable given the overall runtime. Maybe they can be revisited in another PR, but would prefer to skip them for now.

With all of these changes, went through and timed the handling of an N-D array (including the case where the C-contiguous check would fail). Have included the times before and after below. In short, all cases are more than 2x faster. This doesn't sound like much, but these functions are already running `~1µs`.

<details>
<summary>Runtime <b>before</b> of <code>get_buffer_data</code> and <code>get_buffer_nbytes</code>:</summary>

<br>

```python
In [1]: import cupy 
   ...: import numpy 
   ...:  

   ...:     try: 
   ...:     except Exception: 
   ...:         pass 
   ...:                                                                                               

In [3]: s = (10, 11, 12, 13) 
   ...: a = numpy.arange(numpy.prod(s)).reshape(s)                                                    

In [4]: %timeit get_buffer_data(a)                                                                    
2.72 µs ± 8.47 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [5]: %timeit get_buffer_nbytes(a, check_min_size=None, cuda_support=True)                          
3.7 µs ± 11.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: %timeit run(get_buffer_nbytes, a.T, check_min_size=None, cuda_support=True)                   
4.69 µs ± 18.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

</details>

<details>
<summary>Runtime <b>after</b> of <code>get_buffer_data</code> and <code>get_buffer_nbytes</code></summary>

<br>

```python
In [1]: import cupy 
   ...: import numpy 
   ...:  

   ...:     try: 
   ...:     except Exception: 
   ...:         pass 
   ...:                                                                                               

In [3]: s = (10, 11, 12, 13) 
   ...: a = numpy.arange(numpy.prod(s)).reshape(s)                                                    

In [4]: %timeit get_buffer_data(a)                                                                    
1.18 µs ± 9.89 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit get_buffer_nbytes(a, check_min_size=None, cuda_support=True)                          
1.66 µs ± 9.66 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit run(get_buffer_nbytes, a.T, check_min_size=None, cuda_support=True)                   
2.33 µs ± 8.45 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

</details>

Additionally we fast path objects through `get_buffer_nbytes` in particular if their `strides is None`. Currently CuPy, Numba, and NumPy already set `strides` to `None` if the data is C-contiguous. Plus serialization through Dask turns any non-contiguous array into a 1-D contiguous array. So any array that Dask gives us is sent through the fast path.

This was not true of cuDF `Buffer`s or RMM `DeviceBuffer`s for historical reasons (CuPy and Numba compatibility issues). Though these issues no longer apply, so have submitted PR ( https://github.com/rapidsai/cudf/pull/5917 ) and PR ( https://github.com/rapidsai/rmm/pull/477 ), which will fast path these objects in the future.

In short, I think collectively we've probably gotten as much out of this exercise as we can for the moment. We might be able to make further improvements by refactoring (contiguity check can be separate from size check). It's also possible a lightweight extension type could help us find the right balance between extracting the array interface and letting us query different common things easily as needed. Though this can probably be saved for a future PR.

ucp/_libs/utils.pyx
pentschev: But what happens if an array includes both `__array_interface__` and `__cuda_array_interface__`? It seems that the latter would never be used, though I don't know if there are any arrays that implement both.

pentschev: What would happen if an array implements both? It seems that then we would never try `__cuda_array_interface__`, though I don't know if this is valid and if some array type implements both.

pentschev: Sorry, I thought I thought I had lost my first comment, ignore the second which is basically the same.

jakirkham: Do you know of a case where this would happen? TBH I don’t know of any libraries outside of NumPy that use `__array_interface__`. Most libraries rely on the Python Buffer Protocol (typically used via `memoryview`s).

jakirkham: As NumPy also supports the Python Buffer Protocol, it might be worth dropping the `__array_interface__` handling here in favor of just routing that through a `memoryview` also.

pentschev: If the check for `__array_interface__` is exclusive to find whether it's a NumPy array, shouldn't we make it more clear by just checking `isinstance(buffer, np.ndarray)`? The `memoryview` idea makes sense as well, but it remains the question as to what order we should use to do the checking, `__cuda_array_interface__` or `memoryview` first.

jakirkham: If we do `memoryview`, we just drop the `__array_interface__` check. `memoryview` is already checked after `__cuda_array_interface__`.

jakirkham:
Switched to using `memoryview` for NumPy. This lets us get rid of another attribute check (already shown to be expensive). Plus we now check `__cuda_array_interface__` first (as the `memoryview` case was a fallback).

ucp/_libs/utils.pyx
jakirkham: `"__cuda_array_interface__"` now takes priority.

ucp/_libs/utils.pyx
jakirkham: `"__cuda_array_interface__"` now takes priority.

ucp/_libs/utils.pyx
jakirkham: Went ahead and switched NumPy to `memoryview` case in recent commits. So marking this resolved. Though if there are additional thoughts please feel free to start a new thread.

ucp/_libs/utils.pyx
jakirkham: `"__cuda_array_interface__"` is now higher priority.

jakirkham:
Have now also exposed the utility functions directly in Cython to leverage faster calling between Cython code. In Python we are still able to import the functions and use them from Python. So the Python code paths are unaffected.

quasiben:
I know I'm late but this is a really nice performance improvement!

PR #547, opened 2020-06-15
Fix debug tests and remove duplicate functions

pentschev:


debug-tests/debug_utils.py
```diff
@@ -41,15 +37,6 @@ def parse_args(args=None):
     return parser.parse_args()
 
 
-def get_num_gpus():
```
jakirkham, line 44: This appears to still be used [here]( https://github.com/rapidsai/ucx-py/blob/d93baf1f531c58b11ff620cff80ae9ac49d83104/debug-tests/debug_utils.py#L44 ). Should we keep it? Or were you planning to replace it with something else?

pentschev: Thanks for catching, that was duplicate in `utils`, I pushed another commit to import from there instead.

jakirkham: Got it. Thanks! 😄

PR #548, opened 2020-06-15
[REVIEW] Install dependencies via meta package

raydouglass:
Install dependencies via `rapids-build-env` (which is pre-installed in the containers).

This brings the repo in line with https://docs.rapids.ai/maintainers/depmgmt/

Depends on https://github.com/rapidsai/integration/pull/49

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -46,17 +46,13 @@ nvidia-smi
 logger "Activate conda env..."
 source activate gdf
 conda install "cudatoolkit=$CUDA_REL" \
-              "cupy>=6.5.0" "numpy>=1.16" \
               "cudf=${MINOR_VERSION}" "dask-cudf=${MINOR_VERSION}" \
-              "dask>=2.8.1" "distributed>=2.8.1" \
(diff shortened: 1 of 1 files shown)


quasiben:
I think this only passes after rapidsai/integration#49 is merged in.  I believe now it is picking up `autoconf/automake`  from the system instead of from conda-forge

jakirkham:
Should `ucx-py` be [listed in the build package]( https://github.com/rapidsai/integration/blob/f14f5c900fcb02ae5e43b180af3bf6b566f9a776/conda/recipes/rapids-build-env/meta.yaml#L122 ) if that package is being installed here?

raydouglass:
> Should `ucx-py` be [listed in the build package](https://github.com/rapidsai/integration/blob/f14f5c900fcb02ae5e43b180af3bf6b566f9a776/conda/recipes/rapids-build-env/meta.yaml#L122) if that package is being installed here?

I forgot we did that. Thanks, I'll close this PR.

jakirkham:
Would it make sense to drop `ucx-py` from the package and let others install in their CI as needed? This does seem like a nice improvement.

PR #549, opened 2020-06-16
Test `get_buffer_data` and `get_buffer_nbytes`

jakirkham:
Needed for PR ( https://github.com/rapidsai/ucx-py/pull/546 )

Adds some tests for the utility functions `get_buffer_data` and `get_buffer_nbytes` using both Python builtin objects and NumPy/CuPy arrays. Should exercise both the `memoryview` case as well as the `__array_interface__`/`__cuda_array_interface__` case.

### DIFF
--- a/tests/test_libs_utils.py
+++ b/tests/test_libs_utils.py
@@ -47,6 +47,7 @@ def test_get_buffer_nbytes_builtins(buffer):
             memoryview(buffer)[::2], check_min_size=None, cuda_support=True
         )
 
+    # Test exceptional cases with `check_min_size`
     get_buffer_nbytes(buffer, check_min_size=nbytes, cuda_support=True)
     with pytest.raises(ValueError):
         get_buffer_nbytes(buffer, check_min_size=(nbytes + 1), cuda_support=True)
--- a/tests/test_libs_utils.py
+++ b/tests/test_libs_utils.py
@@ -1,0 +1,48 @@
+import array
+import io
+import mmap
+
+import pytest
+from ucp._libs.utils import get_buffer_data, get_buffer_nbytes
+
+
+builtin_buffers = [
+    b"abcd",
+    array.array("I", [0, 1, 2, 3]),
+    memoryview(array.array("B", [0, 1, 2, 3, 4, 5])).cast("B", (3, 2)),
+    memoryview(b"abcd"),
+    memoryview(bytearray(b"abcd")),
+    io.BytesIO(b"abcd").getbuffer(),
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @quasiben @pentschev

tests/test_libs_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        iface_prop = "__array_interface__"
+
+    xp = pytest.importorskip(xp)
+
+    nelem = functools.reduce(operator.mul, shape, 1)
+    data = xp.arange(nelem, dtype=dtype)
+    arr = xp.ndarray(shape, dtype, data.data, strides=strides)
+    iface = getattr(arr, iface_prop)
+
+    return xp, arr, iface
```
pentschev, line 77: Seems like `xp` isn't currently used for anything. Just wanted to check this is intentional.

tests/test_libs_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    nbytes = memoryview(buffer).nbytes
+    result = get_buffer_nbytes(buffer, check_min_size=None, cuda_support=True)
+    assert result == nbytes
+
+    with pytest.raises(ValueError):
+        get_buffer_nbytes(
+            memoryview(buffer)[::2], check_min_size=None, cuda_support=True
+        )
+
+    get_buffer_nbytes(buffer, check_min_size=nbytes, cuda_support=True)
```
pentschev, line 50: Shouldn't we assert here that the result is indeed equal to `nbytes` ?

tests/test_libs_utils.py
jakirkham: Yeah I wasn't sure if we would need it or not. So left in the option. We could drop it if you prefer.

tests/test_libs_utils.py
jakirkham: No we already checked that above. This is checking something else. Namely that setting `check_min_size=nbytes` does not raise an exception.

jakirkham: Added a comment above this line, which should make that clearer.

tests/test_libs_utils.py
pentschev: No worries, just wanted to confirm.

tests/test_libs_utils.py
pentschev: Thanks!

pentschev:
LGTM, thanks @jakirkham .

PR #551, opened 2020-06-23
WIP: UCX RMA take two

MattBBaker:
Version 2 of the UCX RMA api. This one is still a bit rough around the edges, and need to add proper cleanup routines. It also creates a test, it's not properly integrated with pytest yet, but hopefully it illustrates how to use the RMA interface. Next step is making sure data structures are cleaned up/handled properly and then write some tests/benchmarks.

Edit to add:
Also, has a simple implementation of a UcxIO class that will proxy RMA reads/writes to look like a file handle.


GPUtester:

quasiben:
add to whitelist

MattBBaker:
Closing because this one won't be updated anymore.

PR #556, opened 2020-07-07
Endpoint Reuse

madsbk:
This PR implements a workaround of https://github.com/openucx/ucx/issues/3192 by introducing `EndpointReuse`:
```python
class EndpointReuse:
    """
    Class to seamlessly reuse endpoints. It uses the tag feature of send/recv
    to separate "virtual" endpoint pairs from each other.

    Connection Protocol
    -------------------
    1) Client connect to server using a new endpoint.
    2) Client send the IDs of all its existing endpoints.
    3) Server receives the IDs from the client and checks if it has a
       matching existing endpoint. It then sends the matching ID to the
       client or zero if no match.
    4) The client and server now continue with either the existing matching
       endpoints or the new endpoints (which are registered for later reuse).
    """
```

### DIFF
--- a/ucp/endpoint_reuse.py
+++ b/ucp/endpoint_reuse.py
@@ -15,14 +15,14 @@ class EPHandle:
 class EndpointReuse:
     """
     Class to seamlessly reuse endpoints. It uses the tag feature of send/recv
-    to separate "virtual" endpoint pars from eachother.
+    to separate "virtual" endpoint pairs from each other.
 
     Connection Protocol
     -------------------
     1) Client connect to server using a new endpoint.
     2) Client send the IDs of all its existing endpoints.
     3) Server receives the IDs from the client and checks if it has a
-       matching existing endpoint. It then sends the matching IDto the
+       matching existing endpoint. It then sends the matching ID to the
        client or zero if no match.
     4) The client and server now continue with either the existing matching
        endpoints or the new endpoints (which are registered for later reuse).
--- a/ucp/endpoint_reuse.py
+++ b/ucp/endpoint_reuse.py
@@ -1,4 +1,4 @@
-import asyncio
+import ctypes
 import uuid
 
 import numpy as np
@@ -41,7 +41,7 @@ class EndpointReuse:
 
     @classmethod
     async def create_endpoint(cls, ip, port):
-        tag = np.uint32(uuid.uuid4().int % np.iinfo(np.uint32).max)
+        tag = ctypes.c_uint32(uuid.uuid4().int).value
         ep_new = await core.create_endpoint(ip, port)
 
         existing_endpoints = list(cls.existing_endpoints.values())
--- a/ucp/endpoint_reuse.py
+++ b/ucp/endpoint_reuse.py
@@ -1,0 +1,154 @@
+import asyncio
+import uuid
+
+import numpy as np
+
+from . import core
+
+
+class EPHandle:
+    def __init__(self, ep):
+        self.ep = ep
+        self.refcount = 1
+
+
+class EndpointReuse:
+    """
+    Class to seamlessly reuse endpoints. It uses the tag feature of send/recv
+    to separate "virtual" endpoint pars from eachother.
+
+    Connection Protocol
+    -------------------
+    1) Client connect to server using a new endpoint.
+    2) Client send the IDs of all its existing endpoints.
+    3) Server receives the IDs from the client and checks if it has a
+       matching existing endpoint. It then sends the matching IDto the
+       client or zero if no match.
+    4) The client and server now continue with either the existing matching
+       endpoints or the new endpoints (which are registered for later reuse).
+    """
+
+    existing_endpoints = {}
+
+    def __init__(self, handle, tag, close_tag):
+        self.handle = handle
+        self.tag = tag
+        self.close_tag = close_tag
+        self.close_recv = asyncio.create_task(
+            handle.ep.recv(bytearray(1), tag=close_tag)
+        )
+
+    @classmethod
+    async def create_endpoint(cls, ip, port):
+        tag = np.uint32(uuid.uuid4().int % np.iinfo(np.uint32).max)
+        close_tag = np.uint32(uuid.uuid4().int % np.iinfo(np.uint32).max)
+        ep_new = await core.create_endpoint(ip, port)
+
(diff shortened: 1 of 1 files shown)


ucp/endpoint_reuse.py
```diff
@@ -0,0 +1,154 @@
...
+class EPHandle:
+    def __init__(self, ep):
+        self.ep = ep
+        self.refcount = 1
+
+
+class EndpointReuse:
+    """
+    Class to seamlessly reuse endpoints. It uses the tag feature of send/recv
+    to separate "virtual" endpoint pars from eachother.
```
jakirkham, line 18:
```suggestion
    to separate "virtual" endpoint pairs from each other.
```

ucp/endpoint_reuse.py
```diff
@@ -0,0 +1,154 @@
...
+
+    Connection Protocol
+    -------------------
+    1) Client connect to server using a new endpoint.
+    2) Client send the IDs of all its existing endpoints.
+    3) Server receives the IDs from the client and checks if it has a
+       matching existing endpoint. It then sends the matching IDto the
```
jakirkham, line 25:
```suggestion
       matching existing endpoint. It then sends the matching ID to the
```

jakirkham:
Would suggest softening the NumPy dependency by using things like the [`array`]( https://docs.python.org/3/library/array.html ), [`ctypes`]( https://docs.python.org/3/library/ctypes.html ), and/or [`struct`]( https://docs.python.org/3/library/struct.html ). Though this is probably a secondary concern to getting things working.

quasiben:
@cjnolet you may be interested in this feature

ucp/endpoint_reuse.py
```diff
@@ -0,0 +1,154 @@
...
+        self.tag = tag
+        self.close_tag = close_tag
+        self.close_recv = asyncio.create_task(
+            handle.ep.recv(bytearray(1), tag=close_tag)
+        )
+
+    @classmethod
+    async def create_endpoint(cls, ip, port):
+        tag = np.uint32(uuid.uuid4().int % np.iinfo(np.uint32).max)
+        close_tag = np.uint32(uuid.uuid4().int % np.iinfo(np.uint32).max)
```
jakirkham, line 44: As to the question raised earlier about `iinfo`, I think we can just cast these and it will only keep the relevant "lower" bits. IOW `iinfo` shouldn't be needed.

```python
In [1]: import ctypes, uuid                                                     

In [2]: i = uuid.uuid4().int; i                                                 
Out[2]: 287371484024313512047830668633417152082

In [3]: i % 2**32                                                               
Out[3]: 692024914

In [4]: ctypes.c_uint32(i)                                                      
Out[4]: c_uint(692024914)
```

```suggestion
        tag = ctypes.c_uint32(uuid.uuid4().int)
        close_tag = ctypes.c_uint32(uuid.uuid4().int)
```

madsbk: Thanks @jakirkham, added in https://github.com/rapidsai/ucx-py/pull/556/commits/b3b3dcd16049ec6c22dda6d9b98706aa418779dc

madsbk:
> Would suggest softening the NumPy dependency by using things like the [`array`](https://docs.python.org/3/library/array.html), [`ctypes`](https://docs.python.org/3/library/ctypes.html), and/or [`struct`](https://docs.python.org/3/library/struct.html). Though this is probably a secondary concern to getting things working.

Fixed by using https://github.com/rapidsai/ucx-py/pull/559, which also simplifies the code :)

quasiben:
@jakirkham did you have any additional comments ?

ucp/endpoint_reuse.py
```diff
@@ -0,0 +1,125 @@
...
+
+        existing_endpoints = list(cls.existing_endpoints.values())
+        my_ep_ids = []
+        for ep in existing_endpoints:
+            if not ep.ep.closed():
+                ep.refcount += 1
+                my_ep_ids.append(ep.ep._msg_tag_recv)
+
+        await ep_new.send_obj(pickle.dumps((my_ep_ids, tag)))
+        reuse_ep_id = pickle.loads(await ep_new.recv_obj())
```
jakirkham, line 59: Is it possible to get this to work without pickling? Similar question on other lines below.

ucp/endpoint_reuse.py
```diff
@@ -0,0 +1,125 @@
...
+            await ep_new.close()
+        else:
+            reuse_ep = EPHandle(ep_new)
+            assert ep_new._msg_tag_send not in cls.existing_endpoints
+            cls.existing_endpoints[ep_new._msg_tag_send] = reuse_ep
+        return cls(reuse_ep, tag)
+
+    @classmethod
+    def create_listener(cls, cb_coroutine, port):
+        async def _handle(ep_new):
```
jakirkham, line 79: Is it possible to implement this without a closure?

pentschev:
I was just wondering, in general UCX-Py is not thread-safe today, but this PR would be particularly problematic if we had multiple threads trying to reuse endpoints. I'm not sure if there's a use case for us today in which we may have multiple threads operating on endpoints, but we would have some hard time if for instance two threads try to read the same endpoint simultaneously, right? Please correct me if I'm mistaken here.

madsbk:
> I was just wondering, in general UCX-Py is not thread-safe today, but this PR would be particularly problematic if we had multiple threads trying to reuse endpoints. I'm not sure if there's a use case for us today in which we may have multiple threads operating on endpoints, but we would have some hard time if for instance two threads try to read the same endpoint simultaneously, right? Please correct me if I'm mistaken here.

Absolutely, at some point we should make UCX-Py thread-safe, which will require us to protect variables like `EndpointReuse.existing_endpoints` but also much in `Endpoint`, `init()`, etc.

ucp/endpoint_reuse.py
madsbk: Yes but we have to send a variable length list so I am not sure it is worth the extra complexity? 
Note, this is only part of the connection creation.

ucp/endpoint_reuse.py
madsbk: Yes, I can move `_handle` outside `EndpointReuse` and use `functools.partial()` but again I am not sure it is worth the extra code?

PR #557, opened 2020-07-07
Use the isort setup as cuDF

madsbk:
Fixed `isort` error in CI like this [one](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-style/detail/ucx-py-style/722/pipeline). 

Apparently, `THIRDPARTY` must be defined under `sections`. We now use the same order as [cuDF]( https://github.com/rapidsai/cudf/blob/6dc286ba67d2ec7e73674351be7011604b7a2fe7/python/cudf/setup.cfg#L15)

### DIFF
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -10,12 +10,14 @@ import pstats
 import sys
 from time import perf_counter as clock
 
+import cupy
+import numpy as np
+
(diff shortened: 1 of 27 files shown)


madsbk:
Thanks Peter

PR #559, opened 2020-07-08
Send/Recv object

madsbk:
This PR implements `send_obj()` and `recv_obj()`, which are convenient functions that exchange size information.

```python
async def recv_obj(self, tag=None, allocator=bytearray):
    """Receive from connected peer that calls `send_obj()`.

    As opposed to `recv()`, this function returns the received object.

    The transfer includes an extra message containing the size of `obj`,
    which increses the overhead slightly.

    Parameters
    ----------
    tag: hashable, optional
        Set a tag that must match the received message. Notice, currently
        UCX-Py doesn't support a "any tag" thus `tag=None` only matches a
        send that also sets `tag=None`.
    allocator: callabale, optional
        Function to allocate the recvied object. The function should
        take the number of bytes to allocate as input and return a new
        buffer of that size as output.

    Example
    -------
    >>> await pickle.loads(ep.recv_obj())
    """
```

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -677,7 +677,8 @@ class Endpoint:
     async def send_obj(self, obj, tag=None):
         """Send `obj` to connected peer that calls `recv_obj()`.
 
-        The message include size information of the object.
+        The transfer includes an extra message containing the size of `obj`,
+        which increses the overhead slightly.
 
         Parameters
         ----------
@@ -685,6 +686,10 @@ class Endpoint:
             The object to send.
         tag: hashable, optional
             Set a tag that the receiver must match.
+
+        Example
+        -------
+        >>> await ep.send_obj(pickle.dumps([1,2,3]))
         """
 
         nbytes = get_buffer_nbytes(
@@ -698,6 +703,9 @@ class Endpoint:
 
         As opposed to `recv()`, this function returns the received object.
 
+        The transfer includes an extra message containing the size of `obj`,
+        which increses the overhead slightly.
+
         Parameters
         ----------
         tag: hashable, optional
@@ -708,6 +716,10 @@ class Endpoint:
             Function to allocate the recvied object. The function should
             take the number of bytes to allocate as input and return a new
             buffer of that size as output.
+
+        Example
+        -------
+        >>> await pickle.loads(ep.recv_obj())
         """
         nbytes = bytearray(struct.calcsize("Q"))
         await self.recv(nbytes, tag=tag)
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -674,6 +674,48 @@ class Endpoint:
                 % (n, self._finished_recv_count)
             )
 
+    async def send_obj(self, obj, tag=None):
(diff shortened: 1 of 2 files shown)


quasiben:
This looks like a really nice convenience function.  Thanks @madsbk!  Would you be willing to add a test for an allocator other than `bytearray` ?  Does the allocator assume uint8  in the case of numpy/rmm ?

ucp/core.py
```diff
@@ -674,6 +674,60 @@ def close_after_n_recv(self, n, count_from_ep_creation=False):
...
+
+        The transfer includes an extra message containing the size of `obj`,
+        which increses the overhead slightly.
+
+        Parameters
+        ----------
+        tag: hashable, optional
+            Set a tag that must match the received message. Notice, currently
+            UCX-Py doesn't support a "any tag" thus `tag=None` only matches a
+            send that also sets `tag=None`.
```
pentschev, line 714: If `tag=None`, would that just hash `None` and use it as tag? If so, isn't that dangerous due to collisions? I think we had deadlocks due to collisions in the past.

ucp/core.py
```diff
@@ -674,6 +674,60 @@ def close_after_n_recv(self, n, count_from_ep_creation=False):
                 % (n, self._finished_recv_count)
             )
 
+    async def send_obj(self, obj, tag=None):
+        """Send `obj` to connected peer that calls `recv_obj()`.
+
+        The transfer includes an extra message containing the size of `obj`,
+        which increses the overhead slightly.
```
pentschev, line 681:
```suggestion
        which increases the overhead slightly.
```

ucp/core.py
```diff
@@ -674,6 +674,60 @@ def close_after_n_recv(self, n, count_from_ep_creation=False):
...
+
+        Parameters
+        ----------
+        tag: hashable, optional
+            Set a tag that must match the received message. Notice, currently
+            UCX-Py doesn't support a "any tag" thus `tag=None` only matches a
+            send that also sets `tag=None`.
+        allocator: callabale, optional
+            Function to allocate the recvied object. The function should
```
pentschev, line 716:
```suggestion
            Function to allocate the received object. The function should
```

ucp/core.py
madsbk: No,  `tag=None` is no tags but since UCX-Py uses tags to pair endpoint connections we don't support an _any tag_.

madsbk:
> This looks like a really nice convenience function. Thanks @madsbk! Would you be willing to add a test for an allocator other than `bytearray` ? Does the allocator assume uint8 in the case of numpy/rmm ?

Good point, I have added a Numpy test. 
Yes, the allocator should return a `nbytes` buffer. UCX-Py doesn't interpret the data so it is up to the user to _cast_ the received data to something meaningful.

jakirkham:
> UCX-Py doesn't interpret the data so it is up to the user to cast the received data to something meaningful.

FWIW users can usually do things like this...

```python
import numpy as np

a = np.empty((3,), dtype=float)
a.view("u1")
```

PR #560, opened 2020-07-08
[skip ci] Internals and monitoring docs

quasiben:
still very much a rough draft but this PR intends to add two docs pages:

1) Monitoring tools (i always forget how to look at IB traffic)
2) A page on tag-matching (@madsbk would you mind looking over this)


docs/source/ucx-intro.rst
```diff
@@ -0,0 +1,91 @@
+Send/Recv Internals
```
pentschev, line 1: Why do we have this a second time? Seems the same content as of https://github.com/rapidsai/ucx-py/pull/560/files#diff-6d59a7bf7a6266fb37c46a99dcc04b78 .

docs/source/send-recv.rst
```diff
@@ -0,0 +1,93 @@
+Send/Recv Internals
+===================
+
+Generally UCX creates connections between endpoints with the following steps:
+
+1. Create a ``Listener`` with defined ip address an port
+  a. ``Listener`` defines a callback function to process communications from endpoints
+2. Connect an ``Endpoint`` to the ``Listener``
+3. ``Endpoint`` sends and receives data with the ``Listener``
```
pentschev, line 9:
```suggestion
3. ``Endpoint`` communicates with the ``Listener``
```

docs/source/send-recv.rst
```diff
@@ -0,0 +1,93 @@
...
+|        while True:                                   |                                                          |
+|            obj = await ep.recv_obj()                 |     s.bind((HOST, PORT))                                 |
+|            await ep.send_obj(obj)                    |     s.listen(1)                                          |
+|                                                      |     conn, addr = s.accept()                              |
+|    lf = ucp.create_listener(echo_server, port)       |                                                          |
+|                                                      |     while True:                                          |
+|    while not lf.closed():                            |         data = conn.recv(1024)                           |
+|        await asyncio.sleep(0.1)                      |         if not data: break                               |
+|                                                      |         conn.sendall(data)                               |
++------------------------------------------------------+----------------------------------------------------------+
```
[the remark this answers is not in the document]
quasiben: How do you suggest we do that ?  Here we need to block forever, no?  Do you mean something like sending a close listener message ?

docs/source/ucx-intro.rst
quasiben: It's a mistake. I'll clean it up now

quasiben:
Thanks for all the reviewers @pentschev

docs/source/send-recv.rst
```diff
@@ -0,0 +1,93 @@
...
+
+    # echo_msg = await client.recv_obj()
+    [1594305226.988246] [dgx12:67497] UCXPY  DEBUG [Recv #000] ep: 0x7fef8bdc8048, tag: 0xabeef1db009f97af, nbytes: 8, type: <class 'bytearray'>
+    [1594305226.989332] [dgx12:67497] UCXPY  DEBUG [Recv #001] ep: 0x7fef8bdc8048, tag: 0xabeef1db009f97af, nbytes: 12, type: <class 'bytearray'>
+
+
+We can see from the above that when the ``Endpoint`` is created, 4 tags are generated:  ``msg-tag-send``, ``msg-tag-recv``, ``ctrl-tag-send``, and ``ctrl-tag-recv``.  This data is transmitted to the server via a `stream <http://openucx.github.io/ucx/api/v1.8/html/group___u_c_p___c_o_m_m.html#ga9022ff0ebb56cac81f6ba81bb28f71b3>`_ communication in an `exchange peer info <https://github.com/rapidsai/ucx-py/blob/6e1c1d201a382c689ca098c848cbfdc8237e1eba/ucp/core.py#L38-L89>`_ convenience function.
+
+Next, The client sends data on the ``msg-tag-send`` tag.  Two messages are sent, the size of the data ``8 bytes`` and data itself.  The server receives the data and immediately echos the data back.  Lastly, the client the receives two messages the size of the data an the data itself
```
madsbk, line 93:
```suggestion
Next, The client sends data on the ``msg-tag-send`` tag.  Two messages are sent, the size of the data ``8 bytes`` and data itself.  The server receives the data and immediately echos the data back.  Lastly, the client then receives two messages the size of the data an the data itself
```

docs/source/send-recv.rst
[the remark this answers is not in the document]
pentschev: What I had in mind is that endpoint send/recv and then close, same here after one recv/send. This is in line with what `ucx_perftest` does, although you can set the number of iterations there, but not important to set number of iterations here. This is only for completeness of the example, I'm sure people will see this example and forget to close if we don't show how it should be done.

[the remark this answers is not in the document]
quasiben: After chatting with @pentschev we added a note on how users should close the listener but it complicates the simple example

docs/source/send-recv.rst
```diff
@@ -0,0 +1,93 @@
+Send/Recv Internals
+===================
+
+Generally UCX creates connections between endpoints with the following steps:
+
+1. Create a ``Listener`` with defined IP address and port
+  a. ``Listener`` defines a callback function to process communications from endpoints
+2. Connect an ``Endpoint`` to the ``Listener``
+3. ``Endpoint`` communicates with the ``Listener``
```
pentschev, line 9: Maybe we should now add that the endpoint closes a connection, and eventually the listener too?

quasiben: Done

quasiben:
Thanks @madsbk and @pentschev for the reviews

PR #565, opened 2020-07-24
Disable UCP_FEATURE_WAKEUP in non-blocking mode

pentschev:
This is currently the only way to enable shared memory support.

### DIFF
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -189,7 +189,7 @@ cdef class UCXContext(UCXObject):
         ucp_context_h _handle
         dict _config
 
-    def __init__(self, config_dict):
+    def __init__(self, config_dict, blocking_progress_mode=False):
         cdef ucp_params_t ucp_params
         cdef ucp_worker_params_t worker_params
         cdef ucs_status_t status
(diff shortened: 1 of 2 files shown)


quasiben:
Looks good.  Currently failing on flake8:

> ./ucp/core.py:240:89: E501 line too long (106 > 88 characters)

Do you think it's worth exposing `ucp_params` to the Python layer.  Not to set but only for getting.  This would allow us to test and others to confirm UCP configurations.

pentschev:
Thanks for pinging @quasiben , I hadn't seen the error before, should be fixed now.

> Do you think it's worth exposing ucp_params to the Python layer. Not to set but only for getting. This would allow us to test and others to confirm UCP configurations.

I don't know, maybe? TBH, I'm ignorant of that, is that information useful at runtime somehow? Also note that `UCXWorker` too has a similar `worker_params` that aren't exposed. Maybe it would make more sense to open an issue to discuss if it's useful/worth to expose those?

pentschev:
Failures are legit, I'm trying to find out whether disabling `UCP_FEATURE_WAKEUP` is indeed supposed to work in all cases or if there are any limitations, which seems to be the case.

pentschev:
Ok, this is an issue in UCX 1.8.x but as per my local tests it works in 1.9.x, so we'll need to wait for that to merge this PR.

PR #568, opened 2020-08-03
Use cuDF's `assert_eq`

jakirkham:
Previously we were using Dask's `assert_eq`, which caused some issues when testing with cuDF's new nullable support. Here we fix this by relying on cuDF's `assert_eq`, which is designed to handle cuDF objects that we are passing.

cc @pentschev @quasiben


jakirkham:
Thanks for helping diagnose 🙂

PR #570, opened 2020-08-05
Fix-up CI requirements

jakirkham:
Cleans up requirements on CI.

* Use `ucx` package instead of building from source
* Drop/consolidate dependencies

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -53,7 +53,7 @@ conda install "cudatoolkit=$CUDA_REL" \
 
 # needed for asynccontextmanager in py36
 conda install -c conda-forge "async_generator" "autoconf" "automake" "libtool" \
-                              "cython>=0.29.14,<3.0.0a0" \
+                              "make" "cython>=0.29.14,<3.0.0a0" \
                               "pytest" "pkg-config" "pytest-asyncio" \
                               "pynvml" "libhwloc" "psutil"
 
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -52,7 +52,7 @@ conda install "cudatoolkit=$CUDA_REL" \
               -c rapidsai-nightly
 
 # needed for asynccontextmanager in py36
-conda install -c conda-forge "async_generator" "automake" "libtool" \
+conda install -c conda-forge "async_generator" "libtool" \
                               "automake" "autoconf" "cython>=0.29.14,<3.0.0a0" \
                               "pytest" "pkg-config" "pytest-asyncio" \
                               "pynvml" "libhwloc" "psutil"
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -46,12 +46,12 @@ nvidia-smi
 logger "Activate conda env..."
 source activate gdf
 conda install "cudatoolkit=$CUDA_REL" \
-              "cupy>=6.5.0" \
-              "cudf=${MINOR_VERSION}" "dask-cudf=${MINOR_VERSION}" \
-              "dask>=2.8.1" "distributed>=2.8.1" \
               "cython>=0.29.14,<3.0.0a0" \
               "libhwloc" "ucx" "ucx-proc=*=gpu" \
               "numpy>=1.16" \
+              "cupy>=6.5.0" \
+              "cudf=${MINOR_VERSION}" "dask-cudf=${MINOR_VERSION}" \
+              "dask>=2.8.1" "distributed>=2.8.1" \
               "pytest" "pytest-asyncio" \
               "pynvml" "psutil"
 
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -49,12 +49,11 @@ conda install "cudatoolkit=$CUDA_REL" \
               "cupy>=6.5.0" "numpy>=1.16" \
               "cudf=${MINOR_VERSION}" "dask-cudf=${MINOR_VERSION}" \
               "dask>=2.8.1" "distributed>=2.8.1" \
-              "ucx" "ucx-proc=*=gpu" \
-              -c rapidsai-nightly
+              "ucx" "ucx-proc=*=gpu"
 
-conda install -c conda-forge "cython>=0.29.14,<3.0.0a0" \
-                             "pytest" "pytest-asyncio" \
-                             "pynvml" "libhwloc" "psutil"
+conda install "cython>=0.29.14,<3.0.0a0" \
+              "pytest" "pytest-asyncio" \
+              "pynvml" "libhwloc" "psutil"
 
 # Install the master version of dask and distributed
 logger "pip install git+https://github.com/dask/distributed.git --upgrade --no-deps"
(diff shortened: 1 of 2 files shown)


ci/gpu/build.sh
```diff
@@ -52,8 +52,8 @@ conda install "cudatoolkit=$CUDA_REL" \
               -c rapidsai-nightly
 
 # needed for asynccontextmanager in py36
-conda install -c conda-forge "async_generator" "automake" "libtool" \
-                              "cmake" "automake" "autoconf" "cython>=0.29.14,<3.0.0a0" \
+conda install -c conda-forge "async_generator" "autoconf" "automake" "libtool" \
```
kkraus14, line 55: Please remove the `-c conda-forge` as channels are now set via a `.condarc` file.

Can any of these packages be pushed into the integration repo to avoid needing to pull them each time in this script?

jakirkham: Thanks for the info. Can you please provide context as to where this is done? Where can we see this?

We can't use the integration packages here as they require `ucx-py` ( https://github.com/rapidsai/ucx-py/pull/548#issuecomment-644287361 )

kkraus14: `.condarc` file is set in the gpuCI image here: https://github.com/rapidsai/gpuci-build-environment/blob/branch-0.15/rapidsai/devel.Dockerfile#L27-L47

pentschev:
@jakirkham I was thinking, these dev tools are required to build UCX, wouldn't it make more sense to drop all of them and start testing against UCX conda packages? At one point it made sense to build from source, but now we don't test UCX conda packages even in CI, so perhaps it would be a good idea to do so.

ci/gpu/build.sh
mike-wendt: > Thanks for the info. Can you please provide context as to where this is done? Where can we see this?
> 
> We can't use the integration packages here as they require `ucx-py` ( [#548 (comment)](https://github.com/rapidsai/ucx-py/pull/548#issuecomment-644287361) )

We have been wanting to remove `ucx-py` from `rapids-build-env`. If this would help in this case then we should get a list of all of the repos impacted by that change and make it happen. If you agree to the removal can you file an issue with ops and we'll start the process.

pentschev: As per https://github.com/rapidsai/ucx-py/pull/570#issuecomment-669016867, I think we should just stop building UCX from source and rely on UCX conda package, that way we'll make sure we test it to some extent as well which doesn't happen today.

jakirkham:
Yeah that makes sense Peter. I'm ok with just stopping building from source and using the `ucx` package. @madsbk @quasiben, thoughts? 🙂

madsbk:
> Yeah that makes sense Peter. I'm ok with just stopping building from source and using the `ucx` package. @madsbk @quasiben, thoughts? slightly_smiling_face

I think it is a great idea :)

quasiben:
+1 from me as well to drop dev tools and use conda the conda package

ci/gpu/build.sh
```diff
@@ -49,13 +49,13 @@ conda install "cudatoolkit=$CUDA_REL" \
...
               "dask>=2.8.1" "distributed>=2.8.1" \
+              "ucx" "ucx-proc=*=gpu" \
               -c rapidsai-nightly
 
 # needed for asynccontextmanager in py36
-conda install -c conda-forge "async_generator" "automake" "libtool" \
-                              "cmake" "automake" "autoconf" "cython>=0.29.14,<3.0.0a0" \
-                              "pytest" "pkg-config" "pytest-asyncio" \
-                              "pynvml" "libhwloc" "psutil"
+conda install -c conda-forge "async_generator" "cython>=0.29.14,<3.0.0a0" \
```
jakirkham, line 56: Does anyone know why we need `async_generator`? AIUI this is a backport package for Python 3.5 (Python 3.6+ already has it). Though the comment above references Python 3.6, we've also already dropped Python 3.6. Can we just drop this too?

jakirkham:
Thanks all! Dropped the build tools and the UCX build step. Have another question above.

ci/gpu/build.sh
jakirkham: Trying to drop this in the latest commit.

jakirkham: Looks like RAPIDS as a whole doesn't use this either. Proposing we drop this from the integration packages too ( https://github.com/rapidsai/integration/pull/92 ).

jakirkham: No failures from dropping it. So looks safe to leave out.

ci/gpu/build.sh
jakirkham: Thanks Mike! My guess is we are probably ok (especially now that the list has been significantly reduced here by using the `ucx` package). Plus it will probably help us to track these in `ucx-py` as we migrate things upstream to help make sure we don't miss any dependencies in the process.

ci/gpu/build.sh
```diff
@@ -46,16 +46,14 @@ nvidia-smi
...
               "cudf=${MINOR_VERSION}" "dask-cudf=${MINOR_VERSION}" \
               "dask>=2.8.1" "distributed>=2.8.1" \
-              -c rapidsai-nightly
-
-# needed for asynccontextmanager in py36
-conda install -c conda-forge "async_generator" "automake" "libtool" \
-                              "cmake" "automake" "autoconf" "cython>=0.29.14,<3.0.0a0" \
-                              "pytest" "pkg-config" "pytest-asyncio" \
-                              "pynvml" "libhwloc" "psutil"
+              "pynvml" "psutil"
```
mike-wendt, line 56: This still doesn't follow the conventions of the other repos. I'd rather see this instead:

```
conda install "cudatoolkit=$CUDA_REL" \
              "cudf=${MINOR_VERSION}" "dask-cudf=${MINOR_VERSION}" \
              "rapids-build-env=${MINOR_VERSION}"
```

All of your listed packages are in [rapids-build-env](https://github.com/rapidsai/integration/blob/branch-0.15/conda/recipes/rapids-build-env/meta.yaml) with versions shared by all of RAPIDS.

The only version that does not match is `cython` which if necessary can be updated [here](https://github.com/rapidsai/integration/blob/branch-0.15/conda/recipes/versions.yaml#L55)

FWIW `rapids-build-env` is already installed in the gpuCI images so this extra install is only needed to pick up any updates that have happened since they were built overnight.

ci/gpu/build.sh
mike-wendt: Approved the PR for integration

ci/gpu/build.sh
jakirkham: Submitted the `cython` requirement in PR ( https://github.com/rapidsai/integration/pull/93 ).

Though I think we are still in awkward situation with `rapids-build-env` because of `ucx-py` being included. That said, it's interesting to hear `rapids-build-env` is already included. Maybe we should work on dropping `ucx-py` from `rapids-build-env` then?

Independently there's still value in tracking dependencies of `ucx-py` outside of normal RAPIDS processes as we will need to include these when upstreaming to UCX.

pentschev: I'm +1 on https://github.com/rapidsai/integration/pull/94, UCX-Py is not mandatory for most RAPIDS packages.

jakirkham: This is done in the latest commit. Thanks for reviewing all of those PRs Mike! 😄

jakirkham:
Thanks all! 😄

PR #573, opened 2020-08-11
Use `getattr` instead of `hasattr`

jakirkham:
Calling `hasattr` is just as expensive as calling `getattr` and slightly more expensive than direct attribute access. As a result calling `hasattr` and then perform attribute access, we are effectively paying the cost of attribute access twice.

<details>
<summary><code>hasattr</code> vs. <code>getattr</code> vs. direct access</summary>

```python
In [1]: import numpy                                                            

In [2]: a = numpy.arange(110).reshape(10, 11)                                   

In [3]: %timeit hasattr(a, "__array_interface__")                               
1.17 µs ± 3.73 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit getattr(a, "__array_interface__", None)                         
1.17 µs ± 5.61 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit a.__array_interface__                                           
1.11 µs ± 2.27 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

</details>

When we look at the overall runtime of `get_buffer_data` and `get_buffer_nbytes`, we can see that attribute access makes a significant contribution to runtime of these functions overall.

<details>
<summary>Runtime <b>before</b> of <code>get_buffer_data</code> and <code>get_buffer_nbytes</code></summary>

```python
In [1]: import numpy 
   ...:  

In [2]: a = numpy.arange(110).reshape(10, 11)                                   

In [3]: %timeit get_buffer_data(a)                                              
2.48 µs ± 12.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: %timeit get_buffer_nbytes(a, check_min_size=None, cuda_support=True)    
3.67 µs ± 17.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

</details>

So we can simply switch to `getattr` and add a default case of `None` to check for. Overall this is a pretty minor change to the code here. Though the result in terms of `get_buffer_data` and `get_buffer_nbytes` is quite notable, making both ~2x faster.

<details>
<summary>Runtime <b>after</b> of <code>get_buffer_data</code> and <code>get_buffer_nbytes</code></summary>

```python
In [1]: import numpy 
   ...:  

In [2]: a = numpy.arange(110).reshape(10, 11)                                   

In [3]: %timeit get_buffer_data(a)                                              
1.16 µs ± 12.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit get_buffer_nbytes(a, check_min_size=None, cuda_support=True)    
2.13 µs ± 20.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

</details>

cc @pentschev @madsbk @quasiben

### DIFF
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -1,4 +1,4 @@
-# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
 # See file LICENSE for terms.
 
 # cython: language_level=3
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -70,8 +70,9 @@ def get_buffer_nbytes(buffer, check_min_size, cuda_support):
         shape = [int(s) for s in iface["shape"]]
         nbytes = reduce(operator.mul, shape, 1) * itemsize
         # Check that data is contiguous
-        if len(shape) > 0 and iface.get("strides") is not None:
-            strides = [int(s) for s in iface['strides']]
+        strides = iface.get("strides")
+        if len(shape) > 0 and strides is not None:
+            strides = [int(s) for s in strides]
             if len(strides) != len(shape):
                 msg = "The length of shape and strides must be equal"
                 raise ValueError(msg)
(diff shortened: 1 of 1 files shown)


jakirkham:
Absorbed into PR ( https://github.com/rapidsai/ucx-py/pull/546 ).

PR #580, opened 2020-08-14
Some cleanup of topological distance

jakirkham:
Adds a `.pxd` file for `topological_distance`. Types methods arguments and variables in methods. Drops some casts where the type was unchanged.

### DIFF
--- a/ucp/_libs/topological_distance.pxd
+++ b/ucp/_libs/topological_distance.pxd
@@ -10,3 +10,10 @@ from .topological_distance_dep cimport *
 cdef class TopologicalDistance:
     cdef:
         hwloc_topology_t topo
+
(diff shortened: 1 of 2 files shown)


pentschev:
LGTM, but TBH, I'm still ignorant of most of how Cython code should be structured. Could you clarify what is this solving/improving?

jakirkham:
Most of the performance that one gets out of Cython relies on correctly typing variables as tightly as possible as early as possible. This avoids needing to work with Python objects, which require a fair bit of extra handling. Thus allowing Cython to make better choices about what to do with these variables.

pentschev:
Got it, thanks @jakirkham for clarifying! 😄

jakirkham:
Of course. Thanks for the review! 😄

Issue #578, opened 2020-08-13
Failure with RMM (post CNMeM deprecation)


jakirkham:
Have been seeing the following failures on CI and locally with [the benchmarks]( https://github.com/rapidsai/ucx-py/tree/branch-0.15/benchmarks ) since PR ( https://github.com/rapidsai/rmm/pull/466 ) was merged. A fair bit of history is documented in PR ( https://github.com/rapidsai/ucx-py/pull/575 ), a style PR, where this first came up. Though it can be reproduced outside of that PR.

Tried specifying the memory pool size to be close to the amount of memory on the GPU (with some space still free for other miscellaneous usage). However that still fails. Tried a pool size of `0`, which also fails. Changing the chunk size also fails. All with the same error.

In the send/recv case using different devices for server and client avoids the issue. Also using the CuPy memory pool instead of RMM's works fine in send/recv.

At a bit of a loss for what is happening here, would be great to have another set of eyes 🙂 

<details>
<summary>Dataframe merge benchmark failure:</summary>

```bash
python benchmarks/cudf-merge.py --chunks-per-dev 4 --chunk-size 10000
```

```python
Process SpawnProcess-1:
Traceback (most recent call last):
  File "/datasets/jkirkham/miniconda/envs/rapids15dev/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/datasets/jkirkham/miniconda/envs/rapids15dev/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/datasets/jkirkham/miniconda/envs/rapids15dev/lib/python3.8/site-packages/ucp/utils.py", line 163, in _worker_process
    ret = loop.run_until_complete(run())
  File "/datasets/jkirkham/miniconda/envs/rapids15dev/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/datasets/jkirkham/miniconda/envs/rapids15dev/lib/python3.8/site-packages/ucp/utils.py", line 158, in run
    return await func(rank, eps, args)
  File "/datasets/jkirkham/devel/ucx-py/benchmarks/cudf-merge.py", line 169, in worker
    df1 = generate_chunk(rank, args.chunk_size, args.n_chunks, "build", args.frac_match)
  File "/datasets/jkirkham/devel/ucx-py/benchmarks/cudf-merge.py", line 114, in generate_chunk
[...]

PR #581, opened 2020-08-14
Use 2 MB pool for cuDF merge

jakirkham:
This entire computation uses a bit under 2 MB of space. So bump the pool size so that all the allocations should fit comfortably within it.

xref: https://github.com/rapidsai/ucx-py/issues/578#issuecomment-674309920

cc @quasiben


harrism:
Why would you set the RMM pool size so small -- 100 bytes! Even 2MB is tiny. Why not leave the default?

PR #583, opened 2020-08-18
[NO MRG] Enable RMM logging

jakirkham:
Makes changes to include RMM logging during benchmarks.

cc @harrism


pentschev:
Can we make `logging=` an CLI argument?

jakirkham:
We certainly could.

Was mainly putting this up while debugging issues with Mark yesterday. So wasn't necessarily trying to get it in. Though no objection to figuring out how to include this if there is broader interest 🙂

pentschev:
I'll leave it up to you, if it's only for testing and we won't merge, then don't worry about it. :)

PR #592, opened 2020-08-28
Use Python C API in util functions w/host buffers

jakirkham:
Leverages the `Py_buffer` object held by `memoryview` objects to perform some computations and checks in C. Also gets the `memoryview` using `PyMemoryView_FromObject`, which has less overhead.

### DIFF
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -4,7 +4,11 @@
 # cython: language_level=3
 
 
-from cpython.memoryview cimport PyMemoryView_GET_BUFFER
+from cpython.buffer cimport PyBuffer_IsContiguous
+from cpython.memoryview cimport (
+    PyMemoryView_FromObject,
+    PyMemoryView_GET_BUFFER,
+)
 from cython cimport boundscheck, wraparound
 from libc.stdint cimport uintptr_t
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Previous benchmark can be seen [here]( https://github.com/rapidsai/ucx-py/pull/546#issuecomment-673614364 ). This makes a small, but notable improvement on that result.

<details>
<summary>NumPy after:</summary>

```python
In [1]: import cupy
   ...: import numpy
   ...: 

   ...:     try:
   ...:     except Exception:
   ...:         pass
   ...: 

In [3]: s = (10, 11, 12, 13)
   ...: a = numpy.arange(numpy.prod(s)).reshape(s)

In [4]: %timeit get_buffer_data(a)
542 ns ± 1.41 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit get_buffer_nbytes(a, check_min_size=None, cuda_support=True)
637 ns ± 3.29 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit run(get_buffer_nbytes, a.T, check_min_size=None, cuda_support=True)
1.36 µs ± 9.76 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

</details>

PR #593, opened 2020-09-03
Further streamline `get_buffer_data`

jakirkham:
Streamlines the C-code generated by Cython for `get_buffer_data` to avoid unnecessary code generation.

<hr>

A short benchmark can be seen below. Makes a small, but notable improvement to the `__cuda_array_interface__` case. The `memoryview` case is [unchanged from before]( https://github.com/rapidsai/ucx-py/pull/592#issuecomment-683166572 ). The previous benchmark with CuPy can be seen [here]( https://github.com/rapidsai/ucx-py/pull/546#issuecomment-673614364 ).

```python
In [1]: import cupy
   ...: import numpy
   ...: 

In [2]: s = (10, 11, 12, 13)

In [3]: %timeit get_buffer_data(a_c)
548 ns ± 4.85 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit get_buffer_data(a_g)
852 ns ± 4.39 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

### DIFF
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -25,7 +25,7 @@ cpdef uintptr_t get_buffer_data(buffer, bint check_writable=False) except *:
     cdef uintptr_t data_ptr
     cdef bint data_readonly
     if iface is not None:
-        data_ptr, data_readonly = iface["data"]
+        data_ptr, data_readonly = <tuple>iface["data"]
     else:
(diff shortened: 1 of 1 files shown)


madsbk (approved):
LGTM, nice work @jakirkham.

quasiben:
I love these small and impactful improvements!  Thank you @jakirkham

PR #594, opened 2020-09-04
Add fence/flush/worker address APIs

MattBBaker:
First patch to upstream #551 in smaller chunks with a cleaned up API. This brings in all the foundational APIs for RMA short of the changes to end point creation, which a topic I expect to be more controversial than the pure addition changes and get some testing to make sure it doesn't break anything.

In this patch it exposes:

- ucp_worker_fence as ucp.fence()
- ucp_worker_flush_nb as ucp.flush()
- ucp_ep_flush_nb as ep.flush()
- ucp_worker_get_address() as ucp.get_worker_address()
- ucp_worker_release_address() as a destructor in the UCXAddress object

With thanks to @jglaser for bug fixing.

### DIFF
--- a/tests/test_rma.py
+++ b/tests/test_rma.py
@@ -19,9 +19,11 @@ async def test_flush(blocking_progress_mode):
     await ucp.flush()
     ucp.reset()
 
+
 @pytest.mark.parametrize("blocking_progress_mode", [True, False])
 def test_worker_address(blocking_progress_mode):
     ucp.init(blocking_progress_mode=blocking_progress_mode)
 
     addr = ucp.get_worker_address()
+    assert addr is not None
     ucp.reset()
--- a/tests/test_rma.py
+++ b/tests/test_rma.py
@@ -1,0 +1,27 @@
+import pytest
+
+import ucp
+
+
+@pytest.mark.parametrize("blocking_progress_mode", [True, False])
+def test_fence(blocking_progress_mode):
+    ucp.init(blocking_progress_mode=blocking_progress_mode)
+#   this should always succeed
+    ucp.fence()
+    ucp.reset()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("blocking_progress_mode", [True, False])
+async def test_flush(blocking_progress_mode):
+    ucp.init(blocking_progress_mode=blocking_progress_mode)
+
+    await ucp.flush()
+    ucp.reset()
+
+@pytest.mark.parametrize("blocking_progress_mode", [True, False])
+def test_worker_address(blocking_progress_mode):
+    ucp.init(blocking_progress_mode=blocking_progress_mode)
+
+    addr = ucp.get_worker_address()
+    ucp.reset()
--- a/tests/test_rma.py
+++ b/tests/test_rma.py
@@ -1,6 +1,50 @@
+import numpy as np
 import pytest
+import asyncio
 
 import ucp
+
+
+def handle_exception(loop, context):
+    msg = context.get("exception", context["message"])
+    print(msg)
+
+
+# Let's make sure that UCX gets time to cancel
+# progress tasks before closing the event loop.
+@pytest.yield_fixture()
+def event_loop(scope="function"):
+    loop = asyncio.new_event_loop()
+    loop.set_exception_handler(handle_exception)
+    ucp.reset()
+    yield loop
+    ucp.reset()
+    loop.run_until_complete(asyncio.sleep(0))
+    loop.close()
+
+
+def make_echo_server(create_empty_data):
+    """
+    Returns an echo server that calls the function `create_empty_data(nbytes)`
+    to create the data container.`
+    """
+
+    async def echo_server(ep):
+        """
+        Basic echo server for sized messages.
+        We expect the other endpoint to follow the pattern::
+        # size of the real message (in bytes)
+        >>> await ep.send(msg_size, np.uint64().nbytes)
+        >>> await ep.send(msg, msg_size)       # send the real message
+        >>> await ep.recv(responds, msg_size)  # receive the echo
+        """
+        msg_size = np.empty(1, dtype=np.uint64)
+        await ep.recv(msg_size)
+        msg = create_empty_data(msg_size[0])
+        await ep.recv(msg)
+        await ep.send(msg)
+
+    return echo_server
 
 
 @pytest.mark.parametrize("blocking_progress_mode", [True, False])
@@ -27,3 +71,20 @@ def test_worker_address(blocking_progress_mode):
     addr = ucp.get_worker_address()
     assert addr is not None
     ucp.reset()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("blocking_progress_mode", [True, False])
+async def test_send_recv_addr(blocking_progress_mode):
+    ucp.init(blocking_progress_mode=blocking_progress_mode)
+
+    msg = ucp.get_worker_address()
+    msg_size = np.array([len(bytes(msg))], dtype=np.uint64)
+    listener = ucp.create_listener(make_echo_server(lambda n: bytearray(n)))
+    client = await ucp.create_endpoint(ucp.get_address("wlp7s0"), listener.port)
+
+    await client.send(msg_size)
+    await client.send(msg)
+    resp = bytearray(msg_size[0])
+    await client.recv(resp)
+    assert resp == bytes(msg)
--- a/tests/test_rma.py
+++ b/tests/test_rma.py
@@ -1,6 +1,7 @@
+import asyncio
+
 import numpy as np
 import pytest
-import asyncio
 
 import ucp
 
--- a/tests/test_rma.py
+++ b/tests/test_rma.py
@@ -81,7 +81,7 @@ async def test_send_recv_addr(blocking_progress_mode):
     msg = ucp.get_worker_address()
     msg_size = np.array([len(bytes(msg))], dtype=np.uint64)
     listener = ucp.create_listener(make_echo_server(lambda n: bytearray(n)))
-    client = await ucp.create_endpoint(ucp.get_address("wlp7s0"), listener.port)
+    client = await ucp.create_endpoint(ucp.get_address(), listener.port)
 
     await client.send(msg_size)
     await client.send(msg)
--- a/tests/test_rma.py
+++ b/tests/test_rma.py
@@ -6,7 +6,7 @@ import ucp
 @pytest.mark.parametrize("blocking_progress_mode", [True, False])
 def test_fence(blocking_progress_mode):
     ucp.init(blocking_progress_mode=blocking_progress_mode)
-#   this should always succeed
+    # this should always succeed
     ucp.fence()
     ucp.reset()
 
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -10,8 +10,8 @@ import weakref
 
 from posix.stdio cimport open_memstream
 
-from cpython.ref cimport Py_DECREF, Py_INCREF, PyObject
 from cpython.buffer cimport PyBUF_WRITABLE, PyBUF_ND, PyBUF_FORMAT
+from cpython.ref cimport Py_DECREF, Py_INCREF, PyObject
 from libc.stdint cimport uintptr_t
 from libc.stdio cimport FILE, fclose, fflush
 from libc.stdlib cimport free
@@ -460,8 +460,10 @@ cdef class UCXAddress(UCXObject):
         buffer.strides = NULL
         buffer.suboffsets = NULL
         buffer.internal = NULL
+
     def __releasebuffer__(self, Py_buffer *buffer):
         pass
+
 
 def _ucx_endpoint_finalizer(uintptr_t handle_as_int, worker, inflight_msgs):
     assert worker.initialized
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -424,7 +424,7 @@ cdef class UCXAddress(UCXObject):
 
     def __init__(self, UCXWorker worker):
         cdef ucs_status_t status
-        cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
+        cdef ucp_worker_h ucp_worker = worker._handle
         cdef size_t length
 
         status = ucp_worker_get_address(ucp_worker, &self._handle, &length)
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -543,7 +543,11 @@ cdef class UCXEndpoint(UCXObject):
         assert self.initialized
         return int(<uintptr_t>self._handle)
 
-    def flush(self, cb_func, cb_args, cb_kwargs=dict()):
+    def flush(self, cb_func, cb_args=None, cb_kwargs=None):
+        if cb_args is None:
+            cb_args = ()
+        if cb_kwargs is None:
+            cb_kwargs = {}
         cdef ucs_status_ptr_t req
         cdef ucp_send_callback_t _send_cb = <ucp_send_callback_t>_send_callback
 
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -1,4 +1,5 @@
 # Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020       UT-Battelle, LLC. All rights reserved.
 # See file LICENSE for terms.
 
 # cython: language_level=3
@@ -386,6 +387,64 @@ cdef class UCXWorker(UCXObject):
         assert_ucs_status(status)
         return UCXEndpoint(self, <uintptr_t>ucp_ep)
 
+    def fence(self):
+        cdef ucs_status_t status
+        status = ucp_worker_fence(self._handle)
+        assert_ucs_status(status)
+        return status
+
+    def flush(self, cb_func, cb_args, cb_kwargs=dict()):
+        cdef ucs_status_ptr_t req
+        cdef ucp_send_callback_t _send_cb = <ucp_send_callback_t>_send_callback
+
+        cdef ucs_status_ptr_t status = ucp_worker_flush_nb(self._handle, 0, _send_cb)
+        return _handle_status(
+            status, 0, cb_func, cb_args, cb_kwargs, 'flush', self._inflight_msgs
+        )
+
+    def get_address(self):
+        return UCXAddress(self)
+
+
+def _ucx_address_finalizer(uintptr_t handle_as_int, UCXWorker worker):
+    cdef ucp_address_t *address = <ucp_address_t *>handle_as_int
+    cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
+    ucp_worker_release_address(ucp_worker, address)
+
+
+cdef class UCXAddress(UCXObject):
+    """Python representation of ucp_address_t"""
+    cdef ucp_address_t *_handle
+    cdef size_t _length
+    cdef worker
+    __array_interface__ = dict()
+
+    def __init__(self, UCXWorker worker):
+        cdef ucs_status_t status
+        cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
+
+        status = ucp_worker_get_address(ucp_worker, &self._handle, &self._length)
+        assert_ucs_status(status)
+        self.worker = worker
+        self.__array_interface__["data"] = (<uintptr_t>self._handle, True)
+        self.__array_interface__["typestr"] = "|V1"
+        self.__array_interface__["shape"] = [self._length]
+        self.__array_interface__["version"] = 3
+        self.add_handle_finalizer(
+            _ucx_address_finalizer,
+            int(<uintptr_t>self._handle),
+	    worker
+        )
+        worker.add_child(self)
+
+    @property
+    def address(self):
+        return <uintptr_t>self._handle
+
+    @property
+    def length(self):
+        return int(self._length)
+
 
 def _ucx_endpoint_finalizer(uintptr_t handle_as_int, worker, inflight_msgs):
     assert worker.initialized
@@ -455,6 +514,16 @@ cdef class UCXEndpoint(UCXObject):
     def handle(self):
         assert self.initialized
         return int(<uintptr_t>self._handle)
+
+    def flush(self, cb_func, cb_args, cb_kwargs=dict()):
+        cdef ucs_status_ptr_t req
+        cdef ucp_send_callback_t _send_cb = <ucp_send_callback_t>_send_callback
+
+        cdef ucs_status_ptr_t status = ucp_ep_flush_nb(self._handle, 0, _send_cb)
+        return _handle_status(
+            status, 0, cb_func, cb_args, cb_kwargs, 'flush', self._inflight_msgs
+        )
+
 
 
 cdef void _listener_callback(ucp_conn_request_h conn_request, void *args):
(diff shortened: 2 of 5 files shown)


GPUtester:

quasiben:
add to whitelist

quasiben:
Thank you so much for pushing here @MattBBaker !  It's really appreciated.  @jakirkham is out today but he would be best to review the cython changes.  For now can I ask that you add tests for fence/flush ?

MattBBaker:
@quasiben Just added the tests.

MattBBaker:
The test failure seems weird. The output just reads,
```
>>>> FAILED: black style check; begin output











>>>> FAILED: black style check; end output
```

quasiben:
That is strange output -- not sure what is happening there.  If you install black `python -m pip install black` and run `black .` in the root it should take care of linting issues.

I see the following when I run and do `git diff`:

```
diff --git a/tests/test_rma.py b/tests/test_rma.py
index ea78199..e279b51 100644
--- a/tests/test_rma.py
+++ b/tests/test_rma.py
@@ -6,7 +6,7 @@ import ucp
 @pytest.mark.parametrize("blocking_progress_mode", [True, False])
 def test_fence(blocking_progress_mode):
     ucp.init(blocking_progress_mode=blocking_progress_mode)
-#   this should always succeed
+    #   this should always succeed
     ucp.fence()
     ucp.reset()

diff --git a/ucp/comm.py b/ucp/comm.py
index 03006e0..3ea9ffb 100644
--- a/ucp/comm.py
+++ b/ucp/comm.py
@@ -85,17 +85,9 @@ def stream_recv(
     )


-def flush_worker(
-        worker: ucx_api.UCXWorker, event_loop=None
-) -> asyncio.Future:
-    return _call_ucx_api(
-        event_loop, worker.flush
-    )
+def flush_worker(worker: ucx_api.UCXWorker, event_loop=None) -> asyncio.Future:
+    return _call_ucx_api(event_loop, worker.flush)


-def flush_ep(
-        ep: ucx_api.UCXEndpoint, event_loop=None
-) -> asyncio.Future:
-    return _call_ucx_api(
-        event_loop, ep.flush
-    )
+def flush_ep(ep: ucx_api.UCXEndpoint, event_loop=None) -> asyncio.Future:
+    return _call_ucx_api(event_loop, ep.flush)
```

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +387,64 @@ cdef class UCXWorker(UCXObject):
         assert_ucs_status(status)
         return UCXEndpoint(self, <uintptr_t>ucp_ep)
 
+    def fence(self):
+        cdef ucs_status_t status
+        status = ucp_worker_fence(self._handle)
+        assert_ucs_status(status)
+        return status
```
jakirkham, line 394: Not sure if we plan to call this from the C-level as well. If so, would suggest marking it as `cpdef`, which allows the method to be called from both Python-level code (`.py` files) and C-level code (`.pyx` files particularly in other `cdef` or `cpdef` functions/methods). Using `cpdef` allows Cython to use a C-style function call (or function pointer call with methods), which can be much faster.

Added the return type as well. `except *` means a Python exception could be thrown (`assert_ucs_status` may do this). So Cython will insert code to check for a Python exception in that case.

```suggestion
    cpdef ucs_status_t fence(self) except *:
        cdef ucs_status_t status = ucp_worker_fence(self._handle)
        assert_ucs_status(status)
        return status
```

Note: May want to make similar changes to other functions/methods below depending on where you can plan to use them as well.

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +387,64 @@ cdef class UCXWorker(UCXObject):
...
+        self.__array_interface__["typestr"] = "|V1"
+        self.__array_interface__["shape"] = [self._length]
+        self.__array_interface__["version"] = 3
+        self.add_handle_finalizer(
+            _ucx_address_finalizer,
+            int(<uintptr_t>self._handle),
+            worker
+        )
+        worker.add_child(self)
+
```
jakirkham, line 439: If the goal is to have this be accessible to things like NumPy, would implement [the Python Buffer Protocol]( https://docs.python.org/3/c-api/buffer.html ), which not only NumPy uses, but Python itself uses. Cython has [native support]( http://docs.cython.org/en/latest/src/userguide/buffer.html ) for this as well. Have sketched out an implementation below.

```suggestion

    def __getbuffer__(self, Py_buffer *buffer, int flags):
        if (flags & PyBUF_WRITABLE) == PyBUF_WRITABLE:
            raise BufferError("Requested writable view on readonly data")

        buffer.buf = <void*>self._handle
        buffer.obj = self
        buffer.len = self._length * sizeof(ucp_address_t)
        buffer.readonly = True
        buffer.itemsize = sizeof(ucp_address_t)
        if (flags & PyBUF_FORMAT) == PyBUF_FORMAT:
            buffer.format = b"P"
        else:
            buffer.format = NULL
        buffer.ndim = 1
        if (flags & PyBUF_ND) == PyBUF_ND:
            buffer.shape = &self._length
        else:
            buffer.shape = NULL
        buffer.strides = NULL
        buffer.suboffsets = NULL
        buffer.internal = NULL

    def __releasebuffer__(self, Py_buffer *buffer):
        pass
```

Note the format value comes from [this table]( https://docs.python.org/3/library/struct.html#format-characters ). Also `buffer.strides = NULL` is shorthand for C-contiguous data.

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +387,64 @@ cdef class UCXWorker(UCXObject):
...
+        cdef ucs_status_t status
+        cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
+
+        status = ucp_worker_get_address(ucp_worker, &self._handle, &self._length)
+        assert_ucs_status(status)
+        self.worker = worker
+        self.__array_interface__["data"] = (<uintptr_t>self._handle, True)
+        self.__array_interface__["typestr"] = "|V1"
+        self.__array_interface__["shape"] = [self._length]
+        self.__array_interface__["version"] = 3
```
jakirkham, line 432: This can be dropped as the Python Buffer Protocol implementation below handles this.

jakirkham (commented):
Thanks Matt! Sorry for the delay. Have added a few suggestions below. Please let us know if you have any questions 🙂

ucp/_libs/ucx_api.pyx
MattBBaker: I don't think it will be used far in the Cython parts, It could well be and I definitely needed add the except part, since fence should expose fencing problems as exceptions.

ucp/_libs/ucx_api.pyx
MattBBaker: I always meant to do this as a buffer protocol, but the first time I tried that didn't work out so well so I went with the numpy array interface to 'just make it work'. I think I work with your suggestion below.

quasiben:
@jakirkham is there anything left here or is this good to merge in ?

jakirkham:
Still waiting on the changes mentioned above. Matt mentioned he would look at these. Not sure if he's had time to do so though.

MattBBaker:
Had an presentation in IWOMP so that consumed most of my cycles until yesterday. I'm putting the final touches on this patch use the python buffer protocol. Expect a new push Soon(TM)

jakirkham:
No worries. Just let us know 🙂

MattBBaker:
So, the style checker is unhappy with the pyx file, and when I run the isort tool on it the tool with produce garbage because it doesn't understand cimport. Other than that it looks good

@jakirkham @quasiben

quasiben:
@MattBBaker hope you don't mind.  I pushed the isort fix to the PR on your behalf so you wouldn't have to deal with more linting issues

jakirkham:
Let me take another look and get back to you. Should be able to do that today

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,82 @@ cdef class UCXWorker(UCXObject):
...
+        buffer.len = self._length
+        buffer.readonly = True
+        buffer.itemsize = self._length
+        if (flags & PyBUF_FORMAT) == PyBUF_FORMAT:
+            buffer.format = b"p"
+        else:
+            buffer.format = NULL
+        buffer.ndim = 1
+        if (flags & PyBUF_ND) == PyBUF_ND:
+            buffer.shape = <Py_ssize_t *>&self._length
```
jakirkham, line 457: One thing to note here is `Py_ssize_t` is somewhat analogous to `ssize_t` meaning it is `signed` as opposed to `unsigned`. That might cause us some issues with how this is interpreted. Might be better to type `self._length` as `Py_ssize_t` and deal with a small overhead going back and forth between `size_t` as needed.

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,82 @@ cdef class UCXWorker(UCXObject):
...
+def _ucx_address_finalizer(uintptr_t handle_as_int, UCXWorker worker):
+    cdef ucp_address_t *address = <ucp_address_t *>handle_as_int
+    cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
+    ucp_worker_release_address(ucp_worker, address)
+
+
+cdef class UCXAddress(UCXObject):
+    """Python representation of ucp_address_t"""
+    cdef ucp_address_t *_handle
+    cdef size_t _length
```
jakirkham, line 418: As noted elsewhere, would suggest making this `Py_ssize_t` as it will make it a bit easier when working with the Python side of things. We may pay a small cost going to/from `size_t` on the C side, but it is probably worth it in the long run.

```suggestion
    cdef Py_ssize_t _length
```

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,82 @@ cdef class UCXWorker(UCXObject):
...
+    cdef worker
+
+    def __init__(self, UCXWorker worker):
+        cdef ucs_status_t status
+        cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
+
+        status = ucp_worker_get_address(ucp_worker, &self._handle, &self._length)
+        assert_ucs_status(status)
```
jakirkham, line 426: Would suggest just using a local variable that matches the `size_t` expected by UCX and then convert it to `Py_ssize_t` afterwards.

```suggestion
        cdef size_t length

        status = ucp_worker_get_address(ucp_worker, &self._handle, &length)
        assert_ucs_status(status)
        self._length = length
```

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,82 @@ cdef class UCXWorker(UCXObject):
...
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        if (flags & PyBUF_WRITABLE) == PyBUF_WRITABLE:
+            raise BufferError("Requested writable view on readonly data")
+        buffer.buf = <void*>self._handle
+        buffer.obj = self
+        buffer.len = self._length
+        buffer.readonly = True
+        buffer.itemsize = self._length
+        if (flags & PyBUF_FORMAT) == PyBUF_FORMAT:
+            buffer.format = b"p"
```
jakirkham, line 452: Should be capitalized if we want `void *`. The lower cased one is `char[]`.

```suggestion
            buffer.format = b"P"
```

https://docs.python.org/3/library/struct.html#format-characters

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,82 @@ cdef class UCXWorker(UCXObject):
...
+        return int(self._length)
+
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        if (flags & PyBUF_WRITABLE) == PyBUF_WRITABLE:
+            raise BufferError("Requested writable view on readonly data")
+        buffer.buf = <void*>self._handle
+        buffer.obj = self
+        buffer.len = self._length
+        buffer.readonly = True
+        buffer.itemsize = self._length
```
jakirkham, line 450: Should be the size of type contained as opposed to the number of bytes used.

```suggestion
        buffer.itemsize = sizeof(void*)
```

https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.itemsize

ucp/_libs/ucx_api.pyx
jakirkham, line 457: Can drop the casting after the other changes are made.

```suggestion
            buffer.shape = &self._length
```

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,82 @@ cdef class UCXWorker(UCXObject):
...
+    @property
+    def length(self):
+        return int(self._length)
+
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        if (flags & PyBUF_WRITABLE) == PyBUF_WRITABLE:
+            raise BufferError("Requested writable view on readonly data")
+        buffer.buf = <void*>self._handle
+        buffer.obj = self
+        buffer.len = self._length
```
jakirkham, line 448: Includes the `itemsize` as well.

```suggestion
        buffer.len = self._length * sizeof(void *)
```

https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.len

jakirkham (commented):
Thanks for working on this Matt! 😄

Had a couple minor points about the Python Buffer protocol. Mostly it makes things a bit easier to hold a `Py_ssize_t` typed `_length` as opposed to a `size_t` due to it being signed. Some other minor comments.

ucp/_libs/ucx_api.pyx
MattBBaker: That makes sense to me.

ucp/_libs/ucx_api.pyx
MattBBaker: That description sounds to me like an array of (void*)s. The ucp_address_t isn't an array of (void*)s, it's packed bytes. So it seems to me that the char[] is the right type.

jakirkham: Ah ok. Sorry if I misunderstood. In that case maybe we should be doing `sizeof(ucp_address_t)` in a few places below and leave this as-is.

MattBBaker: The trouble is that ucp_address_t is an incomplete type, so using `sizeof(ucp_address_t)` won't work. 
https://github.com/openucx/ucx/blob/8b6641b02ef5d217323f719d54dfb2e9142dad59/src/ucp/api/ucp_def.h#L110

jakirkham: Hmm...so how do we know how much memory this takes up?

MattBBaker: The `ucp_worker_get_address()` function has a length output argument that tells you the length. So you get a byte string and a length and that gets stored in the UCXAddress object.

ucp/_libs/ucx_api.pyx
MattBBaker: This is where you get length, so here length will still be size_t but will be casted into Py_ssize_t when assigned to the object.

ucp/_libs/ucx_api.pyx
jakirkham: Ah ok. If it is just bytes, then maybe we should just type this as `"B"` (IOW `uint8_t`) and the `itemsize` is just `1`.

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,84 @@ cdef class UCXWorker(UCXObject):
...
+        buffer.readonly = True
+        buffer.itemsize = self._length
```
jakirkham, line 452: Since `sizeof(uint8_t)` is just `1`. Though we could use `sizeof` if that's clearer

```suggestion
        buffer.itemsize = 1
```

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,84 @@ cdef class UCXWorker(UCXObject):
...
+        if (flags & PyBUF_FORMAT) == PyBUF_FORMAT:
+            buffer.format = b"p"
```
jakirkham, line 454: Matches `uint8_t` type

```suggestion
            buffer.format = b"B"
```

jakirkham (commented):
Since it sounds like `uint8_t` works for us here, added a few minor comments to update this accordingly.

ucp/_libs/ucx_api.pyx
jakirkham: Ok so if that works, have added a couple suggestions below based on that change ( https://github.com/rapidsai/ucx-py/pull/594#pullrequestreview-500851529 ).

jakirkham:
Outside of the comments above, I think this LGTM. Thanks for working on this Matt! 😄

@madsbk, would you also be able to take a look as well?

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,84 @@ cdef class UCXWorker(UCXObject):
...
+        self._length = <Py_ssize_t>length
+        self.add_handle_finalizer(
+            _ucx_address_finalizer,
+            int(<uintptr_t>self._handle),
+            worker
+        )
+        worker.add_child(self)
+
+    @property
+    def address(self):
```
madsbk, line 438:
```suggestion
    def address(self):
        assert self.initialized
```

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,84 @@ cdef class UCXWorker(UCXObject):
...
+        return <uintptr_t>self._handle
+
+    @property
+    def length(self):
```
madsbk, line 442:
```suggestion
    def length(self):
        assert self.initialized
```

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,84 @@ cdef class UCXWorker(UCXObject):
...
+        return int(self._length)
+
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
```
madsbk, line 445:
```suggestion
    def __getbuffer__(self, Py_buffer *buffer, int flags):
        assert self.initialized
```

tests/test_rma.py
```diff
@@ -0,0 +1,29 @@
...
+    ucp.reset()
+
+
+@pytest.mark.parametrize("blocking_progress_mode", [True, False])
+def test_worker_address(blocking_progress_mode):
+    ucp.init(blocking_progress_mode=blocking_progress_mode)
+
+    addr = ucp.get_worker_address()
+    assert addr is not None
+    ucp.reset()
```
madsbk, line 29: A test that communicate `UCXAddress` between two endpoints would be good. It will show that the buffer protocol works and that different processes can read the address.

madsbk (changes requested):
Nice work @MattBBaker, it looks very good! 
I have some minor suggestions and it would also be great with a communication test of `UCXAddress`.

tests/test_rma.py
pentschev: +1 to this, the easiest is probably just to copy https://github.com/rapidsai/ucx-py/blob/5208b4538a86e2d1e22b596a5050943aeae70bcd/tests/test_send_recv.py#L56-L71 and modify it to pass the address.

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,84 @@ cdef class UCXWorker(UCXObject):
...
+            status, 0, cb_func, cb_args, cb_kwargs, 'flush', self._inflight_msgs
+        )
+
+    def get_address(self):
+        return UCXAddress(self)
+
+
+def _ucx_address_finalizer(uintptr_t handle_as_int, UCXWorker worker):
+    cdef ucp_address_t *address = <ucp_address_t *>handle_as_int
+    cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
```
pentschev, line 411: Wouldn't it make sense to just pass `worker_handle_as_int` here to make both arguments type-uniform with `uintptr_t` and remove the need for a double cast here?

pentschev (approved):
I added a couple minor suggestions, but LGTM to me as well, thanks @MattBBaker !

ucp/_libs/ucx_api.pyx
jakirkham: Good point. Yeah there is a faster way to access this.

```suggestion
    cdef ucp_worker_h ucp_worker = worker._handle
```

MattBBaker: Got ya, will add test described above and change up how the handle is handled.

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,87 @@ cdef class UCXWorker(UCXObject):
         assert_ucs_status(status)
         return UCXEndpoint(self, <uintptr_t>ucp_ep)
 
+    cpdef ucs_status_t fence(self) except *:
+        cdef ucs_status_t status = ucp_worker_fence(self._handle)
+        assert_ucs_status(status)
+        return status
+
+    def flush(self, cb_func, cb_args, cb_kwargs=dict()):
+        cdef ucs_status_ptr_t req
```
jakirkham, line 397: Would suggest using `None` as mutable default arguments (like `dict()`) sometimes have weird behavior in Python, which `None` avoids. Also checking if something is `None` in Cython is quite fast as it is just a pointer comparison. Additionally suggest making `cb_args` optional as we have done elsewhere.

```suggestion
    def flush(self, cb_func, cb_args=None, cb_kwargs=None):
        if cb_args is None:
            cb_args = ()
        if cb_kwargs is None:
            cb_kwargs = {}
        cdef ucs_status_ptr_t req
```

ucp/_libs/ucx_api.pyx
```diff
@@ -456,6 +539,15 @@ cdef class UCXEndpoint(UCXObject):
         assert self.initialized
         return int(<uintptr_t>self._handle)
 
+    def flush(self, cb_func, cb_args, cb_kwargs=dict()):
```
jakirkham, line 542: Would suggest similar changes here as above. Namely using `None` and making `cb_args` also optional.

jakirkham:
Sorry added a couple more suggestions above. Had missed these before. Though these are hopefully minor changes to make.

ucp/_libs/ucx_api.pyx
```diff
@@ -386,6 +388,91 @@ cdef class UCXWorker(UCXObject):
...
+
+cdef class UCXAddress(UCXObject):
+    """Python representation of ucp_address_t"""
+    cdef ucp_address_t *_handle
+    cdef Py_ssize_t _length
+    cdef worker
+
+    def __init__(self, UCXWorker worker):
+        cdef ucs_status_t status
+        cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker.handle
```
jakirkham, line 427: Looks like this got fixed below, but this one is still remaining.

```suggestion
        cdef ucp_worker_h ucp_worker = <ucp_worker_h><uintptr_t>worker._handle
```

ucp/_libs/ucx_api.pyx
jakirkham: Thought this was resolved, but it appears the diff was just outdated. Mentioned in a new review comment below ( https://github.com/rapidsai/ucx-py/pull/594#pullrequestreview-502316610 ).

ucp/_libs/ucx_api.pyx
MattBBaker: Whoops. That one slipped through the cracks.

jakirkham (approved):

ucp/_libs/ucx_api.pyx
jakirkham: No worries. Thanks!

jakirkham:
Thanks Matt! LGTM 😄

@madsbk, would you be able to take another look?

madsbk:
Yes, will take a look tomorrow.

madsbk (approved):
LGTM, nice work!

PR #595, opened 2020-09-08
Implement `get_itemsize`

jakirkham:
This provides the function `get_itemsize` to fast path identification of the `itemsize` some common `typestr`s. Anything else that cannot be easily identified is passed off to NumPy for determination.

Before:

```python
In [1]: import cupy
   ...: import numpy
   ...: 
   ...: from ucp._libs.utils import get_buffer_nbytes

In [2]: s = (10, 11, 12, 13)

In [3]: %timeit get_buffer_nbytes(a_g, check_min_size=None, cuda_support=True)
1.28 µs ± 9.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

After:

```python
In [1]: import cupy
   ...: import numpy
   ...: 
   ...: from ucp._libs.utils import get_buffer_nbytes

In [2]: s = (10, 11, 12, 13)

In [3]: %timeit get_buffer_nbytes(a_g, check_min_size=None, cuda_support=True)
1 µs ± 6.44 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

### DIFF
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -12,10 +12,68 @@ from cpython.memoryview cimport (
 from cython cimport boundscheck, wraparound
 from libc.stdint cimport uintptr_t
 
+try:
+    from numpy import dtype as numpy_dtype
+except ImportError:
+    numpy_dtype = None
+
 
 cdef struct iface_data_t:
     uintptr_t ptr
     bint readonly
+
+
+cdef dict itemsize_mapping = {
+    "|b1": 1,
+    "|i1": 1,
+    "|u1": 1,
+    "<i2": 2,
+    ">i2": 2,
+    "<u2": 2,
+    ">u2": 2,
+    "<i4": 4,
+    ">i4": 4,
+    "<u4": 4,
+    ">u4": 4,
+    "<i8": 8,
+    ">i8": 8,
+    "<u8": 8,
+    ">u8": 8,
+    "<f2": 2,
+    ">f2": 2,
+    "<f4": 4,
+    ">f4": 4,
+    "<f8": 8,
+    ">f8": 8,
+    "<f16": 16,
+    ">f16": 16,
+    "<c8": 8,
+    ">c8": 8,
+    "<c16": 16,
+    ">c16": 16,
+    "<c32": 32,
+    ">c32": 32,
+}
+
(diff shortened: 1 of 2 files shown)


madsbk (approved):
LGTM

ucp/_libs/utils.pyx
```diff
@@ -12,12 +12,70 @@ from cpython.memoryview cimport (
...
+}
+
+cpdef Py_ssize_t get_itemsize(str fmt) except *:
+    """
+    Get the itemsize of the format provided.
+    """
+    if fmt is None:
+        raise ValueError("Expected `str`, but got `None`")
+    elif fmt == "":
+        raise ValueError("Got unexpected empty `str`")
```
pentschev, line 65: Why do we need to special case the empty string, wouldn't that be covered by the `ValueError` exception below?

pentschev:
I took the liberty to fix the conflict @jakirkham , hope you don't mind.

ucp/_libs/utils.pyx
jakirkham: I'm not sure that error message would fit. An empty string would still raise even if NumPy was installed. So it's not some more complicated format (like `<M8[D]`) that we don't include in the `dict` above.

jakirkham:
Thanks Peter! Pushed a few more minor cleanup fixes in the last commit. Should fix CI.

ucp/_libs/utils.pyx
pentschev: Fair enough, thanks for clarifying.

pentschev (approved):
LGTM, thanks @jakirkham !

jakirkham:
Pulled out CI fixes in PR ( https://github.com/rapidsai/ucx-py/pull/598 ).

jakirkham:
Just to add one other thing, simply calling `getattr` takes a fair bit of time (slightly faster in Cython itself). So this optimization puts us a hair above just calling `getattr`. Meaning there probably isn't much more juice to squeeze here.

```python
In [1]: import cupy
   ...: import numpy

In [2]: s = (10, 11, 12, 13)

In [3]: %timeit getattr(a_g, "__cuda_array_interface__", None)
820 ns ± 6.04 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

PR #596, opened 2020-09-09
Simplify `get_buffer_nbytes` `min_size` check

jakirkham:
Makes a few adjustments to `get_buffer_nbytes` signature. Namely replaces `check_min_size` with `min_size`, which is now typed as `Py_ssize_t`. If `min_size` is not positive definite, this is the same as skipping the checking. Also provides some reasonable default values to `min_size` and `cuda_support`.

### DIFF
--- a/tests/test_libs_utils.py
+++ b/tests/test_libs_utils.py
@@ -40,18 +40,16 @@ def test_get_buffer_data_builtins(buffer):
 @pytest.mark.parametrize("buffer", builtin_buffers)
 def test_get_buffer_nbytes_builtins(buffer):
     nbytes = memoryview(buffer).nbytes
-    result = get_buffer_nbytes(buffer, check_min_size=None, cuda_support=True)
+    result = get_buffer_nbytes(buffer, cuda_support=True)
     assert result == nbytes
 
     with pytest.raises(ValueError):
-        get_buffer_nbytes(
-            memoryview(buffer)[::2], check_min_size=None, cuda_support=True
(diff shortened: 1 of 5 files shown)


jakirkham:
FWIW with this makes a small, but notable improvement in how long this check takes to run. Included a benchmark below just for the NumPy case. Though the CuPy case will show a similar improvement (as the check behaves the same independent of being on host or device).

Before:

```python
In [1]: import numpy
   ...: 
   ...: from ucp._libs.utils import get_buffer_nbytes

In [2]: s = (10, 11, 12, 13)

In [3]: %timeit get_buffer_nbytes(a_c, check_min_size=a_c.nbytes, cuda_support=True)
709 ns ± 1.64 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

After:

```python
In [1]: import numpy
   ...: 
   ...: from ucp._libs.utils import get_buffer_nbytes

In [2]: s = (10, 11, 12, 13)

In [3]: %timeit get_buffer_nbytes(a_c, min_size=a_c.nbytes, cuda_support=True)
679 ns ± 7.38 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

madsbk (approved):
LGTM, thanks @jakirkham.

pentschev (approved):
LGTM too, thanks @jakirkham !

jakirkham:
Thanks all! 😄

PR #599, opened 2020-09-10
Make sure type `str`'s in `dict` are `intern`ed

jakirkham:
Lookups in Python `dict`s with `str` keys can be sped up if the keys are `intern`ed and the key accessed is `intern`ed. We use this to `intern` all of the keys of `itemsize_mapping` to speedup lookup times with this `dict`. This way if we encounter a type `str`, which is `intern`ed Python can speedup lookup time a bit.

https://docs.python.org/3/library/sys.html#sys.intern

https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#built-in-functions


jakirkham:
This should give a rough idea of the difference seen between normal and `intern`ed `str`s.

```python
In [1]: import sys
   ...: 
   ...: from ucp._libs.utils import get_itemsize

In [2]: ts = "<f8"
   ...: ts_interned = sys.intern(ts)

In [3]: %timeit get_itemsize(ts)
55.8 ns ± 0.0467 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit get_itemsize(ts_interned)
46.6 ns ± 0.029 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

pentschev (approved):
Thanks @jakirkham , that's another nice improvement, NumPy also uses string interning a lot for the same reason.

PR #600, opened 2020-09-10
Handle shape and strides in C

jakirkham:
Coerces the `tuple`s of `shape` and `strides` into dynamically allocated C arrays, which are then worked with directly. Thus avoiding the need to coerce items from `tuple`s in possibly multiple loops. Also refactors out C functions for computing `nbytes` and whether the array is C-contiguous. These functions are then used for both the CPU and GPU cases.


jakirkham:
This results in a small yet notable performance improvement for both the CPU and GPU cases in `get_buffer_nbytes` as shown by the benchmark below.

<details>
<summary>Before:</summary>

```python
In [1]: import cupy
   ...: import numpy
   ...: 
   ...: from ucp._libs.utils import (
   ...:     get_buffer_data,
   ...:     get_buffer_nbytes,
   ...: )

   ...:     try:
   ...:     except Exception:
   ...:         pass
   ...: 

In [3]: s = (10, 11, 12, 13)

In [4]: %timeit get_buffer_nbytes(a_c, cuda_support=True)
623 ns ± 2.22 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit get_buffer_nbytes(a_g, cuda_support=True)
1.07 µs ± 6.59 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit run(get_buffer_nbytes, a_c.T, cuda_support=True)
1.63 µs ± 7.26 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [7]: %timeit run(get_buffer_nbytes, a_g.T, cuda_support=True)
1.96 µs ± 7.53 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

</details>

<details>
<summary>After:</summary>

```python
In [1]: import cupy
   ...: import numpy
   ...: 
   ...: from ucp._libs.utils import (
   ...:     get_buffer_data,
   ...:     get_buffer_nbytes,
   ...: )

   ...:     try:
   ...:     except Exception:
   ...:         pass
   ...: 

In [3]: s = (10, 11, 12, 13)

In [4]: %timeit get_buffer_nbytes(a_c, cuda_support=True)
624 ns ± 1.65 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit get_buffer_nbytes(a_g, cuda_support=True)
1.02 µs ± 1.55 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit run(get_buffer_nbytes, a_c.T, cuda_support=True)
1.27 µs ± 13.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [7]: %timeit run(get_buffer_nbytes, a_g.T, cuda_support=True)
1.79 µs ± 1.96 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

</details>

pentschev:
Does this provide some performance improvement or what is the reason behind this change? The code is considerably more complex than the original, it would be good to know how much we're gaining with this change.

pentschev:
> Does this provide some performance improvement or what is the reason behind this change? The code is considerably more complex than the original, it would be good to know how much we're gaining with this change.

Sorry, I missed https://github.com/rapidsai/ucx-py/pull/600#issuecomment-689969992 . Yeah, I see we have some improvement. The changes are fine by me overall. Thanks @jakirkham .

pentschev (approved):

jakirkham:
Numbers are in the previous comment.

Would recommend looking at the commit level diff. The overall diff GitHub shows unfortunately muddles things unnecessarily.

PR #601, opened 2020-09-10
Type variables in `assert_ucs_status`

jakirkham:
Types variables in `assert_ucs_status` so that Cython can use more specific Python C API operations related to those objects (namely `str`s). Also use an [f-string]( https://docs.python.org/3.8/reference/lexical_analysis.html#formatted-string-literals ) to clarify message formatting.


jakirkham:
rerun tests

madsbk (approved):
Looks good

jakirkham:
Thanks Mads! 😄

PR #602, opened 2020-09-15
Add `Array` class to wrap array-likes

jakirkham:
As [noted previously]( https://github.com/rapidsai/ucx-py/pull/595#issuecomment-689873497 ), just calling `getattr` to get `__cuda_array_interface__` (or not) takes a good chunk of time and dominates the time spent in `get_buffer_data` and `get_buffer_nbytes`. As such we end up paying for this `getattr` call every time we call those utility functions, which can be quite expensive.

To avoid repeated calls to `getattr`, this creates an `Array` class that calls `getattr` once and then extracts information from `__cuda_array_interface__` (or uses a `memoryview`). The extracted information is stored into C variables of this class (similar in nature to how `get_buffer_data` and `get_buffer_nbytes` operated internally). As a result accessing these values is later very efficient both in Python and in Cython. This should cutdown on the overhead of fetching the underlying pointer or size when providing this info to UCX.


jakirkham:
Currently we make calls to `get_buffer_data` and `get_buffer_nbytes` for everything we send over the wire. So we could benchmark running both functions together and get this result.

<details>

```python
In [1]: import cupy
   ...: import numpy

In [2]: from ucp._libs.utils import (
   ...:     get_buffer_data,
   ...:     get_buffer_nbytes,
   ...: )

In [3]: s = (10, 11, 12, 13)

In [4]: def func(obj):
   ...:     ptr = get_buffer_data(obj)
   ...:     nbytes = get_buffer_nbytes(obj, cuda_support=True)
   ...:     return (ptr, nbytes)
   ...: 

In [5]: %timeit func(a_c)
1.26 µs ± 4.15 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit func(a_g)
1.95 µs ± 12.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

</details>

These functions actually do a few extra things (like check whether the data is C-contiguous). So we could benchmark equivalent behavior using `Array` like so.

<details>

```python
In [1]: import cupy
   ...: import numpy

In [2]: from ucp._libs.arr import Array

In [3]: s = (10, 11, 12, 13)

In [4]: def func(obj):
   ...:     ary = Array(obj)
   ...:     if not ary.c_contiguous:
   ...:         raise ValueError("Array must be C-contiguous")
   ...:     return (ary.ptr, ary.nbytes)
   ...: 

In [5]: %timeit func(a_c)
729 ns ± 2.08 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit func(a_g)
1.28 µs ± 6.23 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

</details>

Though it's worth pointing out that things like accessing attributes of `Array` (like `ptr` or `readonly`) are equivalent to accessing a `struct` variable in C. So this is quite a bit faster in Cython code than the above benchmark would suggest.

ucp/_libs/arr.pyx
```diff
@@ -0,0 +1,287 @@
...
+    intern(">c8"): 8,
+    intern("<c16"): 16,
+    intern(">c16"): 16,
+    intern("<c32"): 32,
+    intern(">c32"): 32,
+}
+
+
+@auto_pickle(False)
+cdef class Array:
```
madsbk, line 59: Would be good with some doc string here

ucp/_libs/arr.pyx
```diff
@@ -0,0 +1,287 @@
...
+            shape = iface["shape"]
+            strides = iface.get("strides")
+            self.ndim = len(shape)
+            if self.ndim > 0:
+                if strides is not None:
+                    if len(strides) != self.ndim:
+                        raise ValueError(
+                            "The length of shape and strides must be equal"
+                        )
+                    self.shape_p = <Py_ssize_t*>PyMem_Malloc(
```
madsbk, line 106: I know we are doing this to reduce overhead but maybe this is going a bit too far?
Could we use a regular python tuples for `self.shape` and `self.strides`?

madsbk (changes requested):
Overall I think it is very good @jakirkham, I really like the `Array` class!
But I think maybe you have taken the optimization a bit too far :)
For instance, I don't think the extra complixity of manually managing the allocation of shape and strides in `Array` is worth it.

ucp/_libs/arr.pyx
jakirkham: This just mirrors the same change that happened in PR ( https://github.com/rapidsai/ucx-py/pull/600 ).

FWIW this class still provides `shape` and `strides` as properties returning `tuple`s. So these remain accessible that way.

madsbk: Sorry I missed #600 but I still think that this is taking the low-level optimization too far. 
I mean ~500 ns should be negligible in this context or am I missing something?

jakirkham: If we are ok with that overhead, what if went with something like Python's builtin `array`? That remains friendly to use from the Python side, but would also allow efficient typed access from the C-side.

madsbk: Great idea, best of both worlds :)

jakirkham: Cool, have pushed changes along those lines.

ucp/_libs/arr.pyx
jakirkham: I've added something simple as a starting point. Though if there's more things you would like to see here, happy to fill this out further 🙂

ucp/_libs/arr.pyx
jakirkham: Just to note, this appears to add `~200ns` to `Array` creation time both for host and device cases. So this seems to be within what we consider negligible here.

madsbk (approved):
Nice work @jakirkham

jakirkham:
Thanks Mads! 😀

Issue #604, opened 2020-09-17
Usage of `nbytes` limits


jakirkham:
In both `send` and `recv`, we have the option to specify `nbytes` if we only want to send or receive part of `buffer`. Am curious how often this is used in practice. Asking in part because NumPy and CuPy arrays can typically be sliced to provide a view onto a smaller part of the array when needed. Would it be reasonable to drop these options? Or are there reasons we might still need these?

https://github.com/rapidsai/ucx-py/blob/c60f43b3cd9c304053e50a0e4d6d06500e450b1d/ucp/core.py#L554

https://github.com/rapidsai/ucx-py/blob/c60f43b3cd9c304053e50a0e4d6d06500e450b1d/ucp/core.py#L603

jakirkham:
cc @madsbk

pentschev:
I think they're useful if the application knows the size beforehand, that would save a little bit of time on UCX trying to figure that out. In practice, I don't think we're using it anywhere, but it's hard to say given the largest use case nowadays is still Dask.

jakirkham:
Ah but it isn't for the size of `buffer`. It is an upper bound for the allowed size of `buffer`. If `buffer` is above that, it errors. So what I'm asking is why do we need that check here (especially if users can do that themselves)?

madsbk:
Removing the option is fine with me. I added them initially just-because-I-could, I don't think anyone are using them.

> I think they're useful if the application knows the size beforehand, that would save a little bit of time on UCX trying to figure that out.

In principle it would make sense to have a feature like this but with @jakirkham's recent work on reducing the overhead of retrieving buffer information I don't think it is worth the effort.

jakirkham:
Thanks Mads :) Ok will add a PR removing this.

PR #605, opened 2020-09-19
Update Array checks

jakirkham:
Fixes https://github.com/rapidsai/ucx-py/issues/604

This moves the CUDA and contiguous checks of `Array`s into Cython. Also allows C or F contiguous arrays. Drops the `nbytes` check in Python.

### DIFF
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -786,8 +786,8 @@ def tag_send_nb(
         cb_kwargs = {}
     if name is None:
         name = "tag_send_nb"
-    if not buffer._c_contiguous():
-        raise ValueError("Array must be C-contiguous")
+    if not buffer._contiguous():
+        raise ValueError("Array must be C or F contiguous")
     cdef ucp_send_callback_t _send_cb = <ucp_send_callback_t>_send_callback
     cdef ucs_status_ptr_t status = ucp_tag_send_nb(
         ep._handle,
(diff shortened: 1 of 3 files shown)


jakirkham:
Would like to move the `cuda_support` check into Cython as well. Not sure if anyone has thoughts on how best to do that.

jakirkham:
Looks like [CI ran into a download error]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/ucx-py/job/prb/job/ucx-py-gpu-build/2512/display/redirect ).

rerun tests

jakirkham:
Looks like [one CI job had issues downloading packages]( https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/2571/pipeline )

rerun tests

jakirkham:
Looks like [another CI job had issues downloading packages]( https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci-a100%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/35/pipeline )

rerun tests

jakirkham:
Had a question about `tag_recv_nb` that came up when working on this ( https://github.com/rapidsai/ucx-py/issues/619 ).

ucp/_libs/ucx_api.pyx
```diff
@@ -786,6 +791,17 @@ def tag_send_nb(
         cb_kwargs = {}
     if name is None:
         name = "tag_send_nb"
+    if buffer.cuda and not ep.worker._context.cuda_support:
```
quasiben, line 794: I think UCX will throw an error first if you don't define a CUDA transport while trying to pass CUDA data.  @pentschev is that right ?

jakirkham: Well we had a test for this before. It was just in Python. Now we have moved it to C. I suppose we could drop the check, but I’m guessing it was there for a reason. So kept it in. Plus this check should be pretty fast

quasiben (approved):
Thanks @jakirkham

ucp/_libs/ucx_api.pyx
pentschev: For completeness: yes, UCX will fail, but we can't recover from C errors at the moment, so keeping them in UCX-Py is useful, regardless in what layer we do so.

PR #608, opened 2020-09-21
Specify `cdef` return types

jakirkham:
Fill in the return types of a few `cdef` functions that were missing before.

### DIFF
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -50,7 +50,7 @@ cdef int _ucx_py_request_counter = 0
 logger = logging.getLogger("ucx")
 
 
-cdef assert_ucs_status(ucs_status_t status, str msg_context=None):
+cdef void assert_ucs_status(ucs_status_t status, str msg_context=None) except *:
     cdef str msg, ucs_status
     if status != UCS_OK:
         ucs_status = ucs_status_string(status).decode("utf-8")
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -624,8 +624,8 @@ cdef UCXRequest _handle_status(
     cb_func,
     cb_args,
     cb_kwargs,
-    name,
-    inflight_msgs
+    unicode name,
+    set inflight_msgs
 ):
     if UCS_PTR_STATUS(status) == UCS_OK:
         return
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -618,7 +618,7 @@ cdef class UCXRequest:
             )
 
 
-cdef _handle_status(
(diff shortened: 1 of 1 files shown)


jakirkham:
Ran into [a download error in one CI job]( https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/2508/pipeline ).

rerun tests

jakirkham:
Still having [download issues on CI]( https://github.com/rapidsai/ucx-py/pull/608 )


rerun tests

jakirkham:
Clearing [stale builder status]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/ucx-py/job/prb/job/ucx-py-prb/964/ ).

rerun tests

ucp/_libs/ucx_api.pyx
```diff
@@ -618,14 +618,14 @@ cdef class UCXRequest:
...
+cdef UCXRequest _handle_status(
     ucs_status_ptr_t status,
     int64_t expected_receive,
     cb_func,
     cb_args,
     cb_kwargs,
-    name,
-    inflight_msgs
+    unicode name,
+    set inflight_msgs
```
pentschev, line 628: I see that making this a `set` matches other entries in the file. I would though expect something named as inflight messages to be a `list` instead. Wouldn't it make more sense? Isn't there a chance that you can have the same message more than once from multiple calls?

jakirkham: The definition comes from [here]( https://github.com/rapidsai/ucx-py/blob/21d6d5cca9fb6b826a87d46cb231cc5a9c43ff5f/ucp/_libs/ucx_api.pyx#L293 ). All we are doing in this PR is making sure it is typed correctly so Cython can use [more efficient C APIs with the `set`]( https://docs.python.org/3/c-api/set.html ). Maybe it would be better to ask this question in a new issue?

pentschev: Yeah, I thought that would be overstepping this PR's context, just making sure. Thanks for clarifying @jakirkham .

pentschev (approved):
Changes look good to me, thanks @jakirkham .

ucp/_libs/ucx_api.pyx
madsbk: >  Wouldn't it make more sense? Isn't there a chance that you can have the same message more than once from multiple calls?

Not as far as I know, `inflight_msgs` contains `UCXRequest`, which [hashes to the UID of the underlying `ucx_py_request`](https://github.com/rapidsai/ucx-py/blob/21d6d5cca9fb6b826a87d46cb231cc5a9c43ff5f/ucp/_libs/ucx_api.pyx#L602).

PR #609, opened 2020-09-21
Cleanup `ucx_config_to_dict`

jakirkham:
* Drop unneeded parentheses from `if` statement
* Type `line`, which was not typed before
* Type `k` and `v`, which were not typed before
* Handle an error from `fflush` if raised
* Simplify slicing of `k`
* Handle an error from `fclose` if raised


pentschev (approved):
LGTM, thanks @jakirkham .

PR #610, opened 2020-09-21
Cleanup `UCXEndpoint.info`

jakirkham:
* Drop unneeded parentheses from `if` statement
* Handle an error from `fflush` if raised
* Handle an error from `fclose` if raised


pentschev (approved):
LGTM, thanks @jakirkham .

PR #611, opened 2020-09-22
Release `config` on error

jakirkham:
As we are responsible for cleaning up the `config` object when we are done with it, make sure that we clean it up ourselves when an exception occurs as no one else will have the opportunity to do so then.


pentschev (approved):
LGTM, thanks @jakirkham .

PR #612, opened 2020-09-22
Type `send`, `recv` arguments and `port`

jakirkham:
* Type `cb_args` and `cb_kwargs` as `tuple` and `dict` respectively
* Type `name` as `str`
* Fix `port` typing to `uint16_t` as UCX and hwloc type it
* Use `None` for default arguments


madsbk (approved):
LGTM

PR #617, opened 2020-09-28
Type `cuda_device_index` as `unsigned int`

jakirkham:
Make sure `cuda_device_index` matches the expected type of `unsigned int`.

### DIFF
--- a/ucp/_libs/topological_distance.pxd
+++ b/ucp/_libs/topological_distance.pxd
@@ -15,5 +15,6 @@ cdef class TopologicalDistance:
                                            int device,
                                            str device_type=*)
 
-    cpdef get_cuda_distances_from_device_index(self, cuda_device_index,
+    cpdef get_cuda_distances_from_device_index(self,
+                                               unsigned int cuda_device_index,
                                                str device_type=*)
--- a/ucp/_libs/topological_distance.pyx
+++ b/ucp/_libs/topological_distance.pyx
@@ -103,7 +103,8 @@ cdef class TopologicalDistance:
 
         return ret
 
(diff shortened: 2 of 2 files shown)


madsbk (commented):
To guard against negative values, I suggest that you make it `int cuda_device_index` in the API and then check for negative values before calling `pynvml.nvmlDeviceGetHandleByIndex()`

jakirkham:
[`pynvml.nvmlDeviceGetHandleByIndex`]( https://github.com/gpuopenanalytics/pynvml/blob/4e34801ffa22b8ac654440f9e4e15b2aec1dc3a4/pynvml/nvml.py#L917 ) just calls [`nvmlDeviceGetHandleByIndex_v2`]( https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g4cc7ff5253d53cc97b1afb606d614888 ), which takes an `unsigned int` for `index`. So it already expects an `unsigned int`. This just aligns the code to that expectation.

Cython will check the argument to see that it is a Python integer and if so that it is within bounds. If not, it throws an `OverflowError`. This can be verified with a simple example.

```cython
def func(unsigned int i):
    pass

func(-1)  # <--- raises `OverflowError`
```

madsbk (approved):
> Cython will check the argument to see that it is a Python integer and if so that it is within bounds. If not, it throws an OverflowError.

Nice! I thought it used C/C++ semantic in this regard but I see that isn't the case. Thanks @jakirkham  :)

jakirkham:
Yeah this one of the reasons typing variables in Cython is so handy. It helps throw errors for all the mismatched values :)

Issue #616, opened 2020-09-23
SUMMIT IB Configuration


quasiben:
On summit, the nodes have the following configuration:

```
> jsrun -n 1 -c 42 -g 6 -a 1 --smpiargs='off' nvidia-smi topo -m
        GPU0    GPU1    GPU2    GPU3    GPU4    GPU5    mlx5_0  mlx5_1  mlx5_2mlx5_3   CPU Affinity
GPU0     X      NV2     NV2     SYS     SYS     SYS     NODE    NODE    SYS   SYS      0-83
GPU1    NV2      X      NV2     SYS     SYS     SYS     NODE    NODE    SYS   SYS      0-83
GPU2    NV2     NV2      X      SYS     SYS     SYS     NODE    NODE    SYS   SYS      0-83
GPU3    SYS     SYS     SYS      X      NV2     NV2     SYS     SYS     NODE  NODE     88-171
GPU4    SYS     SYS     SYS     NV2      X      NV2     SYS     SYS     NODE  NODE     88-171
GPU5    SYS     SYS     SYS     NV2     NV2      X      SYS     SYS     NODE  NODE     88-171
mlx5_0  NODE    NODE    NODE    SYS     SYS     SYS      X      PIX     SYS   SYS
mlx5_1  NODE    NODE    NODE    SYS     SYS     SYS     PIX      X      SYS   SYS
mlx5_2  SYS     SYS     SYS     NODE    NODE    NODE    SYS     SYS      X    PIX
mlx5_3  SYS     SYS     SYS     NODE    NODE    NODE    SYS     SYS     PIX    X 

Legend:

  X    = Self
  SYS  = Connection traversing PCIe as well as the SMP interconnect between NUMA nodes (e.g., QPI/UPI)
  NODE = Connection traversing PCIe as well as the interconnect between PCIe Host Bridges within a NUMA node
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe switches (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing a single PCIe switch
  NV#  = Connection traversing a bonded set of # NVLinks
```

Each node has 6 GPUs and 4 MLNX Devices.  I'm not sure what the optimal pairing of GPU and MLNX device should be.  Normally, I would rely on `hwloc` to figure this out however, on Summit i get errors like the following (when using `--net-devices='auto' with dask-cuda`):
[...]

PR #626, opened 2020-09-30
add warning for when UCX_MEM_MMAP_HOOK_MODE is set

quasiben:
Related to https://github.com/rapidsai/ucx-py/issues/616

cc @pentschev

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -39,6 +39,13 @@ if not os.environ.get("UCX_TCP_TX_SEG_SIZE", False):
 if not os.environ.get("UCX_TCP_RX_SEG_SIZE", False):
     os.environ["UCX_TCP_RX_SEG_SIZE"] = "8M"
 
+if os.environ.get("UCX_MEM_MMAP_HOOK_MODE", False):
+    msg = (
+        "Setting `UCX_MEM_MMAP_HOOK_MODE` disable IB registration cache "
+        "which disable GPURDMAincluding GPU direct. \nPlease unset "
+        "UCX_MEM_MMAP_HOOK_MODE"
+    )
+    logger.warning(msg)
 
 # After handling of environment variable logging, add formatting to the logger
 logger = get_ucxpy_logger()
--- a/tests/test_mmap_warn.py
+++ b/tests/test_mmap_warn.py
@@ -1,0 +1,17 @@
+import os
+
+
+def test_mem_mmap_hook_warn(caplog):
+    """
+    Test warning for UCX_MEM_MMAP_HOOK_MODE
+    """
+    import logging
+
+    os.environ["UCX_MEM_MMAP_HOOK_MODE"] = "none"
+
+    # ucp.init will only print INFO LINES
+    with caplog.at_level(logging.INFO):
+        import ucp
+
+        ucp.init()
+    assert any(["UCX_MEM_MMAP_HOOK_MODE" in rec.message for rec in caplog.records])


ucp/__init__.py
```diff
@@ -39,6 +39,13 @@
 if not os.environ.get("UCX_TCP_RX_SEG_SIZE", False):
     os.environ["UCX_TCP_RX_SEG_SIZE"] = "8M"
 
+if os.environ.get("UCX_MEM_MMAP_HOOK_MODE", False):
```
pentschev, line 42:
```suggestion
if os.environ.get("UCX_MEM_MMAP_HOOK_MODE", False) == "none":
```

ucp/__init__.py
```diff
@@ -39,6 +39,13 @@
...
+    msg = (
+        "Setting `UCX_MEM_MMAP_HOOK_MODE` disable IB registration cache "
+        "which disable GPURDMAincluding GPU direct. \nPlease unset "
+        "UCX_MEM_MMAP_HOOK_MODE"
```
pentschev, line 46: I would prefer to rephrase that to something like:

"WARNING: Disabling UCX memory hooks also disables IB registration cache and zero-copy flows in UCX. This can cause serious performance degradation."

pentschev (changes requested):
@quasiben I added a few suggestions there.

ucp/__init__.py
quasiben: fixed in 37c207c

ucp/__init__.py
quasiben: Done

pentschev (approved):
LGTM, thanks @quasiben !

jakirkham:
Am seeing the following failure [on CI]( https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/2616/pipeline ).

```python
___________________________ test_mem_mmap_hook_warn ____________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x7f138eb82850>

    def test_mem_mmap_hook_warn(caplog):
        """
        Test warning for UCX_MEM_MMAP_HOOK_MODE
        """
        import logging
    
        os.environ["UCX_MEM_MMAP_HOOK_MODE"] = "none"
    
        # ucp.init will only print INFO LINES
        with caplog.at_level(logging.INFO):
            import ucp
    
>           ucp.init()

tests/test_mmap_warn.py:16: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

options = {}, env_takes_precedence = False, blocking_progress_mode = None

    def init(options={}, env_takes_precedence=False, blocking_progress_mode=None):
        """Initiate UCX.
    
        Usually this is done automatically at the first API call
        but this function makes it possible to set UCX options programmable.
        Alternatively, UCX options can be specified through environment variables.
    
        Parameters
        ----------
        options: dict, optional
            UCX options send to the underlying UCX library
        env_takes_precedence: bool, optional
            Whether environment variables takes precedence over the `options`
            specified here.
        blocking_progress_mode: bool, optional
            If None, blocking UCX progress mode is used unless the environment variable
            `UCXPY_NON_BLOCKING_MODE` is defined.
            Otherwise, if True blocking mode is used and if False non-blocking mode is used.
        """
        global _ctx
        if _ctx is not None:
            raise RuntimeError(
>               "UCX is already initiated. Call reset() and init() "
                "in order to re-initate UCX with new options."
            )
E           RuntimeError: UCX is already initiated. Call reset() and init() in order to re-initate UCX with new options.

ucp/core.py:755: RuntimeError
```

pentschev:
Hmmm, yeah testing for UCX variables can be really tricky. The "solution" for that would be calling `ucp.reset` and then `ucp.init`. However, given pytest's behavior of not creating a fresh process for each new test, this could mean that we would be modifying behavior of tests that happen to run after this one. Due to that, I would prefer if we just don't add a test for this, given it's just a warning the test isn't as critical as for other functionality. Thoughts on this @quasiben @jakirkham ?

jakirkham:
Yeah that's ok with me.

PR #629, opened 2020-10-02
Use C attributes, methods, & minor error message cleanup

jakirkham:
Instead of using Python properties in Cython, directly access C level properties. This should cutdown on the overhead of these attribute access calls. Particularly in parts of the code that may be more sensitive to them. Also provide access to C implementations of a few methods, these are comparable to `virtual` methods in C++. Finally cleanup the generation of a remaining error message so that the message is only built if the error occurs. Thus bypassing construction of this message when no error happens.


pentschev (approved):
LGTM, thanks @jakirkham .

PR #635, opened 2020-10-02
Cast `req_info["expected_receive"]` as `size_t`

jakirkham:
Otherwise Cython tries to convert the `size_t` variable to a Python object, which is a bit laborious. This bypasses that.


pentschev (approved):
LGTM, thanks @jakirkham .

jakirkham:
Thanks Peter! 😄

PR #639, opened 2020-10-05
Use new NVTX module

pentschev:
This is being removed from cuDF in https://github.com/rapidsai/cudf/pull/6413, we should use its own new `nvtx` package available from `conda-forge`.

### DIFF
--- a/ucp/utils.py
+++ b/ucp/utils.py
@@ -228,10 +228,9 @@ def run_on_local_network(
 
 
 try:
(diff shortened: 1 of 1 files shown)


jakirkham (approved):

jakirkham:
Adding to RAPIDS integration packages with PR ( https://github.com/rapidsai/integration/pull/141 ).

PR #641, opened 2020-10-05
Fix compile warnings

jakirkham:
Fixes a few issues that were causing compilation warnings. First fix typing of `ucp_listener_h` to be a pointer otherwise Cython will ignore later pointer casts, which results in C code that warns about not casting. Second fixes the typing of `ucp_listener_h`, the type of `ucp_err_handler_cb_t` wasn't correctly declared as a function pointer, which caused some unexpected C warnings about unassigned variables. This fixes that typing, which fixes the C warnings as well. Third note that `_get_error_callback` could raise Python exceptions as it works with Python objects (though this is exceedingly unlikely).

### DIFF
--- a/ucp/_libs/ucx_api_dep.pxd
+++ b/ucp/_libs/ucx_api_dep.pxd
@@ -208,6 +208,7 @@ cdef extern from "ucp/api/ucp.h":
 
     ctypedef enum ucs_config_print_flags_t:
         pass
+
     ucs_config_print_flags_t UCS_CONFIG_PRINT_CONFIG
     void ucp_config_print(const ucp_config_t *config,
                           FILE *stream,
--- a/ucp/_libs/ucx_api_dep.pxd
+++ b/ucp/_libs/ucx_api_dep.pxd
@@ -126,8 +126,10 @@ cdef extern from "ucp/api/ucp.h":
                                    ucp_worker_h *worker_p)
     void ucp_worker_destroy(ucp_worker_h worker)
 
-    ctypedef struct ucp_listener_h:
+    ctypedef struct ucp_listener:
         pass
+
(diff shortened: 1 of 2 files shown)


ucp/_libs/ucx_api_dep.pxd
```diff
@@ -28,8 +28,7 @@ cdef extern from "src/c_util.h":
 
     ctypedef ucp_conn_request* ucp_conn_request_h
 
-    ctypedef struct ucp_err_handler_cb_t:
-        pass
+    ctypedef void(*ucp_err_handler_cb_t)(void *arg, ucp_ep_h ep, ucs_status_t status)
```
jakirkham, line 31: This definition is taken from [the UCX definition]( https://github.com/openucx/ucx/blob/8b6641b02ef5d217323f719d54dfb2e9142dad59/src/ucp/api/ucp_def.h#L335-L348 ).

ucp/_libs/ucx_api_dep.pxd
```diff
@@ -127,9 +126,11 @@ cdef extern from "ucp/api/ucp.h":
                                    ucp_worker_h *worker_p)
     void ucp_worker_destroy(ucp_worker_h worker)
 
-    ctypedef struct ucp_listener_h:
+    ctypedef struct ucp_listener:
         pass
 
+    ctypedef ucp_listener* ucp_listener_h
+
```
jakirkham, line 133: Taken from [this UCX definition]( https://github.com/openucx/ucx/blob/8b6641b02ef5d217323f719d54dfb2e9142dad59/src/ucp/api/ucp_def.h#L168-L175 ).

pentschev (approved):
LGTM, thanks @jakirkham .

PR #646, opened 2020-10-13
Cleaning CI output

madsbk:


pentschev:
Are you trying to get rid of something in specific? Generally having `-vs` makes it easier to identify from what test errors are coming from, which is good to have when we have tests failing in CI.

madsbk:
> Generally having `-vs` makes it easier to identify from what test errors are coming from, which is good to have when we have tests failing in CI.

My hope is that pytest is able to capture output both from Python and UCX and only show it for the tests that fails. I find the current solution where _everything_ are printed to screen is confusing.

@pentschev do you know of a case where it is advantages to have everything printed? Will pytest sometimes suppress useful info? I am still a novice in pytest :)

PR #648, opened 2020-10-13
On sided endpoint rework and onesided endpoint creation

MattBBaker:
The Endpoint rework/new flags that may impact others.
There is a function `create_one_sided_ep()` that can create endpoints from a UCXAddress (or a buffer with an UCXAddress) without the asyncio loop
The Endpoint class is modified to make the control/send/receive tags optional since the one sided EP won't have a pair (it only communicates through the worker)
The ApplicationContext at init asks for features UCP_FEATURE_RMA | UCP_FEATURE_AMO32 | UCP_FEATURE_AMO64 since these will be needed for `mem_map()` and one sided operations

### DIFF
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -220,16 +220,16 @@ cdef class UCXContext(UCXObject):
         memset(&ucp_params, 0, sizeof(ucp_params))
         ucp_params.field_mask = (UCP_PARAM_FIELD_FEATURES |  # noqa
                                  UCP_PARAM_FIELD_REQUEST_SIZE |  # noqa
-                                 UCP_PARAM_FIELD_REQUEST_INIT |  # noqa
-                                 UCP_FEATURE_RMA |  # noqa
-                                 UCP_FEATURE_AMO32 |  # noqa
-                                 UCP_FEATURE_AMO64)
+                                 UCP_PARAM_FIELD_REQUEST_INIT)
 
         # We always request UCP_FEATURE_WAKEUP even when in blocking mode
         # See <https://github.com/rapidsai/ucx-py/pull/377>
         ucp_params.features = (UCP_FEATURE_TAG |  # noqa
                                UCP_FEATURE_WAKEUP |  # noqa
-                               UCP_FEATURE_STREAM)
+                               UCP_FEATURE_STREAM |  # noqa
+                               UCP_FEATURE_RMA |  # noqa
+                               UCP_FEATURE_AMO32 |  # noqa
+                               UCP_FEATURE_AMO64)
 
         ucp_params.request_size = sizeof(ucx_py_request)
         ucp_params.request_init = (
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -206,12 +206,12 @@ def _ucx_context_handle_finalizer(uintptr_t handle):
 
 
 ucp_features = {
-    "UCP_FEATURE_TAG" : UCP_FEATURE_TAG,
-    "UCP_FEATURE_WAKEUP" : UCP_FEATURE_WAKEUP,
-    "UCP_FEATURE_STREAM" : UCP_FEATURE_STREAM,
-    "UCP_FEATURE_RMA" : UCP_FEATURE_RMA,
-    "UCP_FEATURE_AMO32" : UCP_FEATURE_AMO32,
-    "UCP_FEATURE_AMO64" : UCP_FEATURE_AMO64
+    "UCP_FEATURE_TAG": UCP_FEATURE_TAG,
+    "UCP_FEATURE_WAKEUP": UCP_FEATURE_WAKEUP,
+    "UCP_FEATURE_STREAM": UCP_FEATURE_STREAM,
+    "UCP_FEATURE_RMA": UCP_FEATURE_RMA,
+    "UCP_FEATURE_AMO32": UCP_FEATURE_AMO32,
+    "UCP_FEATURE_AMO64": UCP_FEATURE_AMO64
 }
 
 
@@ -238,7 +238,6 @@ cdef class UCXContext(UCXObject):
         else:
             for feature in feature_list:
                 features_flag |= ucp_features[feature]
-
 
         memset(&ucp_params, 0, sizeof(ucp_params))
         ucp_params.field_mask = (UCP_PARAM_FIELD_FEATURES |  # noqa
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -1,4 +1,5 @@
 # Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020       UT-Battelle, LLC. All rights reserved.
 # See file LICENSE for terms.
 
 # cython: language_level=3
@@ -9,6 +10,7 @@ import weakref
 
 from posix.stdio cimport open_memstream
 
+from cpython.buffer cimport PyBUF_FORMAT, PyBUF_ND, PyBUF_WRITABLE
 from cpython.ref cimport Py_DECREF, Py_INCREF, PyObject
 from libc.stdint cimport uint16_t, uintptr_t
 from libc.stdio cimport FILE, clearerr, fclose, fflush
@@ -424,6 +426,91 @@ cdef class UCXWorker(UCXObject):
         assert_ucs_status(status)
         return UCXEndpoint(self, <uintptr_t>ucp_ep)
 
+    cpdef ucs_status_t fence(self) except *:
+        cdef ucs_status_t status = ucp_worker_fence(self._handle)
+        assert_ucs_status(status)
+        return status
+
+    def flush(self, cb_func, tuple cb_args=None, dict cb_kwargs=None):
+        if cb_args is None:
+            cb_args = ()
+        if cb_kwargs is None:
+            cb_kwargs = {}
+        cdef ucs_status_ptr_t req
+        cdef ucp_send_callback_t _send_cb = <ucp_send_callback_t>_send_callback
+
+        cdef ucs_status_ptr_t status = ucp_worker_flush_nb(self._handle, 0, _send_cb)
+        return _handle_status(
+            status, 0, cb_func, cb_args, cb_kwargs, 'flush', self._inflight_msgs
+        )
+
+    def get_address(self):
+        return UCXAddress(self)
+
+
+def _ucx_address_finalizer(uintptr_t handle_as_int, uintptr_t worker_as_int):
+    cdef ucp_address_t *address = <ucp_address_t *>handle_as_int
+    cdef ucp_worker_h ucp_worker = <ucp_worker_h>worker_as_int
+    ucp_worker_release_address(ucp_worker, address)
+
+
+cdef class UCXAddress(UCXObject):
+    """Python representation of ucp_address_t"""
+    cdef ucp_address_t *_handle
+    cdef Py_ssize_t _length
+    cdef worker
+
+    def __init__(self, UCXWorker worker):
+        cdef ucs_status_t status
+        cdef ucp_worker_h ucp_worker = worker._handle
+        cdef size_t length
+
+        status = ucp_worker_get_address(ucp_worker, &self._handle, &length)
+        assert_ucs_status(status)
+        self.worker = worker
+        self._length = <Py_ssize_t>length
+        self.add_handle_finalizer(
+            _ucx_address_finalizer,
+            int(<uintptr_t>self._handle),
+            int(<uintptr_t>worker._handle)
+        )
+        worker.add_child(self)
+
+    @property
+    def address(self):
+        assert self.initialized
+        return <uintptr_t>self._handle
+
+    @property
+    def length(self):
+        assert self.initialized
+        return int(self._length)
+
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        assert self.initialized
+        if (flags & PyBUF_WRITABLE) == PyBUF_WRITABLE:
+            raise BufferError("Requested writable view on readonly data")
+        buffer.buf = <void*>self._handle
+        buffer.obj = self
+        buffer.len = self._length
+        buffer.readonly = True
+        buffer.itemsize = 1
+        if (flags & PyBUF_FORMAT) == PyBUF_FORMAT:
+            buffer.format = b"B"
+        else:
+            buffer.format = NULL
+        buffer.ndim = 1
+        if (flags & PyBUF_ND) == PyBUF_ND:
+            buffer.shape = &self._length
+        else:
+            buffer.shape = NULL
+        buffer.strides = NULL
+        buffer.suboffsets = NULL
+        buffer.internal = NULL
+
+    def __releasebuffer__(self, Py_buffer *buffer):
+        pass
+
 
 def _ucx_endpoint_finalizer(uintptr_t handle_as_int, worker, set inflight_msgs):
     assert worker.initialized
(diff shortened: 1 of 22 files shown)


quasiben:
Do you mind if we target 0.17 for this ?  RAPIDS is in code freeze and I'd like to push anything to the  next branch

MattBBaker:
I'm fine with this, how long will that be?

quasiben:
Release is scheduler for Oct. 21st but we will nightly packages for the upcoming release as well as future work like this.  I'll make the change now

pentschev:
I think @MattBBaker was asking when will 0.17 be released, and it's currently scheduled for December 9th.

jakirkham:
Also we post the release schedule [here]( https://docs.rapids.ai/maintainers ).

jakirkham:
For context, we are entering code freeze today for 0.16.0. Hence why we are asking about moving things to the next release to cutdown on churn while OPS finalizing things so we can ship out the stack. There's a fair bit of work involved in the release process and we want to avoid adding to it if we can.

MattBBaker:
Okay, I understand. Anyways, I'm feeling good about this patch set now, who should we ping to make sure that the changes are not breaking anyone's code?

ucp/core.py
```diff
@@ -473,28 +508,25 @@ class Endpoint:
...
         self._shutting_down_peer = False  # Told peer to shutdown
         self._close_after_n_recv = None
 
+    def wireup_tags(self, msg_tag_send, msg_tag_recv, ctrl_tag_send, ctrl_tag_recv):
+        self._use_tags = True
+        self._msg_tag_send = msg_tag_send
+        self._msg_tag_recv = msg_tag_recv
+        self._ctrl_tag_send = ctrl_tag_send
+        self._ctrl_tag_recv = ctrl_tag_recv
+
```
madsbk, line 529: I suggest that we unify all the tags attributes in a `dict` given as argument to `__init__()` and can be `None` when tags isn't used.

ucp/_libs/ucx_api.pyx
```diff
@@ -226,7 +226,10 @@ cdef class UCXContext(UCXObject):
         # See <https://github.com/rapidsai/ucx-py/pull/377>
         ucp_params.features = (UCP_FEATURE_TAG |  # noqa
                                UCP_FEATURE_WAKEUP |  # noqa
-                               UCP_FEATURE_STREAM)
+                               UCP_FEATURE_STREAM |  # noqa
+                               UCP_FEATURE_RMA |  # noqa
+                               UCP_FEATURE_AMO32 |  # noqa
+                               UCP_FEATURE_AMO64)
```
madsbk, line 232: I think we should make the options optional. `__init__()` could accept a set of future constants and default to:
```python
def __init__(self, config_dict, ucp_features={UCP_FEATURE_TAG, UCP_FEATURE_WAKEUP):
```
or 

```python
def __init__(self, config_dict, ucp_features={"UCP_FEATURE_TAG", "UCP_FEATURE_WAKEUP"):
```

madsbk (changes requested):
It looks good @MattBBaker, I have some minor issues only.

ucp/_libs/ucx_api.pyx
MattBBaker: I was actually thinking about a mechanism like that. How would the features be passed in? If the strings are used then that would involve a lookup table/dict() while the named version would involve the user doing something like features=[ucp.feature.wakup, ucp.feature.tag]. Which one would be prefered?

ucp/core.py
MattBBaker: Any thoughts about adding tags after the fact? I do that on my test for convince, though I'm not sure I would use it in practice

ucp/_libs/ucx_api.pyx
madsbk: I have no strong opinion on this but I prefer it as string with a lookup dict.

ucp/core.py
madsbk: I don't think we should support _tags after the fact_ unless there is a real use case.

MattBBaker:
@quasiben @madsbk @jakirkham Looks like the bots are all happy now, ready for the next round of review

ucp/core.py
```diff
@@ -472,28 +519,23 @@ class Endpoint:
...
         self._shutting_down_peer = False  # Told peer to shutdown
         self._close_after_n_recv = None
+        if tags is not None:
+            self._use_tags = True
+            self._msg_tag_send = tags["msg_tag_send"]
+            self._msg_tag_recv = tags["msg_tag_recv"]
+            self._ctrl_tag_send = tags["ctrl_tag_send"]
+            self._ctrl_tag_recv = tags["ctrl_tag_recv"]
+        else:
+            self._use_tags = False
```
madsbk, line 538: What about keeping the dict of tags like:
```suggestion
        self._tags = tags        
```
Than `self._tags==None` means _no tags_

ucp/core.py
```diff
@@ -531,30 +573,35 @@ async def close(self):
...
-            try:
-                await comm.tag_send(
-                    self._ep, msg_arr, msg_arr.nbytes, self._ctrl_tag_send, name=log,
-                )
-            # The peer might already be shutting down thus we can ignore any send errors
-            except UCXError as e:
-                logging.warning(
-                    "UCX failed closing worker %s (probably already closed): %s"
-                    % (hex(self.uid), repr(e))
+            if self._use_tags:
```
madsbk, line 576:
```suggestion
            if self._tags:
```

madsbk (changes requested):
I have some minor suggestions

ucp/core.py
madsbk: You will also have to change all the other `self._msg_tag_*` and `self._ctrl_tag_*` accesses  to dict lookup

ucp/core.py
```diff
@@ -389,6 +393,49 @@ async def create_endpoint(
         CtrlMsg.setup_ctrl_recv(ep)
         return ep
 
+    def create_endpoint_sync(
+        self,
+        address,
+        guarantee_msg_order=False,
+        endpoint_error_handling=False,
+        tags=None,
```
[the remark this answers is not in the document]
MattBBaker: Yes, I should add the documentation. I wanted to be able to wire up an EP to still use the higher level tag API if you have a mechanism to distribute tags OOB.

ucp/core.py
MattBBaker: I did think about that, the main though is that it would require an additional dict lookup for every tag operation, which my MPI thinking brain says, "4 byte sends are valid, and some people care a lot about that kind of injection rate"

madsbk: I think we can afford dict lookup in this case. Alternatively, we could create a `struct` of the tag values that is `None` when tags aren't used.

MattBBaker:
Whoops, tried updating my PR and something went wrong. Will work on

jakirkham:
No worries. Is it ok if we retarget to `branch-0.18`? RAPIDS 0.17 is already shipped. Incidentally this may fix the same PR updating issue as well.

jakirkham:
> No worries. Is it ok if we retarget to `branch-0.18`? RAPIDS 0.17 is already shipped. Incidentally this may fix the same PR updating issue as well.

Went ahead and made this change. Hope that is ok 🙂

Looks like this has cleaned up the commit history issue

jakirkham:
Looks like `black` is running into an issue. Would you be able to run `black` locally and commit + push those changes? Should fix this issue

madsbk:
> Looks like `black` is running into an issue. Would you be able to run `black` locally and commit + push those changes? Should fix this issue

Remember to use `black` version **19.10b0**

MattBBaker:
Yea, it looks like I rebased on 0.18 by accident. And I swear I ran black before pushing this. I'll have a check on the version. and fix things up

MattBBaker:
Turns out running black in the ucp directory will make it miss the test directory

jakirkham:
JFYI I think all of us are out over the next couple weeks. Guessing you are also taking time off at some point. Thought you should know in case it gets a bit quiet here 🙂 Happy holidays btw! 🎉

ajschmidt8:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

jakirkham:
Would you have another chance to look at this Mads? 🙂

ucp/core.py
```diff
@@ -232,6 +235,7 @@ class ApplicationContext:
 
     def __init__(self, config_dict={}, blocking_progress_mode=None):
         self.progress_tasks = []
+        self.progress_tasks = []
```
madsbk, line 238:
```suggestion
```

madsbk (approved):
Great work @MattBBaker!
I only have a minor issue where you a duplicate `self.progress_tasks = []` line.
Thanks!

MattBBaker:
Yea, I wanted to push something out because I figured it would be a little bit before I could circle back to this and wanted to push something out before the winter break

ucp/core.py
MattBBaker: Whoops. For some reason that has creeped in more than once.

ucp/_libs/ucx_api.pyx
```diff
@@ -205,28 +205,51 @@ def _ucx_context_handle_finalizer(uintptr_t handle):
     ucp_cleanup(<ucp_context_h> handle)
 
 
+ucp_features = {
```
jakirkham, line 208:
```suggestion
cdef dict ucp_features = {
```

jakirkham (approved):
Thanks Matt! LGTM as well. 😄 Made one minor suggestion below.

ucp/_libs/ucx_api.pyx
MattBBaker: This change seems good.

MattBBaker:
I'm looking through the error messages, but I'm having a hard time interpreting what all they mean. I see a lot of time out errors but I see no indication of what is timing out and why.

pentschev:
> I'm looking through the error messages, but I'm having a hard time interpreting what all they mean. I see a lot of time out errors but I see no indication of what is timing out and why.

The errors are legit, I checked locally with branch-0.18 and this PR and it's only reproduced with this PR. I did a little digging but I couldn't find what's the reason either. At first I thought these were a consequence of the new flags introduced in the `UCXContext` creation, but after reverting those, I still see the same issues. I'm really not sure what else could be the cause, after reviewing the PR again I don't see any other fundamental changes, but reverting both `ucp/core.py` and `ucp/endpoint_reuse.py` does resolve the issue, while keeping the new flags (but pretty much erasing what this PR introduces).

MattBBaker:
I've been digging into the source of failures and I have come up with something interesting but I'm not sure how to proceed. Specifically in one of the new tag matching tests, `test_no_tag()` I get this trace back that is new
```sys:1: RuntimeWarning: coroutine 'BlockingMode._arm_worker' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
task: <Task cancelling coro=<BlockingMode._arm_worker() running at /home/shadesfox/ucx-py/ucp/continuous_ucx_progress.py:88>>```
I'm not sure how/why this warning happens, but I suspect it is related to the timeouts, since it looks like the worker never gets armed. Can anyone help me figure out why this happens? 
@pentschev @jakirkham @madsbk

**Comment by pentschev** [comment]
I actually have seen this sort of issue over the past few weeks for certain tests/conditions, and I checked now that it's been happening since January 23rd with my nightly runs on `test_get_ucx_version`, plus it happens with some of dask-cuda benchmarks when there's something else running on the same machine causing an RMM OOM error, thus it indicates some sort of cleanup issue in that case. I remember I've seen it under other conditions (I can't rememeber exactly what, since I've been running dozens of different combinations lately) and had little success debugging it (ah, how much I love Python's async...).

What version of UCX are you running? UCX 1.9 is still ok with UCX-Py, but 1.10 and above are very flaky and we're working with UCX devs to fix some of those issues, so I would advise against anything above 1.9.

Could you also try running with https://github.com/rapidsai/ucx-py/pull/671 ? Generally speaking, that's a necessary fix, but I've seen hangs with it in some cases which I haven't managed to figure out why yet.

**Comment by MattBBaker** [comment]
I tried it with the patch, it seems to help but the problem doesn't completely go away

File: ucp/_libs/ucx_api.pyx [hunk 8]
```diff
@@ -205,28 +205,51 @@ def _ucx_context_handle_finalizer(uintptr_t handle):
...
+        else:
+            for feature in feature_list:
+                features_flag |= ucp_features[feature]
 
         memset(&ucp_params, 0, sizeof(ucp_params))
         ucp_params.field_mask = (UCP_PARAM_FIELD_FEATURES |  # noqa
                                  UCP_PARAM_FIELD_REQUEST_SIZE |  # noqa
-                                 UCP_PARAM_FIELD_REQUEST_INIT)
+                                 UCP_PARAM_FIELD_REQUEST_INIT |  # noqa
+                                 UCP_FEATURE_RMA |  # noqa
```
vchuravy, line 246: Why are the feature flags listed in the `field_mask`?

pentschev: Good catch @vchuravy, looks like a mistake to me. Could you check that @MattBBaker ?

madsbk:
I think this PR is getting a bit too complex to merge easily. I will try to get all feature here implemented using a divide-and-conquer approach starting with https://github.com/rapidsai/ucx-py/pull/688.
I hope that is okay @MattBBaker?

MattBBaker:
@madsbk I am 100% in favor of this. This was getting quite unwieldy anyways. And hopefully it will help narrow down what makes Dask especially unhappy

madsbk:
@MattBBaker, with https://github.com/rapidsai/ucx-py/pull/691, we now support the creation of an endpoints from a `UCXAddress`. What is the next step? 
This PR brings this functionally to the _asyncio-level_ API through `core.create_endpoint_sync()` and `create_one_sided_ep()` but I am wondering if we should let it be a low-level non-asyncio feature for now?

Some context; the overall plan is to split UCX-Py into a high-level asyncio module and a low-level Cython module (`ucx_api.pyx`). In the next couple of weeks, I will work on making `ucx_api.pyx` self-contained with its own tests and docs.

MattBBaker:
@madsbk What i the diff between what is in now and this PR?

Overall I'm in favor of allowing this to be a low level feature. For short term planning, the next step is to include the put/get APIs. On the subject of the split between the low level and high level modules, what is the plan to upstream work to the main UCX repo? That plan will inform the rest of the plan on how to fill in more functionality.

madsbk:
> @madsbk What i the diff between what is in now and this PR?

It doesn't include the changes to `ucp/core.py` like `create_endpoint_sync` and `create_one_sided_ep`.

GPUtester:

MattBBaker:
Okay, so right now it live in the lower layer with no exposure to the higher level ucp? In that case it might be best to hold off for now on trying to get the higher level API in and let it remain a low level API.

kkraus14:
add to allowlist

madsbk:
> Okay, so right now it live in the lower layer with no exposure to the higher level ucp? In that case it might be best to hold off for now on trying to get the higher level API in and let it remain a low level API.

Yes exactly, let's implement some examples that make use of the one-sided low level API and get some experience before move to the high level API.

I am closing this PR but let's use the remaining features not already in `branch-0.19` as the bases for future work. Thanks for all the work @MattBBaker!

jakirkham:
Would it make sense to target `branch-0.20` for this work?

madsbk:
> Would it make sense to target `branch-0.20` for this work?

Yes the work should target future versions but I think it is better to submit new PRs that cherry-picks from this PR instead of trying to get this PR merged.

MattBBaker:
Yea, this PR has gotten beastly. I'll pull in more work from my previous RMA branch too and focus on this strategy of filling out the low level API

PR #650, opened 2020-10-20
Remove default values for UCX_TCP_{TX,RX}_SEG_SIZE

pentschev:
These raise lots of warnings and only benefit the CUDA with TCP-only case, which is not very common right now. If this becomes necessary, we should consider adding them directly into the application, such as dask-cuda when both NVLink and IB are disabled.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -84,8 +84,6 @@ UCX_TCP_TX_SEG_SIZE
 
 Size of send copy-out buffer when transmitting.  This environment variable controls the size of the buffer on the host when sending data over TCP.
 
-UCX-Py uses ``8M`` as the default value for both RX/TX.
-
 .. note::
     Users should take care to properly tune ``UCX_TCP_{RX/TX}_SEG_SIZE`` parameters when mixing TCP with other transports methods as well as when
     using TCP over UCX in isolation.  These variables will impact CUDA transfers when no NVLink or InfiniBand is available between UCX-Py processes.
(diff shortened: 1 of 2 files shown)


jakirkham:
What happens in the CUDA TCP case when they are not set?

pentschev:
It's slow, because all transfers have to pass to host, implying various DtoH `cudaMemcpy`s of size `UCX_TCP_TX_SEG_SIZE=8K` (8K is the default).

quasiben:
I think this change is ok. As @pentschev TCP-only is not the common case right now and removing is not to injurious.  Users are still free to set this.  In general, i think it might be good to start prioritizing defaults for the common case

jakirkham:
Yeah I guess we [already document these]( https://ucx-py.readthedocs.io/en/latest/configuration.html?highlight=UCX_TCP_#ucx-tcp-rx-seg-size ). Maybe it's worth mentioning in the docs that users should set these for the TCP-only case?

pentschev:
That's what is explained in the "Note", not sure what more you'd like to have there?

jakirkham:
Sure I'm wondering if we should be mentioning this outside of the configuration section though. For example [this page]( https://ucx-py.readthedocs.io/en/latest/dask.html ) shows using TCP over UCX. Should it include a brief note and link to the configuration section?

pentschev:
I don't particularly think that's necessary, we might eventually end up cross-referencing everything to everything else. Although I'm not against that either if someone really wants to do it. :)

quasiben (approved):
In https://github.com/rapidsai/ucx-py/pull/654 I am trying to address @jakirkham concerns.  I think we should merge these changes in and continue the discussion on doc improvements in that PR

PR #657, opened 2020-11-13
DOC: Fix UCX_TLS environment variable for Infiniband

dmitrygx:
The suggested configuration doesn't really work for intra-node communication, only inter-node communications are possible, since it uses CUDA_IPC.
```
$ UCX_TLS=sockcm,cuda_copy,cuda_ipc ucx_info -eptw -u tra -D net
[1605294229.512128] [r-vmb-ppc-jenkins:12834:0]    ucp_context.c:1108 UCX  ERROR no usable transports/devices (asked sockcm,cuda_copy,cuda_ipc on all devices)
<Failed to create UCP context>
```
User has to specify `ib` to `UCX_TLS` environment variable to use Infiniband

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ export UCX_SOCKADDR_TLS_PRIORITY=sockcm
 export UCXPY_IFNAME="eth0"
 
 # InfiniBand using "ib0" and CUDA support
-export UCX_TLS=sockcm,cuda_copy,cuda_ipc
+export UCX_TLS=ib,sockcm,cuda_copy,cuda_ipc
 export UCX_SOCKADDR_TLS_PRIORITY=sockcm
 export UCXPY_IFNAME="ib0"
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ export UCX_SOCKADDR_TLS_PRIORITY=sockcm
 export UCXPY_IFNAME="eth0"
 
 # InfiniBand using "ib0" and CUDA support
-export UCX_TLS=ib,sockcm,cuda_copy,cuda_ipc
+export UCX_TLS=rc,sockcm,cuda_copy,cuda_ipc
 export UCX_SOCKADDR_TLS_PRIORITY=sockcm
 export UCXPY_IFNAME="ib0"
 


GPUtester:

GPUtester:

pentschev:
That's a good catch @dmitrygx , thanks for the PR. We're using generally `rc` with UCX-Py and its docs, would you suggest `ib` instead of `rc`?

pentschev:
add to allowlist

ajschmidt8:
add to allowlist

ajschmidt8:
add to whitelist

mike-wendt:
add to allowlist

mike-wendt:
rerun tests

mike-wendt:
ok to test

mike-wendt:
rerun tests

mike-wendt:
This is a markdown-only change and will need to me admin merged

dmitrygx:
> We're using generally rc with UCX-Py and its docs, would you suggest ib instead of rc?

No, I don't suggest it :) "ib" should be used if UCX-Py needs to fallback from RC to DC/UD on large scale runs. If UCX_TLS="rc", it uses RC QP and it could be scalability uses then.
Maybe it worth to mention in README that there are RC-only/UD-only/DC-only and IB (that could use all of QPs depending on a system characteristics)?

dmitrygx:
@pentschev ah, I see...

>  "ib" should be used if UCX-Py needs to fallback from RC to DC/UD on large scale runs. 

It is not true for UCX-Py, since it does provide to UCX how many instances are will be opened through `UCX_NUM_EPS` env (or `estimated_num_eps` field in a `ucp_params_t`).
"rc" will be used in either way
Should I change "ib" to "rc"? wdyr?

pentschev:
> Should I change "ib" to "rc"? wdyt?

I think changing to `rc` for now would be best. We haven't used anything else so far, so probably best to stick to what we did test. :)

dmitrygx:
> I think changing to rc for now would be best.

@pentschevmakes sense to me. thank you for the review, fixed

pentschev (approved):
LGTM, thanks for the fix @dmitrygx !

pentschev:
rerun tests

pentschev:
> This is a markdown-only change and will need to me admin merged

Sorry, I missed this before. Can you merge it @mike-wendt ?

dmitrygx:
> LGTM, thanks for the fix @dmitrygx !

@pentschev @ajschmidt8 @mike-wendt thank you for the review, folks! glad to help

PR #664, opened 2020-12-08
Move test_endpoint_error_callback to main debug-tests dir

pentschev:
Moving test to the main `debug-tests` directory, which I seems to have missed when others were moved.


quasiben (approved):
Thanks @pentschev

PR #666, opened 2021-01-13
Update install instructions for new UCX version

manopapad:
Note that the "Edit on GitHub" links on https://ucx-py.readthedocs.io point to https://github.com/quasiben/ucx-py/tree/fix-rtd, so I don't know if these changes will get automatically applied.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -113,7 +113,7 @@ Instructions for building ucx 1.8:
     make -j install
 
 .. note::
-    If you're running on a machine without CUDA then you _must NOT_ apply the patch.
+    If you're running on a machine without CUDA then you _must NOT_ apply any of the patches above.
 
 UCX + OFED
 ~~~~~~~~~~
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -31,7 +31,7 @@ Without GPU support:
     conda create -n ucx -c conda-forge -c rapidsai \
       ucx-proc=*=cpu ucx ucx-py python=3.7
 
-Note: These use UCX's ``v1.8.x`` branch.
+Note: These use UCX's ``v1.10.x`` branch.
 
 Source
 ------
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @quasiben (in case there’s something special we need to do with RTD per Manolis’ comment)

pentschev (approved):
LGTM, thanks @manopapad !

quasiben:
Thanks @manopapad -- I'm wondering if we should keep the 1.8 build instructions for a little bit while we sort out 1.10 UCX issues.  I'm happy to make the changes to this PR if you agree

pentschev (changes requested):
Yes, that's a good idea, please refer to #668 . I'm blocking merging until we get things sorted out.

quasiben:
Pushed a commit to update instructions for both 1.9 and 1.8 ucx building

jakirkham (approved):

jakirkham:
Is this ok to merge Peter or are there other things we should be doing here? Seems like Ben has updated the instructions. Also it sounds like 1.9 doesn't have the problems we say in RCs of 1.10. Does that sound correct or are there things I've mixed up here?

docs/source/install.rst
```diff
@@ -88,7 +113,7 @@ UCX
     make -j install
 
 .. note::
-    If you're running on a machine without CUDA then you _must NOT_ apply the patches.
+    If you're running on a machine without CUDA then you _must NOT_ apply the patch.
```
pentschev, line 116:
```suggestion
    If you're running on a machine without CUDA then you _must NOT_ apply any of the patches above.
```

pentschev:
I did a minor phrasing modification, given we have instructions for both 1.8 and 1.9 above. Other than that, LGTM. Thanks @manopapad and @quasiben for the changes and @jakirkham for staying on top of this.

pentschev (approved):

jakirkham:
Thanks Peter! 😄

PR #673, opened 2021-01-29
Prevent closing event_loop too early

pentschev:
This fixes test warnings such as:

```python
task: <Task pending name='Task-31' coro=<BlockingMode._arm_worker() running at ucp/continuous_ucx_progress.py:88>>
```


quasiben (approved):
Thanks @pentschev !

PR #677, opened 2021-02-05
Replace stream by tag in exchange_peer_info

pentschev:
This PR should allow us to move to tag send/recv entirely and get rid of the stream interface. There are still some tests occasionally failing, but seems mostly functional.


pentschev:
This isn't going to work as we don't have enough information _a prior_, more details in https://github.com/rapidsai/ucx-py/issues/497#issuecomment-825803351 . Closing.

PR #685, opened 2021-03-05
Remodel multinode tests

pentschev:
Testing many listeners and many clients on the same process can be prone to hangs. More realistic tests involved multiple processes to handle both sides, this PR adds new tests for that.


quasiben:
Thanks Peter, I think this is failing due to isort

quasiben (approved):

PR #688, opened 2021-03-12
Clean up send/recv tags

madsbk:
Now using a `dict` for the endpoint tags.

This PR is part **1** of an effort to merge https://github.com/rapidsai/ucx-py/pull/648 in a divide-and-conquer approach.

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -133,11 +133,14 @@ class CtrlMsg:
     @staticmethod
     def setup_ctrl_recv(ep):
         """Help function to setup the receive of the control message"""
-        log = "[Recv shutdown] ep: %s, tag: %s" % (hex(ep.uid), hex(ep._ctrl_tag_recv),)
+        log = "[Recv shutdown] ep: %s, tag: %s" % (
+            hex(ep.uid),
+            hex(ep._tags["ctrl_recv"]),
+        )
         msg = bytearray(CtrlMsg.nbytes)
         msg_arr = Array(msg)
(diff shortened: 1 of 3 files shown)


ucp/core.py
```diff
@@ -472,28 +478,17 @@ class Endpoint:
...
-        self._ctrl_tag_send = ctrl_tag_send
-        self._ctrl_tag_recv = ctrl_tag_recv
         self._guarantee_msg_order = guarantee_msg_order
         self._send_count = 0  # Number of calls to self.send()
         self._recv_count = 0  # Number of calls to self.recv()
         self._finished_recv_count = 0  # Number of returned (finished) self.recv() calls
         self._shutting_down_peer = False  # Told peer to shutdown
         self._close_after_n_recv = None
+        self._tags = tags
+        assert tags is not None
```
pentschev, line 491: I don't think this line makes sense. Why have `tags=None` in the kwargs and assert here it isn't `None`?

pentschev (commented):
I have a question here @madsbk , but otherwise I think it's a good idea to work on multiple PRs and unblock https://github.com/rapidsai/ucx-py/pull/648 . Thanks for working on this.

ucp/core.py
madsbk: Oops, you are right that was only for debugging :)

pentschev (approved):
Thanks @madsbk !

pentschev:
@gpucibot merge

PR #690, opened 2021-03-12
Support `ucp_listener_query` to find listener's IP/port

pentschev:
This will allow us to pass the responsibility of assigning a listener port to UCX when one isn't specified by the user, rather than doing so in UCX-Py.

### DIFF
--- a/ucp/_libs/tests/test_listener.py
+++ b/ucp/_libs/tests/test_listener.py
@@ -1,0 +1,18 @@
+from ucp._libs.ucx_api import UCXContext, UCXListener, UCXWorker
+
+
+def test_listener_ip_port():
+    def _listener_handler(conn_request, callback_func):
+        pass
+
+    context = UCXContext({})
+    worker = UCXWorker(context)
+
+    listener = UCXListener(
+        worker=worker, port=0, cb_func=_listener_handler, cb_args=(lambda: None,)
+    )
+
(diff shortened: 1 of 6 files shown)


ucp/_libs/tests/test_listener.py
```diff
@@ -0,0 +1,18 @@
...
+    worker = UCXWorker(context)
+
+    listener = UCXListener(
+        worker=worker, port=0, cb_func=_listener_handler, cb_args=(lambda: None,)
+    )
+
+    assert isinstance(listener.ip, str) and listener.ip
+    assert (
+        isinstance(listener.port, int) and listener.port >= 0 and listener.port <= 65535
+    )
```
pentschev, line 18: @madsbk this is a very rough test using the UCX-Py Cython backend only. We can do the same for more complex tests, as we discussed yesterday.

madsbk (approved):
This is great @pentschev, it will make it much easier to implement low-level tests! 
I have push some minor changes including fixing the header `c_util_sockaddr_get_ip_str => c_util_sockaddr_get_ip_port_str`.

pentschev:
Thanks @madsbk and sorry for adding more code to `c_util`, it was great to see it shrinking, but writing this in Cython would require adding many structs from system libraries, which I thought was not worth it now.

PR #692, opened 2021-03-16
[REVIEW] Low level tests and type hints

madsbk:
This PR implements new low-level tests and type hints


ucp/_libs/ucx_api.pyx
```diff
@@ -683,7 +688,34 @@ def _ucx_listener_handle_finalizer(uintptr_t handle):
...
+
+    Notice, the listening is closed when the returned Listener
+    goes out of scope thus remember to keep a reference to the object.
+
+    Parameters
+    ----------
+    worker: UCXWorker
+        Listening worker.
+    port: int, optional
+        An unused port number for listening
```
pentschev, line 703:
```suggestion
        An unused port number for listening or `0` to let
        UCX assign a random port.
```

ucp/_libs/tests/test_server_client.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def _listener_handler(conn_request):
+        pass
+
+    listener = ucx_api.UCXListener(worker=worker, port=0, cb_func=_listener_handler)
+
+    assert isinstance(listener.ip, str) and listener.ip
+    assert (
+        isinstance(listener.port, int) and listener.port >= 0 and listener.port <= 65535
+    )
```
pentschev, line 26: I think leaving this test in its own file as it was originally makes more sense, after all this doesn't semantically match a server/client test as this file name suggests.

ucp/_libs/tests/test_server_client.py
```diff
@@ -0,0 +1,94 @@
...
+            worker, msg, msg.nbytes, tag=0, cb_func=_recv_handle, cb_args=(ep, msg)
+        )
+
+    listener = ucx_api.UCXListener(worker=worker, port=0, cb_func=_listener_handler)
+    queue.put(listener.port)
+
+    while True:
+        worker.progress()
+        try:
+            queue.get(block=False, timeout=0.1)
```
pentschev, line 64: I'm not really sure I understand all this. The listener is created and immediately puts its port into the queue. Won't this cause `worker.progress()` to be called only once and break the loop?

pentschev (commented):
Looks great, a few minor questions/suggestions only.

ucp/_libs/ucx_api.pyi
```diff
@@ -0,0 +1,71 @@
+from typing import Callable, Iterable, Mapping
```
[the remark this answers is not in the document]
madsbk: Did it manually based on https://mypy.readthedocs.io/en/stable/stubs.html

ucp/_libs/tests/test_server_client.py
madsbk: Fixed

ucp/_libs/tests/test_server_client.py
madsbk: It works because the main process calls `queue.get()` right after starting the server: https://github.com/rapidsai/ucx-py/blob/8ac3163f5ffbad787e47d06968225783581aedbd/ucp/_libs/tests/test_server_client.py#L72

But you are right, there is a race condition here :)

madsbk: Fixed by using two queues

mike-wendt (approved):
Approved for changes in our `CODEOWNERS` files

ci/gpu/build.sh
```diff
@@ -100,6 +100,7 @@ else
     # Test with TCP/Sockets
     gpuci_logger "TEST WITH TCP ONLY"
     py.test --cache-clear -vs --ignore-glob tests/test_send_recv_two_workers.py tests/
+    py.test --cache-clear -vs ucp/_libs/tests
```
ajschmidt8, line 103: Can these two lines be consolidated? (I haven't tested this suggestion)

```suggestion
    py.test --cache-clear -vs --ignore-glob tests/test_send_recv_two_workers.py tests/ ucp/_libs/tests/
```

madsbk: I tried, but pytest complains about name collision.

pentschev (approved):
LGTM, thanks @madsbk !

PR #694, opened 2021-03-17
Respect somaxconn in test_multiple_nodes

pentschev:
Creating too many connections simultaneously may cause UCX to hang due to the SOMAXCONN limit. This PR resolves that issue in `test_multiple_nodes`.


pentschev:
Thanks @alinask and @dmitrygx for help in identifying the cause for this error!

dmitrygx (approved):

quasiben (approved):
Thanks @pentschev

PR #695, opened 2021-03-17
Document OS limits

pentschev:
This PR adds a new RTD section to document some known issues that may arise due to OS limits.

### DIFF
--- a/docs/source/os-limits.rst
+++ b/docs/source/os-limits.rst
@@ -13,12 +13,12 @@ In sockets-based connections, multiple file descriptors may be open to establish
 
     ucp.exceptions.UCXError: User-defined limit was reached
 
-One possible cause for this is that the limit establish by the OS or system administrators has been reached by the user. This limit can be checked with:
+One possible cause for this is that the limit established by the OS or system administrators has been reached by the user. This limit can be checked with:
 
 ::
     $ ulimit -n
 
-If the user has permission to do so, this limit can be increased by typing the new limit after the command above. For example, to set a new limit of 1 million, the following should be executed:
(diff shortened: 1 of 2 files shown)


docs/source/os-limits.rst
```diff
@@ -0,0 +1,47 @@
...
+File Descriptors
+----------------
+
+In sockets-based connections, multiple file descriptors may be open to establish connections between endpoints. When UCX is establishing connection between endpoints via protocols such as TCP, an error such as below may occur:
+
+::
+
+    ucp.exceptions.UCXError: User-defined limit was reached
+
+One possible cause for this is that the limit establish by the OS or system administrators has been reached by the user. This limit can be checked with:
```
quasiben, line 16:
```suggestion
One possible cause for this is that the limit established by the OS or system administrators has been reached by the user. This limit can be checked with:
```

docs/source/os-limits.rst
```diff
@@ -0,0 +1,47 @@
...
+
+::
+    $ ulimit -n
+
+If the user has permission to do so, this limit can be increased by typing the new limit after the command above. For example, to set a new limit of 1 million, the following should be executed:
```
quasiben, line 21:
```suggestion
If the user has permission to do so, the file descriptor limit can be increased by typing the new limit after the command above. For example, to set a new limit of 1 million, the following should be executed:
```

quasiben (commented):
Thank you for writing this up @pentschev . Left a few small suggestions

pentschev:
Thanks @quasiben , suggestions applied!

pentschev:
@quasiben could you approve/merge if you agree with the current text?

quasiben (approved):
Thanks @pentschev

quasiben:
gpuCI merge

quasiben:
@gpucibot merge

PR #697, opened 2021-03-18
Add new text file descriptor helper functions

pentschev:
This PR helps deduplicate some code on handling file descriptors to get UCP info that's only available as text. On a future PR these functions will be used more as we introduce support to get also context and worker text information.


quasiben (approved):
👍

PR #704, opened 2021-04-09
Memory allocation/registration and rkey management

MattBBaker:
Filling out more of the low level UCX API. This one covers the mapping of existing memory/allocation of new memory registered to a UCX context and packing an Rkey that will, after a future PR adds RMA/AMO operations, allow for one sided operations against this memory.

### DIFF
--- a/ucp/_libs/tests/test_mem.py
+++ b/ucp/_libs/tests/test_mem.py
@@ -1,7 +1,8 @@
-import pytest
 import array
 import io
 import mmap
+
+import pytest
 
 import ucp._libs.ucx_api as ucx_api
 
--- a/ucp/_libs/tests/test_mem.py
+++ b/ucp/_libs/tests/test_mem.py
@@ -51,3 +51,13 @@ def test_ctx_map(buffer):
     mem = ctx.map(buffer)
     rkey = mem.pack_rkey()
     assert rkey is not None
+
+
+def test_rkey_unpack():
+    ctx = ucx_api.UCXContext({})
+    mem = ucx_api.UCXMemoryHandle.alloc(ctx, 1024)
+    packed_rkey = mem.pack_rkey()
+    worker = ucx_api.UCXWorker(ctx)
+    ep = worker.ep_create_from_worker_address(worker.get_address(), ctx)
+    rkey = ep.unpack_rkey(packed_rkey)
+    assert rkey is not None
--- a/ucp/_libs/tests/test_mem.py
+++ b/ucp/_libs/tests/test_mem.py
@@ -1,0 +1,52 @@
+import pytest
+import array
+import io
+import mmap
+
+import ucp._libs.ucx_api as ucx_api
+
+builtin_buffers = [
+    b"",
+    b"abcd",
+    array.array("i", []),
+    array.array("i", [0, 1, 2, 3]),
+    array.array("I", [0, 1, 2, 3]),
+    array.array("f", []),
+    array.array("f", [0, 1, 2, 3]),
+    array.array("d", [0, 1, 2, 3]),
+    memoryview(array.array("B", [0, 1, 2, 3, 4, 5])).cast("B", (3, 2)),
+    memoryview(b"abcd"),
+    memoryview(bytearray(b"abcd")),
+    io.BytesIO(b"abcd").getbuffer(),
+    mmap.mmap(-1, 5),
+]
+
+
+def test_alloc():
+    ctx = ucx_api.UCXContext({})
+    mem = ucx_api.UCXMemoryHandle.alloc(ctx, 1024)
+    rkey = mem.pack_rkey()
+    assert rkey is not None
+
+
+@pytest.mark.parametrize("buffer", builtin_buffers)
+def test_map(buffer):
+    ctx = ucx_api.UCXContext({})
+    mem = ucx_api.UCXMemoryHandle.map(ctx, buffer)
+    rkey = mem.pack_rkey()
+    assert rkey is not None
+
+
+def test_ctx_alloc():
+    ctx = ucx_api.UCXContext({})
+    mem = ctx.alloc(1024)
+    rkey = mem.pack_rkey()
+    assert rkey is not None
+
+
+@pytest.mark.parametrize("buffer", builtin_buffers)
+def test_ctx_map(buffer):
+    ctx = ucx_api.UCXContext({})
+    mem = ctx.map(buffer)
+    rkey = mem.pack_rkey()
+    assert rkey is not None
--- a/ucp/_libs/tests/test_mem.py
+++ b/ucp/_libs/tests/test_mem.py
@@ -4,7 +4,7 @@ import mmap
 
 import pytest
 
-import ucp._libs.ucx_api as ucx_api
+from ucp._libs import ucx_api
 
 builtin_buffers = [
     b"",
--- a/ucp/_libs/ucx_api.pyx
+++ b/ucp/_libs/ucx_api.pyx
@@ -468,8 +468,8 @@ cdef class UCXMemoryHandle(UCXObject):
     @classmethod
     def alloc(cls, ctx, size):
         """ Allocate a new pool of registered memory. This memory can be used for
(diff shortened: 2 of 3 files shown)


pentschev:
Thanks @MattBBaker , it seems that there are some style issues. You'll need to run use `isort`, `black`, `flake8`, `flake --config=.flake8.cython` (in that order) to fix them, and also you'll need to use the exact versions listed in https://github.com/rapidsai/integration/blob/branch-0.19/conda/recipes/versions.yaml .

MattBBaker:
Oops, forgot about isort and cython. Though it looks like isort didn't hit the test directory for some reason, hmm...

pentschev:
I've just updated the pre-commit hook script in https://github.com/rapidsai/ucx-py/pull/705 , in case you prefer to use that instead.

MattBBaker:
I will have to look into the pre-commit hooks. Also, should I be targeting 0.19 or 0.20? I may just close this PR and mess with things a bit before reopening on a new commit

pentschev:
You should target 0.20. I don't see a reason to close it, if you want to continue working on it with this PR, that should be fine. When we merge we also squash the PR, so it won't create lots of commits anyway if that's your concern.

MattBBaker:
That awkward moment when you expected the 'git pull' to be a one liner, but then it was a lot more. Then that moment when you realized you did a merge on the github IU so you just got fast forwarded. Anyone want to have a look, @pentschev @jakirkham @madsbk

ucp/_libs/ucx_api.pyx
```diff
@@ -326,6 +332,169 @@ cdef class UCXContext(UCXObject):
...
+
+    @property
+    def key(self):
+        return int(<uintptr_t><void*>self._key)
+
+    @property
+    def length(self):
+        return int(self._length)
+
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
```
madsbk, line 378: The implementation of another `__getbuffer__` exist in `UCXAddress.__getbuffer__`, it would be good if we could implement it once in an utility function.

ucp/_libs/ucx_api.pyx
```diff
@@ -326,6 +332,169 @@ cdef class UCXContext(UCXObject):
...
+
+def _ucx_mem_handle_finalizer(uintptr_t handle_as_int, UCXContext ctx):
+    assert ctx.initialized
+    cdef ucp_mem_h handle = <ucp_mem_h>handle_as_int
+    cdef ucs_status_t status
+    status = ucp_mem_unmap(ctx._handle, handle)
+    assert_ucs_status(status)
+
+
+cdef class UCXMemoryHandle(UCXObject):
```
madsbk, line 417: I would be good with a description of what this class represent and its parameters.

ucp/_libs/ucx_api.pyx
```diff
@@ -326,6 +332,169 @@ cdef class UCXContext(UCXObject):
...
         return decode_text_fd(text_fd)
 
+    def map(self, mem):
+        return UCXMemoryHandle.map(self, mem)
+
+    def alloc(self, size):
+        return UCXMemoryHandle.alloc(self, size)
+
+
+cdef class PackedRemoteKey:
```
madsbk, line 342: I would be good with a description of what this class represent and its parameters.

ucp/_libs/tests/test_mem.py
```diff
@@ -0,0 +1,53 @@
...
+    rkey = mem.pack_rkey()
+    assert rkey is not None
+
+
+@pytest.mark.parametrize("buffer", builtin_buffers)
+def test_ctx_map(buffer):
+    ctx = ucx_api.UCXContext({})
+    mem = ctx.map(buffer)
+    rkey = mem.pack_rkey()
+    assert rkey is not None
```
madsbk, line 53: I think it would be good with a test that pickles the `rkey` and send it to another peer.

ucp/_libs/ucx_api.pyx
```diff
@@ -326,6 +332,169 @@ cdef class UCXContext(UCXObject):
...
+def _ucx_mem_handle_finalizer(uintptr_t handle_as_int, UCXContext ctx):
+    assert ctx.initialized
+    cdef ucp_mem_h handle = <ucp_mem_h>handle_as_int
+    cdef ucs_status_t status
+    status = ucp_mem_unmap(ctx._handle, handle)
+    assert_ucs_status(status)
+
+
+cdef class UCXMemoryHandle(UCXObject):
+    cdef ucp_mem_h _memh
```
madsbk, line 418: I suggest that you rename `memh` to `mem_handle` or `memory_handle` everywhere.

madsbk (changes requested):
Thanks @MattBBaker, LGTM. I only have some minor requests.

ucp/_libs/ucx_api.pyx
```diff
@@ -326,6 +332,169 @@ cdef class UCXContext(UCXObject):
...
+    def from_memh(self, UCXMemoryHandle mem):
+        cdef void *key
+        cdef size_t len
+        cdef ucs_status_t status
+        status = ucp_rkey_pack(mem._context._handle, mem._memh, &key, &len)
+        assert_ucs_status(status)
+        return PackedRemoteKey(<uintptr_t>key, len)
+
+    def __dealloc__(self):
+        ucp_rkey_buffer_release(self._key)
```
pentschev, line 368: The [UCX docs]() say:

> This routine releases the buffer that was allocated using ucp_rkey_pack().

This seems to me like it will deallocate the value of `self._key`, but we still need to `free(self._key)`, no?

MattBBaker: Ooops, forgot I was memcopying in the constructor. Yes I should be calling free instead.

ucp/_libs/tests/test_mem.py
MattBBaker: My old RMA branch has some code like this, I'll see what I can grab from that

MattBBaker: I think this is the only comment that I didn't completely address. I added a test that will unpack the rkey locally so we know that the rkey is valid and will unpack correctly. I plan on extending `test_peer_send_recv.py` in the next update with tests to the added puts and gets as well as sending the `rkey`.

madsbk (approved):
LGTM, thanks @MattBBaker

ucp/_libs/ucx_api.pyx
```diff
@@ -326,6 +355,203 @@ cdef class UCXContext(UCXObject):
...
+            _ucx_mem_handle_finalizer,
+            int(<uintptr_t>self._mem_handle),
+            self._context
+        )
+        ctx.add_child(self)
+
+    @classmethod
+    def alloc(cls, ctx, size):
+        """ Allocate a new pool of registered memory. This memory can be used for
+            RMA and AMO operations. This memory should no be accessed from outside these
```
pentschev, line 471:
```suggestion
            RMA and AMO operations. This memory should not be accessed from outside these
```

pentschev (approved):
Just a typo fix suggestion, otherwise LGTM as well. Thanks @MattBBaker !

pentschev:
I took the liberty to commit the typo fix myself, we can merge it as soon as CI passes then.

ucp/_libs/ucx_api.pyx
```diff
@@ -326,6 +355,203 @@ cdef class UCXContext(UCXObject):
...
+            int(<uintptr_t>self._mem_handle),
+            self._context
+        )
+        ctx.add_child(self)
+
+    @classmethod
+    def alloc(cls, ctx, size):
+        """ Allocate a new pool of registered memory. This memory can be used for
+            RMA and AMO operations. This memory should not be accessed from outside these
+            operations.
```
pentschev, line 472:
```suggestion
            RMA and AMO operations. This memory should not be accessed from outside
            these operations.
```

pentschev (approved):

pentschev:
This is in, thanks @MattBBaker ! 😄

PR #701, opened 2021-03-26
Add support for UCX AM (Active Messages)

pentschev:
This PR adds initial support for the Active Message UCX API. There are currently some limitations:

- Receiving rendezvous messages is not supported yet;
- Only host messages are supported.

The limitations above are intended to avoid this PR getting even longer and will be addressed in a follow-up PR.

PR #707, opened 2021-04-12
Add Active Message support for receiving rendezvous messages

pentschev:
Extends Active Message (AM) support for receiving rendezvous messages. All CUDA messages are rendezvous messages, thus this is required for CUDA as well which is also part of this PR.

This builds on top/extends #701 .


ucp/_libs/ucx_api.pyx
```diff
@@ -372,17 +378,227 @@ def _ucx_worker_handle_finalizer(
...
+                    if cb_args is None:
+                        cb_args = ()
+                    cb_kwargs = req_info["cb_kwargs"]
+                    if cb_kwargs is None:
+                        cb_kwargs = {}
+                    cb_func(cb_args[0], exception, **cb_kwargs)
+            finally:
+                req.close()
+
+    cdef ucs_status_t _am_recv_callback(
```
jakirkham, line 439: We may want to add some directives here. In particular `boundscheck` to avoid checking if indexing `buf[0]` will raise an error. Disabling some of the other checks may make sense as well

https://github.com/rapidsai/ucx-py/blob/0fca081f960aa0372e1d2c6414c8c74ecc3a9b11/ucp/_libs/arr.pyx#L203-L206

pentschev:
rerun tests

ucp/_libs/ucx_api.pyx
pentschev: This is done now.

pentschev:
@madsbk @quasiben this is ready now, it includes support for both host and CUDA messages, but more tests and better integration with `Array` class are needed and will be worked on future PR(s). Could one of you review it when you have a chance?

ucp/_libs/ucx_api.pyx
```diff
@@ -373,6 +380,67 @@ def _ucx_worker_handle_finalizer(
...
+        assert user_data != NULL
+
+        with log_errors():
+            req = UCXRequest(<uintptr_t><void*> request)
+            assert not req.closed()
+            req_info = <dict>req._handle.info
+            req_info["status"] = "finished"
+
+            if "cb_func" not in req_info:
+                # This callback function was called before _am_recv_callback() returned
```
madsbk, line 412:
```suggestion
                # This callback function was called before _am_recv_completed_callback() returned
```

ucp/_libs/ucx_api.pyx
```diff
@@ -495,6 +638,36 @@ cdef class UCXWorker(UCXObject):
         )
         context.add_child(self)
 
+    IF CY_UCP_AM_SUPPORTED:
```
madsbk, line 641: I suggest that you put `IF CY_UCP_AM_SUPPORTED:` inside `register_am_allocator()` and raise an error if AM isn't supported.

ucp/_libs/ucx_api.pyx
```diff
@@ -373,6 +380,67 @@ def _ucx_worker_handle_finalizer(
 
 
 IF CY_UCP_AM_SUPPORTED:
+    cdef void _am_recv_completed_callback(
+        void *request,
+        ucs_status_t status,
+        size_t length,
+        void *user_data
```
madsbk, line 387: `void *user_data` doesn't seems to be used?

ucp/core.py
```diff
@@ -874,6 +883,10 @@ def get_config():
         return _get_ctx().get_config()
 
 
+def register_am_allocator(allocator, allocator_type):
```
madsbk, line 886: Would be good with a doc string here

madsbk (changes requested):
Nice work @pentschev, only have some minor suggestions.

ucp/_libs/ucx_api.pyx
pentschev: Done.

ucp/_libs/ucx_api.pyx
pentschev: Good idea, done in https://github.com/rapidsai/ucx-py/pull/707/commits/6ec6b447071f24de51593fba988d3dbca1ab33ff .

ucp/_libs/ucx_api.pyx
pentschev: Yes, this is the data buffer where the message is received. This is part of the [ucp_am_recv_data_nbx_callback_t](http://openucx.github.io/ucx/api/latest/html/group___u_c_p___c_o_m_m.html#ga242399982563973351591516039dd74b) prototype, but the copy is handled automatically by UCX, so we don't need to do anything with it.

ucp/core.py
pentschev: Good catch, added one in https://github.com/rapidsai/ucx-py/pull/707/commits/dcf0a8e7912b535791badb5dc594205ff9b97855 .

pentschev:
Thanks @madsbk for reviewing, I think I addresses all questions/suggestions, could you take another look when you have a chance?

ucp/_libs/ucx_api.pyx
madsbk: Yeah, but you are assigning it in `_am_recv_callback()`:
```
            request_param.user_data = <void *>buf
            status = ucp_am_recv_data_nbx(
                worker._handle, data, buf_ptr, length, &request_param
            )
```
and checking that it isn't `NULL` in `_am_recv_completed_callback()`:
```
assert user_data != NULL
```

I know it is a minor thing but it is just a bit confusing :)

pentschev: Yes, I agree, it took me a while to understand that too. I see two options:

1. Don't assign `user_data` and lose the ability to log the data pointer; or
2. Keep `user_data` but remove the assertion, thus (hopefully) making it clearer it's only used for debugging.

I personally would prefer 2, do you have a preference here, or maybe even another suggestion on how to deal with this?

madsbk: >     1. Don't assign `user_data` and lose the ability to log the data pointer;

Isn't it possible to use `req_info["cb_args"]` for logging, which is also assigned the buffer pointer in `_am_recv_callback`?
```
req_info["cb_args"] = (buf, )
```
If not, maybe just remove `req_info["cb_args"]` and rely on `user_data` to get the buffer pointer?

pentschev: Yes, but that requires other checks before we can log anything, so if anything fails we lose the ability to do so anyway. I guess this is reasonable nevertheless, I'll do that.

ucp/_libs/ucx_api.pyx
pentschev: I actually realized now that it was correct before. The function here is `_am_recv_completed_callback`, and it needs to be called after `_am_recv_callback` returns.

ucp/_libs/ucx_api.pyx
pentschev: This is done now.

madsbk (approved):
Looks good @pentschev, thanks!

pentschev:
Thanks @madsbk for the review! 😄

PR #716, opened 2021-04-26
Split transfer and UCXWorker callback functions into multiple files

pentschev:
This PR split transfer functions into multiple files for AM, Stream and TAG APIs and a common file. Also moves the UCXWorker callback functions to new file.


ucp/_libs/transfer_common.pyx
```diff
@@ -0,0 +1,54 @@
...
+# cython: language_level=3
+
+from libc.stdint cimport uintptr_t
+
+from .ucx_api_dep cimport *
+
+from ..exceptions import UCXCanceled, UCXError, log_errors
+
+
+cdef void _send_callback(void *request, ucs_status_t status):
```
madsbk, line 14: I would be good a comment saying it is used by `stream_send_nb` and `tag_send_nb`

madsbk (approved):
LGTM, only a minor suggestion

ucp/_libs/transfer_common.pyx
pentschev: Done in https://github.com/rapidsai/ucx-py/pull/716/commits/d45563064a5bd555c080a7e7a54ad8376e1e49df . Also `UCXEndpoint.flush` and `UCXWorker.flush` depend on it as well.

pentschev:
Thanks for the review @madsbk !

PR #717, opened 2021-04-26
Handling UCS_INPROGRESS in assert_ucs_status()

MattBBaker:
While working on the RMA branch I ran into this interesting behavior. As it stands now the `assert_ucs_status()` function will treat `UCS_INPROGRESS` as an error and raise an exception. I looked through the code and don't see how this change would affect anything outside of the RMA implicit interface, but I wanted to break this change out into it's own one liner to see if any tests have a problem with this change

### DIFF
--- a/ucp/_libs/utils.pyx
+++ b/ucp/_libs/utils.pyx
@@ -90,7 +90,7 @@ cdef get_ucx_object(Py_buffer *buffer, int flags,
 
 cdef void assert_ucs_status(ucs_status_t status, str msg_context=None) except *:
     cdef str msg, ucs_status
-    if status != UCS_OK:
+    if not (status == UCS_OK or status == UCS_INPROGRESS):
         ucs_status = ucs_status_string(status).decode("utf-8")
         if msg_context is not None:
             msg = f"[{msg_context}] {ucs_status}"


pentschev:
Interesting catch @MattBBaker . I haven't tested this patch yet, but are you sure this is ok for the places where we actually call `assert_ucs_status`? IOW, won't they every return `UCS_INPROGRESS` that this would now silently ignore?

MattBBaker:
I think that the `put/get_nbi` functions are the only UCP functions that ever return `UCS_INPROGRESS`. I'll double check though.

MattBBaker:
@pentschev Looks like this has been going on for over a day

pentschev:
rerun tests

pentschev:
Yeah, that's strange. I triggered a rerun, let's see what happens now.

pentschev:
> I think that the `put/get_nbi` functions are the only UCP functions that ever return `UCS_INPROGRESS`. I'll double check though.

Btw, did you find anything besides `put`/`get_nbi`? TBH, I'm not sure based on what can we expect functions to ever return `UCS_INPROGRESS`.

MattBBaker:
Yea, looking through the documentation none if really defines if something may or may not return `UCS_INROGRESS`. I wouldn't expect something like context or worker create to return in progress. I'll have to poke Yossi.

pentschev:
> I'll have to poke Yossi.

I poked UCX devs on our internal Slack channel, let's see what they say, I'll merge this once I hear back from them.

pentschev:
@MattBBaker Yossi said functions should state in their docs whether it may or may not return `UCS_INPROGRESS`. Do you know which function was triggering that issue here?

MattBBaker:
It is `ucp_put_nbi()`. I double checked in the docs and it does state it can return UCS_INPROGRESS in the description of the function, though it doesn't state that in the `Returns` section of the docs, where I was looking.

MattBBaker:
Also in the communications routine section it's just the RMA functions that use it. And some of the nbr functions, but I don't think that the python interface exposes the nbr functions

pentschev:
I think maybe what we should do here is something different. What I imagine is you have to manage the `UCS_INPROGRESS` on a per-case basis, e.g.:

```python
status = ucp_put_nbi(...)
while status == UCS_INPROGRESS:
    worker.progress()
assert_ucs_status(status)
```

In the case above, we guarantee that `UCS_INPROGRESS` will not be passed to `assert_ucs_status`. If that happens to be passed to `assert_ucs_status` we should indeed raise an error because it cannot handle that case. If raising an error alone is a wormhole, we could potentially print that to UCX-Py logging as well to help tracking that down:

```python
cdef void assert_ucs_status(ucs_status_t status, str msg_context=None) except *:
...
    if status == UCS_INPROGRESS:
        logger.debug("assert_ucs_status called with `UCS_INPROGRESS` status")
```

I fear that the change as it is here currently is a potential case to just sweep a different issue under the rug, namely `UCS_INPROGRESS` being ignored.

MattBBaker:
You're example won't work. Remember that you are getting an immediate value, so status will never change off of UCS_INPROGRESS during progress. The only way to know if it completes is to call flush. An example I have in the extended `test_peer_send_recv.py` test I have for my RMA branch:
```
        right_ep, right_mem = _rma_setup(worker, right_address, right_prkey, right_base)
        right_msg = bytearray(msg_size)
        right_mem.get_nbi(Array(right_msg))

        left_ep, left_mem = _rma_setup(worker, right_address, right_prkey, right_base)
        left_msg = bytearray(msg_size)
        left_mem.get_nbi(Array(left_msg))

        blocking_flush(worker, worker)
        assert left_msg == bytes(repeat(left_rank, msg_size))
        assert right_msg == bytes(repeat(right_rank, msg_size))

```
Since there is no request structure the only thing I can know for sure is that the bytes will only be in `right_msg` and `left_msg` for sure after the flush. Right now the way it works in the `get_nbi()` function is that the return status will represent if the operation completed immediately, returning True or False:
```
status = ucp_put_nbi()
assert_ucs_status(status)
if status == UCS_OK:
    return True
else:
    return False
```
That way if you want to know if you can skip the flush you can check the return value. Otherwise you can arrange all of your puts and gets in a row and flush to complete all of them. And errors still propagate as exceptions.

pentschev:
Yes, you're right. I mixed that with `ucs_status_ptr_t` that can be checked with `ucp_request_check_status`, as in https://github.com/rapidsai/ucx-py/blob/51d11c6a0b6f1294c2f74d318386de9523ecb0b3/ucp/_libs/ucx_endpoint.pyx#L41-L42

For now I think the PR is good as is, but I'm still wondering if we should treat `UCS_INPROGRESS` differently, as to avoid confusion with "ready to consume". We can discuss further on that in a separate issue, I'll file one.

pentschev (approved):

pentschev:
I filed https://github.com/rapidsai/ucx-py/issues/722 so we can keep track of it. Thanks @MattBBaker for finding this and the PR!

PR #719, opened 2021-04-27
Add GitHub Actions workflow for docs build

charlesbluca:
Adds a GitHub Actions workflow triggered on merges to `branch-0.20` to build and publish the UCX-Py documentation through GitHub Pages. This handles the issue of libnuma dependencies which currently make this process difficult with RTD.

cc @quasiben

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,5 +21,6 @@ jobs:
           auto-activate-base: false
       - name: Build documentation
         run: |
+          conda activate ucx_dev
           cd docs
           make html
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,14 +32,16 @@ jobs:
           environment-file: latest/conda/environments/builddocs_py37.yml
           auto-activate-base: false
       - name: Build documentation
+        working-directory: latest/docs
+        run: make html
+      - name: Copy docs to gh-pages
         run: |
-          cd latest/docs
-          make html
-      - name: Commit changes to docs
-        run: |
-          COMMIT_SHA=$(echo "${COMMIT_SHA}" | cut -c 1-7)
           rm -rf gh-pages/*
           cp -r latest/docs/build/html/* gh-pages/
+      - name: Commit changes to docs
+        working-directory: gh-pages
+        run: |
+          COMMIT_SHA=$(echo "${COMMIT_SHA}" | cut -c 1-7)
           git config user.name "GitHub Actions"
           git config user.email "action@github.com"
           git add --all
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,9 +15,5 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           activate-environment: ucx_dev
-          environment-file: ../conda/environments/builddocs_py37.yml
+          environment-file: conda/environments/builddocs_py37.yml
           auto-activate-base: false
-      - name: Build documentation
-        uses: ammaraskar/sphinx-action@0.4
-        with:
-          docs-folder: "docs/"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Commit changes to docs
         run: |
           git clone https://github.com/charlesbluca/ucx-py.git --branch gh-pages --single-branch gh-pages
-          cp -r build/html/* gh-pages/
+          cp -r docs/build/html/* gh-pages/
           cd gh-pages
           touch .nojekyll
           git config --local user.email "action@github.com"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,23 @@
+on:
+  push:
+    branches: ["gha-docs-build"]
+
+jobs:
+  build:
+    name: Build Sphinx documentation
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Install libnuma
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libnuma-dev
+      - name: Create docs environment
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: ucx_dev
+          environment-file: conda/environments/builddocs_py37.yml
+          auto-activate-base: false
+      - name: Build documentation
+        uses: ammaraskar/sphinx-action@0.4
+        with:
+          docs-folder: "docs/"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,6 +1,8 @@
+name: UCX-Py documentation build
+
 on:
   push:
-    branches: ["gha-docs-build"]
+    branches: ["branch-0.20"]
 
 jobs:
   build:
@@ -29,7 +31,7 @@ jobs:
           make html
       - name: Commit changes to docs
         run: |
-          git clone https://github.com/charlesbluca/ucx-py.git --branch gh-pages --single-branch gh-pages
+          git clone https://github.com/rapidsai/ucx-py.git --branch gh-pages --single-branch gh-pages
           cp -r docs/build/html/* gh-pages/
           cd gh-pages
           touch .nojekyll
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -11,9 +11,11 @@ jobs:
         run: |
           sudo apt-get update -y
           sudo apt-get install -y libnuma-dev
-      - name: Create docs environment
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
+name: UCX-Py documentation build
+
+on:
+  push:
+    branches: ["branch-0.20"]
```
charlesbluca, line 5: Unfortunately, GitHub Actions does not have a way to dynamically select the default branch of a repo, so we would need to change this with every new release cycle.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,47 @@
...
+          environment-file: conda/environments/builddocs_py37.yml
+          auto-activate-base: false
+      - name: Build documentation
+        run: |
+          conda activate ucx_dev
+          cd docs
+          make html
+      - name: Commit changes to docs
+        run: |
+          git clone https://github.com/rapidsai/ucx-py.git --branch gh-pages --single-branch gh-pages
```
charlesbluca, line 34: The build will fail here until we add a (preferably empty) `gh-pages` branch to this repo.

charlesbluca (commented):
Some comments; also I'm pretty sure we should be able to speed up builds by caching the environment and using Mamba ([here's an xarray workflow doing that](https://github.com/pydata/xarray/blob/0021cdab91f7466f4be0fb32dae92bf3f8290e19/.github/workflows/ci.yaml#L54-L73)), but I'm not exactly sure if/how we could cache the libnuma dependency.

pentschev:
I would prefer that we keep RTD for UCX-Py, the reason being that [UCX uses RTD](https://openucx.readthedocs.io/en/master/) and our intent is to upstream UCX-Py to their repo/org in the next few months.

charlesbluca:
I'm okay with that - in that case, we might want to pursue either:

- a form of these docs that doesn't require a UCX installation
- the possibility of packaging UCX without libnuma dependencies (like the version used [here](https://github.com/quasiben/ucx-py/blob/fix-rtd/conda/environments/builddocs_py37.yml))

In either case, I saw a GitHub Pages site as a short term solution to keep the docs up to date and centralized in this repo until building them through RTD is possible.

pentschev:
I'm a bit confused, what exactly is the problem today with RTD? We do have the docs up, so I'm guessing that's something to do with automatic updates, is that right?

charlesbluca:
From what @quasiben told me, the problem is that the [docs env](https://github.com/rapidsai/ucx-py/blob/branch-0.20/conda/environments/builddocs_py37.yml) in this repo contains UCX, which depends on libnuma, which I don't think can be installed via Conda. Since RTD doesn't allow you to pre-configure the machine docs are being built on, this means that trying to build the docs from this repo will always result in failure.

The docs that are currently published are being built from:

- https://github.com/quasiben/ucx-py/tree/fix-rtd

Which instead uses a specific version of [UCX with no libnuma dependencies](https://github.com/quasiben/ucx-py/blob/eb6bc0a4a5af7eeab0c85eaab35db72116e36fa1/conda/environments/builddocs_py37.yml#L20) which is hosted from a channel I assume @quasiben created. The docs are then periodically updated whenever he merges upstream changes into that branch.

.github/workflows/docs.yml
ajschmidt8: For this, you could adapt the lines of code below from `cudf` for `ucx-py`. That script gets run when we create new development branches and the `sed` commands can update this line for you automatically.

https://github.com/rapidsai/cudf/blob/branch-0.20/ci/release/update-version.sh#L44-L75

.github/workflows/docs.yml
ajschmidt8: I've just created an empty `gh-pages` branch for you. https://github.com/rapidsai/ucx-py/tree/gh-pages

ajschmidt8 (approved):
Approved. For reference, I did something similiar for the `node-rapids` repo. You can check out my implementation below. In particular, I thought you might be able to use [these two `checkout` actions](https://github.com/rapidsai/node-rapids/blob/main/.github/workflows/docs.yaml#L19-L27) to simplify your second `git clone`.


https://github.com/rapidsai/node-rapids/blob/main/.github/workflows/docs.yaml

pentschev:
Thanks for the detailed explanation @charlesbluca . I understand the problem we're trying to solve now, but I'm still unclear on whether this will resolve the problem or just push it down the road. As we intend to move everything here into OpenUCX, we're probably going to need a way of updating RTD, so I don't think the solution here is a proper long-term one.

charlesbluca:
Agreed - I'm hoping that we might be able to find a way to package UCX without libnuma dependencies so that we can eventually move back to RTD, but until then this just seemed like a better alternative to the current setup.

pentschev:
With this, will we still be able to use RTD or does this mean a complete change to GH Pages now that we'll eventually need to move back again to RTD?

charlesbluca:
Thanks for the review @ajschmidt8! Yeah, those `node-rapids` checkout/commit steps definitely look cleaner, I'll adjust my workflow to use that approach then hopefully this should be good to merge.

charlesbluca:
@pentschev we should be able to keep both simultaneously - I was considering having the RTD copy of the docs redirect to the GH pages version, but maybe a different domain name would be better.

pentschev:
If we can keep both, then I think it's ok, but otherwise ping-ponging users back and forth is not really a great approach and we should first consider what other options we have.

charlesbluca:
@gpucibot merge

charlesbluca:
@gpucibot merge

charlesbluca:
Looks like I can't merge this - @pentschev @ajschmidt8 would one of you mind merging this?

pentschev:
> Looks like I can't merge this - @pentschev @ajschmidt8 would one of you mind merging this?

Can you first confirm that RTD will still be available with this change?

pentschev:
Actually, RTD should not just be available, but kept up-to-date, is that the case?

charlesbluca:
Sure! Since the RTD project is based on @quasiben's repo:

image

That should be unaffected by this merge and the RTD site should remain up. However, it would not be up-to-date - if we wanted the RTD site to mirror this one, we would need @quasiben's repo to regularly pull upstream changes and patch them with the no libnuma version of UCX so that the builds don't fail; that could probably be done with a cron job.

Another option that I'm not entirely sure would work is having the RTD site silently redirect to the GitHub pages one, so that they would always be identical. I know RTD supports redirects, though I'm not entirely sure if this is how they work.

pentschev:
Yes, but then we would be duplicating documentation and they will be out-of-sync. That's precisely what I don't want to do. Given we'll soon upstream this repo and will need to update RTD then, I don't really think going to GH Pages is something we should do.

@quasiben can we please hold on to this change and sync first on how we really want/need to handle docs?

charlesbluca:
After a chat with @pentschev and @quasiben, we have a clear roadmap of how to continue publishing the docs through RTD:

- Have a GitHub Actions workflow triggered on merges to the latest branch that checks it out, patches it to use a no libnuma version of UCX, and pushes it to a RTD specific branch (something like `readthedocs`)
- Change RTD builds to be triggered by pushes to `rapidsai/ucx-py@readthedocs` now

In that case, we won't need this workflow anymore, or the empty `gh-pages` branch (though we will need one for RTD later on).

pentschev:
> * Have a GitHub Actions workflow triggered on merges to the latest branch that checks it out, patches it to use a no libnuma version of UCX, and pushes it to a RTD specific branch (something like `readthedocs`)

Just a brief clarification here, I don't think we need to patch anything. The new `readthedocs` branch would be a fork of the current default `branch-0.20` with the necessary [RTD build file](https://github.com/quasiben/ucx-py/blob/fix-rtd/conda/environments/builddocs_py37.yml#L20). Once a PR is merged into `branch-0.20`, GH Actions would merge `branch-0.20` into the `readthedocs` branch as well, so that RTD is always up-to-date.

charlesbluca:
Thanks for the clarification, @pentschev - opened up a PR doing this #721

ajschmidt8:
> After a chat with @pentschev and @quasiben, we have a clear roadmap of how to continue publishing the docs through RTD:
> 
> * Have a GitHub Actions workflow triggered on merges to the latest branch that checks it out, patches it to use a no libnuma version of UCX, and pushes it to a RTD specific branch (something like `readthedocs`)
> * Change RTD builds to be triggered by pushes to `rapidsai/ucx-py@readthedocs` now
> 
> In that case, we won't need this workflow anymore, or the empty `gh-pages` branch (though we will need one for RTD later on).

can this PR be closed then?

charlesbluca:
> can this PR be closed then?

Yes. I think the resolution here is that we actually don't need either workflow, as RTD builds are working fine with the latest version of UCX.

PR #720, opened 2021-04-27
Fix use after free bug in PackedRemoteKey

MattBBaker:
This one was fun to track down. Small fix so that the `from_mem_handle` function path returns the correct object, instead of returning a new copy with a 'use after free' bug.

### DIFF
--- a/ucp/_libs/packed_remote_key.pyx
+++ b/ucp/_libs/packed_remote_key.pyx
@@ -56,7 +56,7 @@ cdef class PackedRemoteKey:
         packed_key = PackedRemoteKey(<uintptr_t>key, len)
         ucp_rkey_buffer_release(key)
         assert_ucs_status(status)
-        return PackedRemoteKey(<uintptr_t>key, len)
+        return packed_key
 
     def __dealloc__(self):
         free(self._key)


ucp/_libs/packed_remote_key.pyx
```diff
@@ -56,7 +56,7 @@ cdef class PackedRemoteKey:
         packed_key = PackedRemoteKey(<uintptr_t>key, len)
         ucp_rkey_buffer_release(key)
         assert_ucs_status(status)
-        return PackedRemoteKey(<uintptr_t>key, len)
+        return packed_key
```
pentschev, line 59: Good catch, but what are we actually returning here? It seems like `key` is released by `ucp_rkey_buffer_release`, so I'm not sure what exactly is that `packed_key` contains.

MattBBaker: It's a new instance of the `PackedRemoteKey` class. The class makes it's own copy of the key since it can either from from UCX (thus needs to be freed with UCX functions) or from the external runtime (thus needs to be managed and freed as a bytes/buffer routines). Though now that I'm thinking about the problem and scaling it, a better solution would be to just do the copy/buffer stuff in the memory handle case and then not make a copy in the buffer case. Hmm... add one to the todo file.

pentschev: Makes sense, it's indeed a bit misleading to read just this function without looking at the constructor, but for now this will do. Thanks Matt!

pentschev (approved):

PR #721, opened 2021-04-29
Add workflow to fix RTD build failures

charlesbluca:
As discussed in #719, the RTD builds for this repo are failing due to a libnuma dependency. After talking with @pentschev, we agree that a GitHub Pages site wouldn't work well for this repo (since it is eventually going to be migrated to [OpenUCX](https://github.com/openucx/), who use RTD for their docs. Instead, we will opt to have a branch in this repo specifically for RTD builds, using a version of UCX with no libnuma dependencies.

This workflow will merge the latest changes to that RTD branch whenever the latest development branch is updated.

cc @pentschev


.github/workflows/rtd.yml
```diff
@@ -0,0 +1,21 @@
+name: Update readthedocs branch
+
+on:
+  push:
+    branches: ["branch-0.20"]
```
charlesbluca, line 5: Like before, noting that this needs to be changed with every release, which can be done with a release script like this:

- https://github.com/rapidsai/cudf/blob/branch-0.20/ci/release/update-version.sh#L44-L75

.github/workflows/rtd.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  update:
+    name: Update readthedocs branch
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Checkout readthedocs
+        uses: actions/checkout@v2.3.4
+        with:
+          ref: readthedocs
```
charlesbluca, line 15: Are we okay with this name for the branch? Regardless of the name, we'll need to make a new branch for this that uses the no libnuma version of UCX, as shown in:

- https://github.com/quasiben/ucx-py/blob/fix-rtd/conda/environments/builddocs_py37.yml

charlesbluca (commented):
Couple things to note:

.github/workflows/rtd.yml
pentschev: Sounds good, I don't have a preference, but for now we can do this manually and later have a script. Either way is fine by me.

.github/workflows/rtd.yml
pentschev: I'm fine with that name. Do you have permission to push a new branch? If not I can try it (not sure if I have permission as well).

charlesbluca: I don't have permission - is this limited to `ucxpy-python-codeowners` or some other group?

ajschmidt8: I see the new `readthedocs` branch, but it's not an empty branch. Is that intended? If not, I can re-create it as an empty branch.

ajschmidt8 (approved):

.github/workflows/rtd.yml
pentschev: That's intended. The `readthedocs` branch is a fork of `branch-0.20` with additional changes to https://github.com/rapidsai/ucx-py/blob/readthedocs/conda/environments/builddocs_py37.yml, specifically the changes we need are in https://github.com/rapidsai/ucx-py/commit/8535c9c4e4452bd51e3b57f1ccd72fdf41d936f2 , which gets a UCX build that's not dependent on `libnuma`. From this point on, every time a PR is merged into `branch-0.20`, the changes from the present PR will be merged into that and RTD rebuild and published.

pentschev (approved):
Thanks @charlesbluca !

PR #724, opened 2021-04-29
Fix wireup assertion in AM core test

pentschev:
This PR fixes an issue when asserting the wireup message and NumPy is used as a host allocator, as `numpy.ndarray` can't be directly compared to `bytearray`.


quasiben (approved):
Thanks @pentschev !

pentschev:
Thanks @quasiben for review and merging!

PR #725, opened 2021-04-29
Fix RTD workflow

charlesbluca:
Forgot you need to set a Git username and email in order to make the commit and push it; this should fix build failures.


pentschev (approved):
LGTM, thanks @charlesbluca .

We'll still need an approval from @rapidsai/ops-codeowners though.

ajschmidt8 (approved):

PR #729, opened 2021-05-03
Close endpoints in ucp.core tests

pentschev:
This is a preparation for endpoint error handling. UCX-Py currently try as best as possible to ensure endpoints are closed according to each object's lifetime. However, this can't always be ensured given the async behavior in `ucp.core`. For example, a `ucp.core.Endpoint` object does not call `abort` or `close`, specifically `close` would be desired to inform the other end that it is closing. Because that is an async function, we can't rely on it for a destructor. It's ensured that the underlying synchronous `ucp._libs.UCXEndpoint` object is destroyed when its weakref count goes to zero, and `ucp_ep_close_nb` is called. With the endpoint error handling work in https://github.com/rapidsai/ucx-py/pull/693 , we will be switching to `UCP_EP_CLOSE_MODE_FORCE` that doesn't guarantee inflight messages will be flushed before the endpoint closes, potentially leading to "Connection reset by peer" errors if the other side is waiting to read a message.

From discussions with UCX devs, we should always call `ucp_ep_close_nb` on the _listener_ side, and currently we don't do it in many of our tests, relying solely on an endpoint's lifetime management, which is a bad practice from us. In many cases, simply not calling `ep.close()` would be fine when the last message is sent by client/received by listener, but reverting the order will not. This PR intends to correct this and avoid the need to rely on the correct order to close endpoints based on message direction.

### DIFF
--- a/tests/test_send_recv_two_workers.py
+++ b/tests/test_send_recv_two_workers.py
@@ -191,6 +191,9 @@ def cupy_obj():
 )
 @pytest.mark.parametrize("comm_api", ["tag", "am"])
 def test_send_recv_cu(cuda_obj_generator, comm_api):
+    if comm_api == "am" and not ucp._libs.ucx_api.is_am_supported():
+        pytest.skip("AM only supported in UCX >= 1.11")
+
     base_env = os.environ
     env_client = base_env.copy()
     # grab first two devices
--- a/tests/test_multiple_nodes.py
+++ b/tests/test_multiple_nodes.py
@@ -25,6 +25,7 @@ async def hello(ep):
 async def server_node(ep):
     await hello(ep)
     assert isinstance(ep.ucx_info(), str)
+    await ep.close()
 
 
 async def client_node(port):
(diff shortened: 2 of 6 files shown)


pentschev:
@madsbk @quasiben could you take a look at this and comment on whether you agree with the change based on the description or you think we should do something differently?

pentschev:
One additional information, I believe this is the last fix I need to ensure all UCX-Py tests pass with/without endpoint error handling, but this is a change that may be looked at a functional change in UCX-Py, even though we [explicitly document it](https://ucx-py.readthedocs.io/en/latest/quickstart.html#process-1-server), we don't always follow it ourselves, as the tests here prove.

tests/test_send_recv_two_workers.py
```diff
@@ -191,6 +191,9 @@ def cupy_obj():
 )
 @pytest.mark.parametrize("comm_api", ["tag", "am"])
 def test_send_recv_cu(cuda_obj_generator, comm_api):
+    if comm_api == "am" and not ucp._libs.ucx_api.is_am_supported():
+        pytest.skip("AM only supported in UCX >= 1.11")
+
```
pentschev, line 196: This change is unrelated to the PR's description, I intended to open a separate PR but accidentally had it pushed here. It should be harmless but necessary, so I'll just leave it here anyway.

pentschev:
@madsbk @quasiben ping

quasiben (approved):
Thanks @pentschev

PR #730, opened 2021-05-04
Remove try/except block for Cython import in `setup.py`

charlesbluca:
It looks like Cython is required to build from source, so the setup should just fail immediately if it can't be imported.


raydouglass (approved):

pentschev (approved):
Makes sense, thanks @charlesbluca !

Issue #722, opened 2021-04-29
Return different status for `UCS_INPROGRESS` in `assert_ucs_status()`


pentschev:
`UCS_INPROGRESS` was being treated as error in `assert_ucs_status()`, which was then fixed to pass in #717 , as per @MattBBaker 's discovery. However, this can be misleading with a user understanding this could mean "ready to consume". Since `assert_ucs_status` is meant to be used with `ucs_status_t`, we cannot block until progress is completed as we would do with `ucs_status_ptr_t`. Perhaps in the future we should think of a different interface for `assert_ucs_status()` with the return status informing the user whether the operation is still in progress or not.

MattBBaker:
What about renaming it to something like `assert_ucs_ok()` and defining it as, "Returns `True` if the underlying op completed in place, `UCS_OK`. Return `False` is the operation in in progress, `UCS_INPROGRESS`. Raises an UCXException if an error is indicated, `UCS_ERR_*`. This is how I am translating it for my RMA NBI functions. That way a blocking operation can be implemented like,
```
if not rmem.put_nbi(data):
    blocking_flush(ep)
```

pentschev:
Yes, I think that's a good solution.

@madsbk what do you think of the above? Any other ideas?

madsbk:
Sounds good to be!

PR #732, opened 2021-05-06
Change ucs_assert_status() to return a bool

MattBBaker:
Implements what was discussed in #722


ucp/_libs/utils.pyx
```diff
@@ -88,15 +88,24 @@ cdef get_ucx_object(Py_buffer *buffer, int flags,
     buffer.internal = NULL
 
 
-cdef void assert_ucs_status(ucs_status_t status, str msg_context=None) except *:
+# Helper function to process ucs return codes. Returns True of the status is UCS_OK to
```
pentschev, line 91:
```suggestion
# Helper function to process ucs return codes. Returns True if the status is UCS_OK to
```

pentschev (changes requested):

pentschev (approved):
LGTM, thanks @MattBBaker . Will merge as soon as CI passes.

PR #735, opened 2021-05-21
New RemoteMemory and UCXIO classes for RMA operations

MattBBaker:
New round of UCX RMA operations. This includes a low level interface that just lifts the C api up to taking in python objects and returning python objects, along with a higher level object model that ensures memory safety. The idea is that users can use the RemoteMemory class in a memory safe way, and reach down to lower level UCX if they have other ideas on how the RMA operations should work.

There is also a UcxIO class that inherets RawIO from python so that one can turn rkeys into file handle like objects. I've toyed with it some, and I'm curious to see how far this can go.


pentschev:
We merged https://github.com/rapidsai/ucx-py/pull/693 today that [moved EP creation methods to `UCXEndpoint`](https://github.com/rapidsai/ucx-py/blob/f2d2fe6c78ff01b44b4705f0d3b65bf9f4cec22a/ucp/_libs/ucx_endpoint.pyx#L157-L227). Could you merge from upstream into your branch and update the relevant code?

Also we just entered [burndown phase for the next RAPIDS release](https://docs.rapids.ai/maintainers) a couple days ago, which means new features have to be merged before May 26. So if you need this for the June release, we need to have that in soon. 🙂

MattBBaker:
@pentschev Working on that as we speak. Would be nice to have in on time, but not mission critical.

pentschev:
Seems like there are some style issues, referring an old comment for reference on how to fix them: https://github.com/rapidsai/ucx-py/pull/704#issuecomment-816727231 . Alternatively, you can install `pip install pre-commit
` and then run `pre-commit install` inside the UCX-Py directory, and that will automatically fix the issues on new commits, or at least inform what the issues are for flake8 errors.

pentschev:
Hi @MattBBaker it seems that the tests are failing with a corrupted double-linked list error and seems like a legit error from this PR. Not sure if you had already seen, but let us know if you need any support in tracking down those errors.

MattBBaker:
Thanks @pentschev . I'm actually kind of stumped right now, the tests work fine on my desktop with no IB cards. I'm thinking I'll try a round of debugging on summit, though if you've noticed something silly that I did let me know.

pentschev:
I can reproduce the failing tests with both UCX 1.9 and master:

```
# UCX 1.9
UCXPY_IFNAME=enp1s0f0 UCX_TLS=tcp,cuda_copy,sockcm UCX_SOCKADDR_TLS_PRIORITY=sockcm pytest --cache-clear -vs ucp/_libs/tests/test -k test_ucxio

# UCX master
UCXPY_IFNAME=enp1s0f0 UCX_TLS=tcp,cuda_copy UCX_SOCKADDR_TLS_PRIORITY=tcp timeout 10m py.test --ignore-glob debug-tests --cache-clear -vs ucp/_libs/tests -k test_ucxio
```

The first test fails on the assertion, whereas the second hangs there forever. How are you setting up your environment and running the tests locally @MattBBaker ?

MattBBaker:
It's on my home desktop, so it's just an APU in this sucker so the only TLS being exercised is tcp, sm, and self. My GPU boxes are Power9 so that is always exciting when debugging. I'll poke at it

pentschev:
Interesting, I can confirm adding `self` to `UCX_TLS` indeed passes. To be honest, I'm not even sure what the `self` transport is, why is is necessary for this particular test?

MattBBaker:
There is no particular need for it. I just like to do as little with the UCX_* env as possible. If I set it to just tcp I get the error back, so now I can debug

pentschev:
Code freeze just hit, so I retargeted this PR for the next release.

MattBBaker:
@pentschev Roger that. Though the check-style seems to have gone insane complaining about undefined name 'time' after the time module was imported?

MattBBaker:
The rebase sounds like a good idea. Hopefully that cleans up the segfaults.

MattBBaker:
So, all the errors look unrelated. Looks largely like this,
```
file /opt/conda/envs/rapids/lib/python3.8/site-packages/distributed/comm/tests/test_ucx.py, line 228

  @pytest.mark.slow

  @pytest.mark.asyncio

  @pytest.mark.parametrize("n", [int(1e9), int(2.5e9)])

  async def test_large_cupy(n, cleanup):

      cupy = pytest.importorskip("cupy")

      com, serv_com = await get_comm_pair()



      arr = cupy.ones(n, dtype="u1")

      msg = {"op": "ping", "data": to_serialize(arr)}



      _, result = await asyncio.gather(com.write(msg), serv_com.read())

      data2 = result.pop("data")



      assert result["op"] == "ping"

      assert len(data2) == len(arr)

      await com.close()

      await serv_com.close()

E       fixture 'cleanup' not found

>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, benchmark, benchmark_weave, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cov, doctest_namespace, event_loop, monkeypatch, no_cover, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, testrun_uid, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, unused_tcp_port, unused_tcp_port_factory, worker_id

>       use 'pytest --fixtures [testpath]' for help on them.



/opt/conda/envs/rapids/lib/python3.8/site-packages/distributed/comm/tests/test_ucx.py:228
```

pentschev:
Yeah, this is unrelated to your PR and due to some changes in Distributed, I filed https://github.com/dask/distributed/issues/4902 for that issue.

I think in the meantime I'll just change the way we test Distributed to unblock CI here.

pentschev:
Opened https://github.com/rapidsai/ucx-py/pull/751 to resolve that, will trigger CI rerun once that's merged.

pentschev:
rerun tests

ucp/_libs/tests/test_rma.py
```diff
@@ -0,0 +1,140 @@
...
+    send_msg = bytes(os.urandom(msg_size))
+    req = self_mem.put_nb(send_msg, _)
+    while req is None:
+        counter = counter + 1
+        req = self_mem.put_nb(send_msg, _)
+        # This `if` is here because some combinations of transports, such as
+        # normal desktop PCs, will never have their transports exhausted. So
+        # we have a break to make sure this test still completes
+        if counter > 10000:
+            return
```
pentschev, line 132: I don't quite understand what's being done here. You first `put_nb` (outside of the loop), and then you count up to `10000` while `put_nb` is called every time again. Could you elaborate on what's this test really attempting to achieve? Furthermore, if you have to break here so that the test completes, what's the correct way to use this in systems such as desktop PCs as mentioned in the comment?

pentschev (commented):
Thanks @MattBBaker , looks good overall. I added a few comments mostly on cosmetic changes, and another one in a test that's not clear to me.

ucp/_libs/tests/test_mem.py
```diff
@@ -4,7 +4,7 @@
 
 import pytest
 
-from ucp._libs import ucx_api
+from ucp._libs import ucx_api as ucx_api
```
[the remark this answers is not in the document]
MattBBaker: No idea how that got in there. I blame the linting tools. Will fix up

ucp/_libs/ucxio.pyx
```diff
@@ -0,0 +1,68 @@
...
+    def write(self, buff):
+        data = Array(buff)
+        size = data.nbytes
+        if self.pos + size > self.length:
+            size = self.length - self.pos
+        finished = put_nbi(data, size, self.remote_addr + self.pos, self.rkey)
+        self.pos += size
+        if not finished:
+            self.flush()
+        return size
```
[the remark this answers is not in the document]
MattBBaker: I didn't realize how similar these ended up being. Will refactor

ucp/_libs/tests/test_rma.py
MattBBaker: The goal of this test is to hammer the put function until it runs out of resources and fails and goes into the request code path. On some transports like ugni and ib you will have a number of hardware resources where the op completes inline as a buffered send. The usual 'fire once and see if it works' style test will test this code path fine, but the code path that generates requests touches more code and is generally less tested. So this test will try and induce this behavior intentionally. 

The break is in there because on some transports like 'self' the test will never exhaust hardware resources as mentioned above because put will ultimately proxy into `memcpy()` and you will never get into a situation where you get a request, so the put will succeed forever and leave the test in an infinite loop.

pentschev: Sorry Matt for the delay, I'm on PTO so it's taking me longer to get to see PRs.

On cases such as `self` that you mentioned above, what will happen then if it will never exhaust, is it invalid to use `put_nb`? And if so, should we instead skip this test when certain transports are in use?

MattBBaker: That's fine, working on fixing the IO problem on ubuntu18. Which seems odd it's just on that one. But I can reproduce problems on my desktop, so it's a real thing.

Anyways, I should explain part of what I'm trying to do here at the low level driver/UCT layer. With a NIC like Cray Gemini/Aries or Mellanox IB the NIC will have a certain amount of hardware resources to send messages to remote NICs asynchronously without the host CPU getting involved. They are finite though, and if you do too many sends you can end up in a situation where UCP is using a queuing mechanism in UCT so that when a worker is progressed the worker can drive the TX queue, then try and empty the pending queue onto the TX queue. All of which involves callbacks. The goal of this test is to make sure that UCX gets into this state and then can be drained and everything that callbacks into Python works correctly.

The trouble with self is that since it doesn't have dedicated resources it has to use the host CPU and use it with `memcpy()`. It's still valid to call `put_nb` with self, it's just that it will never end up in the queuing code path. I don't want to try and restrict these mechanisms by particular transport, as that would run into problems with future transports and transports that aren't tested regularly. I'd rather have the test bang it's head into a wall and then decide, "If I didn't get to the queue code path by now, it is not going to happen ever."

pentschev: > I don't want to try and restrict these mechanisms by particular transport, as that would run into problems with future transports and transports that aren't tested regularly. I'd rather have the test bang it's head into a wall and then decide, "If I didn't get to the queue code path by now, it is not going to happen ever."

That's mostly fair, the only thing I don't like is this "silent success". Would it make more sense if instead of `return` we write `pytest.xfail` with some description of what's happening there?

MattBBaker: > That's mostly fair, the only thing I don't like is this "silent success". Would it make more sense if instead of return we write pytest.xfail with some description of what's happening there?
That sounds like the right way to go about that, though I have no idea how to use the `pytest.xfail` mechanism.

MattBBaker:
The errors don't seem to be related to my PR, 
`ImportError: libthrift.so.0.14.1: cannot open shared object file: No such file or directory`

madsbk:
rerun tests

madsbk:
rerun tests

pentschev (approved):
LGTM, thanks @MattBBaker !

PR #737, opened 2021-05-21
Drain tag messages on worker finalizer

pentschev:
The changes here resolve the issue that causes warnings to be raised at process termination, such as below:

```
[1621606990.204443] [dgx14:77681:0]       tag_match.c:61   UCX  WARN  unexpected tag-receive descriptor 0x55c9b1ff2fc0 was not matched
```

The reason for that was the we weren't draining the UCX worker during shutdown, where some messages may have gone uncatched, such as shutdown messages.

### DIFF
--- a/ucp/_libs/ucx_worker.pyx
+++ b/ucp/_libs/ucx_worker.pyx
@@ -69,6 +69,9 @@ def _ucx_worker_handle_finalizer(
     assert ctx.initialized
     cdef ucp_worker_h handle = <ucp_worker_h>handle_as_int
 
+    # This drains all the receive messages that were not received by the user with
+    # `tag_recv_nb`. Without this, UCX raises warnings such as below upon exit:
+    # `unexpected tag-receive descriptor ... was not matched`
     _drain_worker_tag_recv(handle)
 
     # Cancel all inflight messages
--- a/ucp/_libs/ucx_worker.pyx
+++ b/ucp/_libs/ucx_worker.pyx
@@ -19,11 +19,55 @@ from ..utils import nvtx_annotate
 logger = logging.getLogger("ucx")
 
 
+cdef _drain_worker_tag_recv(ucp_worker_h handle):
+    cdef ucp_tag_message_h
+    cdef ucp_tag_recv_info_t info
+    cdef ucs_status_ptr_t status
+    cdef void *buf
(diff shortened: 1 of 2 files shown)


pentschev:
@madsbk @quasiben could you take a look?

ucp/_libs/ucx_worker.pyx
```diff
@@ -19,12 +19,56 @@ from ..utils import nvtx_annotate
...
+        )
+        req = _handle_status(
+            status, info.length, _req_cb, (), {}, u"ucp_tag_msg_recv_nb", set()
+        )
+
+        if req is not None:
+            while _finished[0] is not True:
+                ucp_worker_progress(handle)
+
+        free(buf)
```
madsbk, line 61: `_handle_status()` can raise an exception so I think we should put it and the `free()` call in a `try-finally`

ucp/_libs/ucx_worker.pyx
```diff
@@ -19,12 +19,56 @@ from ..utils import nvtx_annotate
 logger = logging.getLogger("ucx")
 
 
+cdef _drain_worker_tag_recv(ucp_worker_h handle):
```
madsbk, line 22: I think it would be good with a description here.

madsbk (changes requested):
Nice, it would be great to get rid of that warning!

ucp/_libs/ucx_worker.pyx
pentschev: Thanks for catching this! Done in https://github.com/rapidsai/ucx-py/pull/737/commits/86f0f35382edb71cbdd97e79ef928ddb4e6ad859 .

ucp/_libs/ucx_worker.pyx
pentschev: Done in https://github.com/rapidsai/ucx-py/pull/737/commits/3ba02b7a2424ff8aa55f12c82296a147638d2bac .

pentschev:
Thanks @madsbk for the review, both requests were addressed. Could you take another look when you have the chance?

madsbk (approved):
LGTM, thanks @pentschev

pentschev:
Thanks again for the review @madsbk !

PR #739, opened 2021-05-26
Add support for AM API in local-send-recv benchmark

pentschev:


quasiben:
Do you see any performance differences between AM and Tag?

pentschev:
So far performance has been exactly the same with both, but I still need to run more extensive tests to check all transports.

pentschev:
Reported benchmarks with UCX master (pre-1.11) are as follows:

| Benchmark | TAG (GB/s) | AM (GB/s) |
|-|-|-|
| `tcp,sm` (NumPy) | 3.53 | 2.60 |
| `tcp,sm` (CuPy) | 2.69 | 2.72 |
| `tcp` (NumPy) | 2.50 | 2.15 |
| `tcp` (CuPy) | 2.68 | 2.69 |
| `rc` (NumPy) | 15.87 | 3.03 |
| `rc` (CuPy) | 10.52 | 10.51 |
| `cuda_ipc` (CuPy, Self) | 308.53 | 307.03 |
| `cuda_ipc` (CuPy, NV1) | 22.29 | 22.25 |
| `cuda_ipc` (CuPy, NV2) | 44.17 | 44.08 |

In most cases they're virtually the same, although there's some noticeable difference on CPU-only (NumPy) benchmarks, very visible in the `tcp,sm` and even more in `rc` transports. Maybe @dmitrygx would have some insight on what could be the cause for this?

pentschev:
I think this PR is in a good state now, would you mind taking a look @madsbk @quasiben ?

dmitrygx:
> In most cases they're virtually the same, although there's some noticeable difference on CPU-only (NumPy) benchmarks, very visible in the `tcp,sm` and even more in `rc` transports. Maybe @dmitrygx would have some insight on what could be the cause for this?

thanks for posting interesting results.
+ @brminich
@pentschev do you know what the message sizes are used in NumPy benchmark?

pentschev:
In all cases above I'm using 1GB messages, I've edited that comment to add that information.

brminich:
@pentschev, can you please try  rc (NumPy) with AM and UCX_RNDV_THRESH=64K set?
Just to make sure UCX is using rendezvous protocol

pentschev:
@brminich I tried this now but results are the same, let me know if there's something else I could try.

ajschmidt8:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

benchmarks/local-send-recv.py
```diff
@@ -31,20 +31,36 @@
 mp = mp.get_context("spawn")
 
 
+def register_am_allocators(args):
+    if not args.enable_am:
+        return
```
quasiben, line 36: If the use tried to call this, should we instead return False if `enable_am` is False as a sanity check?

quasiben: Sorry, this in the test not the ucp package, you can skip over my previous comment

quasiben (approved):
LGTM.  Thanks @pentschev

pentschev:
Thanks @quasiben for the review.

PR #743, opened 2021-05-31
Add local-send-recv benchmark using core library only

pentschev:


pentschev:
@madsbk this is based on your idea during our conversation earlier this week, a benchmark that's entirely built on the core library. Would you mind taking a look?

@quasiben also tagging you in case you'd like to take a look as well.

madsbk (approved):
LGTM but I think we should remove the `local-` prefix from the names of the `local-send-recv` benchmarks. 
The `local-` prefix is from before they supported running on multiple nodes.

pentschev:
> LGTM but I think we should remove the `local-` prefix from the names of the `local-send-recv` benchmarks.
> The `local-` prefix is from before they supported running on multiple nodes.

Agreed, this was something that was bugging me as well. Removed the prefix now.

ajschmidt8 (approved):
Approving `ops-codeowner` file changes

pentschev:
Thanks @madsbk and @ajschmidt8 for the reviews!

PR #755, opened 2021-06-23
Fix bad documentation in UCXIO

MattBBaker:
While making this class I found out that I could remove an ep parameter after I found out it was in rkey and I could use that instead of passing it in. It wasn't removed from the docs though. Fix the docs so that users won't be confused.


quasiben (commented):
Thanks @MattBBaker

quasiben (approved):

PR #758, opened 2021-06-30
Add `slow` marker to pytest

pentschev:
By adding the `slow` marker to `pytest` we can avoid using `--ignore-glob` on each test based on the condition of them being slow, providing also a clean output to users on skipped tests that may seem like they hang otherwise.


madsbk (approved):
LGTM, thanks @pentschev

pentschev:
Thanks for review @madsbk , we'll still need an approval from @rapidsai/ops-codeowners .

raydouglass (approved):

PR #760, opened 2021-07-09
Add all-to-all benchmark

pentschev:


quasiben:
cc @jakirkham @gjoseph92 you both may find this PR interesting for running networking experiments which compare Tornado, UCX, and Asyncio

pentschev:
rerun tests

jakirkham:
It might be interesting to try with uvloop as well. Since the event loop is largely handled by libuv in C, would expect that performs better than asyncio alone

pentschev:
Thanks for the suggestion @jakirkham , I added that now.

pentschev:
rerun tests

pentschev:
rerun tests

mrocklin:
I will be curious to see the results that come out of this.  If anyone has anything preliminary that they want to share I highly encourage that :)

tests/benchmark_multiple_processes_all_to_all.py
```diff
@@ -0,0 +1,240 @@
+import argparse
```
jakirkham, line 1: Do we want this in `tests/` or would it make sense to include in `benchmarks/`?

pentschev: I'm still thinking about it. I don't want it in tests, but I want a test there (for the UCX part only). However, a lot of the code is going to be shared and we don't have a good place currently where that common code would be visible to both. I'm not even sure the non-UCX code should live in this repo as we'll soon upstream it to OpenUCX, so it doesn't really make sense to have non-UCX code there. I'm still thinking of an appropriate place for this, if you have any suggestions, please let me know.

jakirkham: Yeah that makes sense. Will think about it as well.

PR #763, opened 2021-07-19
Try different interfaces if UCXPY_IFNAME is undefined

pentschev:
If `UCXPY_IFNAME` is undefined, let `ucp.get_address` try to pick a suitable interface to establish connections. This is helpful to allow CI testing on different systems without knowing what interfaces are available to establish connection between two UCX processes.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -83,8 +83,6 @@ function run_tests() {
     ls tests/
 
     # Setting UCX options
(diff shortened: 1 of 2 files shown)


ajschmidt8 (approved):

pentschev:
Thanks @ajschmidt8 !

quasiben (approved):

PR #764, opened 2021-07-20
Skip tests on missing imports

pentschev:
Many tests currently fail on missing optional packages (e.g., `distributed` and `rmm`), skips them when they're not available.

### DIFF
--- a/debug-tests/client.py
+++ b/debug-tests/client.py
@@ -11,9 +11,9 @@ from debug_utils import (
     start_process,
     total_nvlink_transfer,
 )
-from utils import recv, send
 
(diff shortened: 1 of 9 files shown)


pentschev:
rerun tests

quasiben (approved):

quasiben:
rerun tests

pentschev:
Weird, running the benchmark that failed twice here for 100 times locally didn't fail even once. Since this PR doesn't touch that code I'm gonna go ahead and merge it, but will keep an eye to see whether the benchmark fails again.

PR #767, opened 2021-07-23
Change core tests default endpoint error handling according to UCX version

pentschev:
On newer UCX versions it's important to have endpoint error handling enabled, this ensures that core tests will enable it depending on the UCX version installed.


quasiben:
Thanks for doing this @pentschev.  Let's merge when tests pass

quasiben (approved):

pentschev:
Thanks for the approval @quasiben !

PR #769, opened 2021-07-23
Drop disconnect/shutdown tests without error handling in UCX >= 1.10

pentschev:
Endpoint error handling shall always be enabled on UCX 1.10 and above, this removes disconnect/shutdown testing of UCX 1.10+ when endpoint error handling is disabled.


quasiben (approved):
LGTM

PR #772, opened 2021-08-19
Cancel waiting `am_recv` if endpoint closes

pentschev:
Cancelling `am_recv` was only done when an error occurred during AM message receive callback in the worker, but if an endpoint closed while the client is waiting for that message it would hang indefinitely. This change cancels any waiting `am_recv` at the endpoint finalizer as well.

### DIFF
--- a/tests/test_shutdown.py
+++ b/tests/test_shutdown.py
@@ -25,6 +25,11 @@ def event_loop(scope="function"):
     loop.close()
 
 
+def _skip_if_not_supported(message_type):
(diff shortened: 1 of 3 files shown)


pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
rerun tests

madsbk (approved):
LGTM, thanks @pentschev

pentschev:
Thanks @madsbk for reviewing!

PR #773, opened 2021-08-19
Fix UCXAddress finalizer

pentschev:
According to the UCX documentation, the [worker address cannot be used after `ucp_worker_release_address`](https://openucx.github.io/ucx/api/v1.11/html/group___u_c_p___w_o_r_k_e_r.html#ga94260829739496267d2c8d86414b863d). This change fixes that by adding a finalizer to `UCXAddress` that will only release the handle when the object is destroyed.

Even though `test_peer_communication` passed, the errors would be printed to stdout, which this change resolves as well.

```python
  File "/datasets/pentschev/miniconda3/envs/gdf/lib/python3.8/site-packages/ucp/exceptions.py", line 13, in log_errors
    yield
  File "ucp/_libs/transfer_common.pyx", line 55, in ucp._libs.ucx_api._send_callback
    cb_func(req, exception, *cb_args, **cb_kwargs)
  File "ucp/_libs/ucx_rkey.pyx", line 13, in ucp._libs.ucx_api._ucx_remote_mem_finalizer_post_flush
    assert exception is None
AssertionError
[1629385917.006941] [dgx13:18276:0]           mpool.c:44   UCX  WARN  object 0x558daf1c9a40 was not returned to mpool ucp_rkeys
[1629385917.006965] [dgx13:18276:0]           mpool.c:44   UCX  WARN  object 0x558daf1c9bc0 was not returned to mpool ucp_rkeys
[1629385917.020369] [dgx13:18174:0]           mpool.c:44   UCX  WARN  object 0x5606b2b67bc0 was not returned to mpool ucp_rkeys
[1629385917.020377] [dgx13:18174:0]           mpool.c:44   UCX  WARN  object 0x5606b2b67c80 was not returned to mpool ucp_rkeys
```

### DIFF
--- a/ucp/_libs/ucx_address.pyx
+++ b/ucp/_libs/ucx_address.pyx
@@ -32,7 +32,7 @@ cdef class UCXAddress(UCXObject):
     def __cinit__(
             self,
             uintptr_t address_as_int,
-            Py_ssize_t length,
+            size_t length,
             UCXWorker worker=None,
     ):
         address = <ucp_address_t *> address_as_int
@@ -63,7 +63,7 @@ cdef class UCXAddress(UCXObject):
         cdef size_t length
         status = ucp_worker_get_address(ucp_worker, &address, &length)
         assert_ucs_status(status)
-        return UCXAddress(int(<uintptr_t>address), length, worker=worker)
+        return UCXAddress(<uintptr_t>address, length, worker=worker)
 
     @property
     def address(self):
--- a/ucp/_libs/ucx_address.pyx
+++ b/ucp/_libs/ucx_address.pyx
@@ -12,20 +12,42 @@ from .arr cimport Array
 from .ucx_api_dep cimport *
 
 
-cdef class UCXAddress:
+def _ucx_address_finalizer(
+    uintptr_t handle_as_int,
+    uintptr_t worker_handle_as_int,
+):
+    cdef ucp_address_t *address = <ucp_address_t *>handle_as_int
+    cdef ucp_worker_h worker = <ucp_worker_h>worker_handle_as_int
+    if worker_handle_as_int != 0:
+        ucp_worker_release_address(worker, address)
+    else:
+        free(address)
+
+
+cdef class UCXAddress(UCXObject):
     """Python representation of ucp_address_t"""
     cdef ucp_address_t *_address
     cdef Py_ssize_t _length
 
-    def __cinit__(self, uintptr_t address_as_int, Py_ssize_t length):
+    def __cinit__(
+            self,
+            uintptr_t address_as_int,
+            Py_ssize_t length,
+            UCXWorker worker=None,
+    ):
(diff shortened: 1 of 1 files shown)


pentschev:
rerun tests

ucp/_libs/ucx_address.pyx
```diff
@@ -41,10 +63,7 @@ cdef class UCXAddress:
         cdef size_t length
         status = ucp_worker_get_address(ucp_worker, &address, &length)
         assert_ucs_status(status)
-        try:
-            return UCXAddress(int(<uintptr_t>address), length)
-        finally:
-            ucp_worker_release_address(ucp_worker, address)
+        return UCXAddress(int(<uintptr_t>address), length, worker=worker)
```
jakirkham, line 66: Do we still need to convert to a Python `int` given the constructor accepts `uintptr_t`?

```suggestion
        return UCXAddress(<uintptr_t>address, length, worker=worker)
```

jakirkham, line 66: It seems we are using `size_t` for `length` here, but the constructor takes `Py_ssize_t`. Should we consolidate to one of these two types?

jakirkham:
Thanks Peter! 😄

Had a couple of questions above

pentschev:
Thanks @jakirkham for the review. I'm guessing the PR itself is perfect, given both your comments were relating to the original code that I didn't touch! 🤣 

Anyway, both your comments were addressed in https://github.com/rapidsai/ucx-py/pull/773/commits/37b8c23c7b86663de4d00ff2252834600ffebb9b .

ucp/_libs/ucx_address.pyx
```diff
@@ -12,20 +12,42 @@ from .arr cimport Array
...
+    if worker_handle_as_int != 0:
+        ucp_worker_release_address(worker, address)
+    else:
+        free(address)
+
+
+cdef class UCXAddress(UCXObject):
     """Python representation of ucp_address_t"""
     cdef ucp_address_t *_address
     cdef Py_ssize_t _length
```
jakirkham, line 30:
```suggestion
    cdef size_t _length
```

pentschev: Good catch!

jakirkham:
Yeah it seems ok to me.

Sorry for the other changes. Just became apparent there may be other issues here that we should probably cleanup.

cc @madsbk @MattBBaker (as both of you have worked on this class before and may have thoughts)

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
rerun tests

madsbk (approved):
LGTM, thanks @pentschev

pentschev:
rerun tests

jakirkham:
Thanks Peter for the PR and Mads for the review! 😄

pentschev:
Thanks @jakirkham and @madsbk for reviews!

PR #775, opened 2021-08-23
Remove `guarantee_msg_order`

### LINKED ISSUE #501: Removing the guarantee_msg_order option
Currently, the [`guarantee_msg_order` option is disable by default`](https://github.com/rapidsai/ucx-py/pull/496) but at some point we should remove the option completely.

pentschev:
This feature was originally introduced to overcome a supposed limitation on the UCX Tag API that would not necessarily guarantee message ordering. However, this is not true (at least in recent UCX versions), and thus this option is unnecessary. This has been disabled by default for a long time and has not presented any issues, thus removing it should have no negative impact.

Closes #501 .


pentschev:
rerun tests

pentschev:
rerun tests

madsbk (approved):
LGTM, thanks @pentschev

pentschev:
Thanks @madsbk for review and approval!

PR #777, opened 2021-08-24
Add support for endpoints from worker address on async API

pentschev:
Using a UCX listener may not be possible for systems that are IP-less, as that is required to create an endpoint to the listening server. For such cases, it's possible to use the UCX worker address to connect to a remote worker. This PR exposes the worker address feature to the UCX-Py async API, additionally includes a couple of tests that can be used as reference for setting up a "server" worker that exposes its address and handle client connections in a similar way to how the UCX-Py listener does.

### DIFF
--- a/tests/test_from_worker_address.py
+++ b/tests/test_from_worker_address.py
@@ -1,0 +1,72 @@
+import asyncio
+import multiprocessing as mp
+
+import numpy as np
+
+import ucp
+
+mp = mp.get_context("spawn")
+
+
+def _test_from_worker_address_server(queue):
+    async def run():
+        # Send worker address to client process via multiprocessing.Queue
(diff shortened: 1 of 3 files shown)


pentschev:
@madsbk this is a first pass on handling workers without a listener, created from the worker address. I'm still working on a more elaborate test/example, but this should be good to give an idea of how I imagine things could work.

pentschev:
@madsbk also added a multinode test now that transfers UCXAddress along with tag on a single tag operation, ensuring each incoming message directly on the worker is an incoming client. The received message is then used to create an endpoint to the client and handle the connection similar to what a listener does.

madsbk (approved):

pentschev:
Thanks @madsbk for the review and approval.

PR #779, opened 2021-09-09
Add FutureWarning of UCX < 1.11.1 deprecation

pentschev:
Besides stability, UCX 1.11.1 includes important features such as:

* Automatic InfiniBand<->CUDA device mapping without specifying `UCX_NET_DEVICES` explicitly;
* InfiniBand allocation registration cache without patching source;
* Endpoint error handling with CUDA IPC;
* etc.

All features above will simplify usage both with and without Dask and RAPIDS. Therefore, we would like to deprecate older versions of UCX, to be removed at a later release.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -91,8 +91,8 @@ Instructions for building UCX 1.11.1:
     make -j install
 
 
-UCX-1.9
-~~~~~~~
+UCX-1.9 (Deprecated)
+~~~~~~~~~~~~~~~~~~~~
 
 Instructions for building ucx 1.9:
 
@@ -131,7 +131,7 @@ As noted above, the UCX conda package no longer builds support for IB/RDMA.  To 
 
 If OFED drivers are not installed on the machine, you can download drivers at directly from `Mellanox <https://www.mellanox.com/products/infiniband-drivers/linux/mlnx_ofed>`_.  For versions older than 5.1 click on, *archive versions*.
 
-Building UCX 1.9 or 1.11 as shown previously should automatically include IB/RDMA support if available in the system. It is possible to explicitly activate those, ensuring the system satisfies all dependencies or fail otherwise, by including the ``--with-rdmacm`` and ``--with-verbs`` build flags. For example:
+Building UCX 1.11.1 or 1.9 (deprecated) as shown previously should automatically include IB/RDMA support if available in the system. It is possible to explicitly activate those, ensuring the system satisfies all dependencies or fail otherwise, by including the ``--with-rdmacm`` and ``--with-verbs`` build flags. For example:
 
 ::
 
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -5,6 +5,7 @@
 
 import logging
 import os
+import warnings
 
 from ._version import get_versions as _get_versions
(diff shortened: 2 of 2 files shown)


pentschev:
I'm not necessarily tied to this idea, but perhaps we should start warning users of UCX < 1.11.1 deprecation? Also it doesn't need to be right now, we could delay it for 21.12 or another RAPIDS release.

quasiben:
I'm not against deprecating but I think the sentiment you raised around waiting a bit is probably correct.  However, I don't know of anyone using UCX-Py who can't upgrade UCX

pentschev:
> I'm not against deprecating but I think the sentiment you raised around waiting a bit is probably correct. However, I don't know of anyone using UCX-Py who can't upgrade UCX

It's not too much on upgrading or not, but fundamentally older UCX (such as 1.9) lacked a lot of functionality that are now implemented in UCX 1.11.1 (e.g., automatic IB device detection, IB allocation registration cache, endpoint error handling with CUDA IPC) that really simplifies things for us by a lot, and removing that code sometime in the near future would be great. Therefore, alerting users would probably be a good first step, also to hear from users who can't potentially upgrade rather sooner than later.

jakirkham:
This is just adding a warning right? Not actually dropping anything. Think it makes sense to get this in sooner rather than later. It will give us flexibility later on when deciding whether we want to drop or not. The longer we wait to communicate this to users, the longer it takes for us to actually do the deprecation, which may come at a point where things are more painful

jakirkham:
That said, it's probably worth raising this with key stakeholders first just so they are aware and we can make any changes to the deprecation schedule if needed.

pentschev:
cc @dantegd @cjnolet for cuML
cc @BradReesWork @rlratzel for cuGraph
cc @randerzander @beckernick @VibhuJawa @ayushdg for Dask workflows
cc @JohnZed generally for RAPIDS
cc @benjha @MattBBaker @jglaser for Summit/ORNL

benjha:
Hi all,

What do you recommend for a "quick" check ? Build ucx-py  1.11.1 in a stand alone environment and run the test suite ?

pentschev:
@benjha yes, it would be good to run UCX-Py and Dask-CUDA tests if you can:

```
# UCX-Py
git clone https://github.com/rapidsai/ucx-py
cd ucx-py
pip install .
pytest --cache-clear -vs ucp/_libs/tests/
pytest --cache-clear -vs tests/

# Dask-CUDA
git clone https://github.com/rapidsai/dask-cuda
cd dask-cuda
pip install .
pytest -vs --cache-clear dask_cuda/tests/
```

pentschev:
I've also checked with DLFW folks offline, and they're already running some RC version of UCX 1.11.1, so dropping compatibility code isn't a problem for them, as that only affects UCX 1.9 and older.

pentschev:
Since no objections were raised in over 2 months, I'll plan to merge this later today. We'll discuss a plan to completely drop support for UCX < 1.11.1 early next year.

pentschev:
rernu tests

pentschev:
rerun tests

pentschev:
I'm doing some final testing, will merge this tomorrow or Monday.

PR #781, opened 2021-09-16
Add reference Docker container

pentschev:
To help users get started with UCX-Py, add a reference Docker container. The container is based off of the official `python:3` container, we then install dependencies required to build and use UCX and UCX-Py via `apt`. When running, the container will install Python dependencies required by UCX-Py, build UCX from source, build UCX-Py and run tests and benchmarks. Currently includes a minimal setup only, without CUDA, MOFED or rdma-core support.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:3
+
+RUN apt-get update && \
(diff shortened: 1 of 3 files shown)


madsbk (approved):
LGTM, thanks @pentschev

pentschev:
Thanks @madsbk for the review!

PR #782, opened 2021-09-17
Add compile-time variable UCXPY_DISABLE_HWLOC

pentschev:
For UCX 1.9 and older we needed hwloc to determine the closest InfiniBand device to a GPU. With UCX 1.10+ this isn't necessary anymore, so it's a deprecated feature. Since building with hwloc can be problematic for some use cases, add a new compile-time environment variable to allow disabling build of topological distance code.


madsbk (approved):
LGTM, thanks @pentschev

pentschev:
Thanks for the review @madsbk !

PR #784, opened 2021-09-22
Make ucx-py installation blocks consistent

charlesbluca:
Consistently instructs users to use `pip install -v .` to install ucx-py, to remove any ambiguity from the instructions on disabling building with hwloc support.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -163,4 +163,6 @@ UCX-Py
 In UCX 1.10 and above, or for builds that don't need CUDA and InfiniBand support, users can disable building with hwloc support:
 
 ::
+
+    UCXPY_DISABLE_HWLOC=1 python setup.py build_ext --inplace
     UCXPY_DISABLE_HWLOC=1 pip install .
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -155,8 +155,7 @@ UCX-Py
     conda activate ucx
     git clone https://github.com/rapidsai/ucx-py.git
     cd ucx-py
-    python setup.py build_ext --inplace
-    pip install .
+    pip install -v .
     # or for develop build
     pip install -v -e .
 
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -163,4 +163,6 @@ UCX-Py
 In UCX 1.10 and above, or for builds that don't need CUDA and InfiniBand support, users can disable building with hwloc support:
 
 ::
+
+    UCXPY_DISABLE_HWLOC=1 python setup.py build_ext --inplace
```
pentschev, line 167: I really don't understand the need for this. Why isn't just `UCXPY_DISABLE_HWLOC=1 pip install .` enough? I tested it locally just like that and it did work.

pentschev: Forgot to mention, but I also had another user testing it just the same and it worked for him as well.

charlesbluca: Ah I understand now that `pip install .` implicitly does the Cython build step, so we only need to pass the flag to that one command. Knowing that, this line is definitely unnecessary.

Out of interest, why in the [block directly above this](https://github.com/rapidsai/ucx-py/blob/branch-0.22/docs/source/install.rst#ucx-py) do we instruct to build the Cython separately from the `pip install`?

```
python setup.py build_ext --inplace
pip install .
```

Could this not just be `pip install -v .`?

pentschev: I'm not entirely sure, I think it was @quasiben who originally wrote it, it's possible he took the steps to build as he did when doing more verbose/debug builds. I'm in favor of removing that line and simplifying install steps. Could you do that @charlesbluca ?

charlesbluca: Sure!

pentschev (approved):
LGTM, thanks @charlesbluca !

docs/source/install.rst
quasiben: I can't recall if I did this, and if I did, why it is the case.  +1 for removing

PR #785, opened 2021-09-28
Add capability to register a user callback for receiving AM messages

pentschev:


pentschev:
The tests that are failing here are due to the behavior described in https://github.com/openucx/ucx/issues/6721#issuecomment-932226627 . It's still unclear whether this is a UCX bug or by design, but I'm fairly confident this PR is good once that is clarified.

PR #787, opened 2021-10-05
Add --delay-progress option to send-recv-core benchmark

pentschev:
After a lot of benchmarking of UCX-Py and digging into the details of `ucx_perftest`, I've found out that one of the biggest differences in performance there, particularly for short messages, is that it doesn't do blocking send/recv, but instead delays calling `ucp_worker_progress` based on the number of outstanding operations. This is great to show the maximum potential of UCX/UCX-Py, but I'm still unsure whether this can be used in the async API, much less in applications such as Dask.

### DIFF
--- a/benchmarks/send-recv-core.py
+++ b/benchmarks/send-recv-core.py
@@ -133,9 +133,12 @@ def server(queue, args):
 
     def _tag_recv_handle(request, exception, ep, msg):
         assert exception is None
-        ucx_api.tag_send_nb(
+        req = ucx_api.tag_send_nb(
             ep, msg, msg.nbytes, tag=0, cb_func=_send_handle, cb_args=(msg,)
         )
+        if req is None:
+            with finished_lock:
+                finished[0] += 1
 
     def _am_recv_handle(recv_obj, exception, ep):
         assert exception is None
--- a/benchmarks/send-recv-core.py
+++ b/benchmarks/send-recv-core.py
@@ -142,7 +142,7 @@ def server(queue, args):
         msg = Array(recv_obj)
         ucx_api.am_send_nbx(ep, msg, msg.nbytes, cb_func=_send_handle, cb_args=(msg,))
 
-    def _listener_handler(conn_request):
+    def _listener_handler(conn_request, msg):
         global ep
         ep = ucx_api.UCXEndpoint.create_from_conn_request(
             worker,
@@ -150,23 +150,13 @@ def server(queue, args):
             endpoint_error_handling=ucx_api.get_ucx_version() >= (1, 10, 0),
         )
 
-        if not args.enable_am:
-            msg_recv_list = []
-            if not args.reuse_alloc:
-                for _ in range(args.n_iter):
-                    msg_recv_list.append(xp.zeros(args.n_bytes, dtype="u1"))
-            else:
-                t = xp.zeros(args.n_bytes, dtype="u1")
-                for _ in range(args.n_iter):
-                    msg_recv_list.append(t)
-
-            assert msg_recv_list[0].nbytes == args.n_bytes
-
         for i in range(args.n_iter):
             if args.enable_am is True:
                 ucx_api.am_recv_nb(ep, cb_func=_am_recv_handle, cb_args=(ep,))
             else:
-                msg = Array(msg_recv_list[i])
+                if not args.reuse_alloc:
+                    msg = Array(xp.zeros(args.n_bytes, dtype="u1"))
+
                 ucx_api.tag_recv_nb(
                     worker,
                     msg,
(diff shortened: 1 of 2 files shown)


pentschev:
Also for comparison, below are the same results but for the async API (about 3x slower than the sync API and 5-10x slower when also using `--delay-progress`):

<details><summary>NumPy (async)</summary>

```
$ UCX_RNDV_THRESH=8K UCX_TLS=tcp,cuda_copy,cuda_ipc python benchmarks/send-recv.py --reuse-alloc -o numpy -n 100000 -b 0 -c 1 --n-iter 1000
Server Running at 10.33.225.163:33278
Client connecting to server at 10.33.225.163:33278
Roundtrip benchmark
--------------------------
n_iter          | 1000
n_bytes         | 97.66 kiB
object          | numpy
reuse alloc     | True
transfer API    | TAG
UCX_TLS         | tcp,cuda_copy,cuda_ipc
UCX_NET_DEVICES | all
==========================
Device(s)       | CPU-only
Server CPU      | 0
Client CPU      | 1
Average         | 86.95 MiB/s
Median          | 486.94 MiB/s
```

</details>

<details><summary>CuPy (async)</summary>

```
$ UCX_RNDV_THRESH=8K UCX_TLS=tcp,cuda_copy,cuda_ipc python benchmarks/send-recv.py --reuse-alloc -o cupy -n 100000 -d 0 -e 3 -b 0 -c 1 --n-iter 1000
Server Running at 10.33.225.163:52799
Client connecting to server at 10.33.225.163:52799
Roundtrip benchmark
--------------------------
n_iter          | 1000
n_bytes         | 97.66 kiB
object          | cupy
reuse alloc     | True
transfer API    | TAG
UCX_TLS         | tcp,cuda_copy,cuda_ipc
UCX_NET_DEVICES | all
==========================
Device(s)       | 0, 3
Average         | 103.06 MiB/s
Median          | 645.10 MiB/s
```

</details>

benchmarks/send-recv-core.py
```diff
@@ -290,13 +351,13 @@ def client(queue, port, server_address, args):
...
-        ts = format_bytes(2 * args.n_bytes / t)
-        ts = (" " * (9 - len(ts))) + ts
-        print("%03d         |%s/s" % (i, ts))
+    # print("--------------------------")
+    # print("Iterations")
+    # print("--------------------------")
+    # for i, t in enumerate(times):
+    #     ts = format_bytes(2 * args.n_bytes / t)
+    #     ts = (" " * (9 - len(ts))) + ts
+    #     print("%03d         |%s/s" % (i, ts))
```
madsbk, line 360: Should we make this info optional?

ucp/_libs/utils_test.py
```diff
@@ -51,6 +51,49 @@ def blocking_recv(worker, ep, msg, tag=0):
...
+        msg,
+        msg.nbytes,
+        tag=tag,
+        cb_func=non_blocking_handler,
+        cb_args=(completed_cb,),
+        ep=ep,
+    )
+    if req is None:
+        completed_cb()
+    # print(f"recv: {count[0]}")
```
madsbk, line 93:
```suggestion
```

ucp/_libs/utils_test.py
```diff
@@ -51,6 +51,49 @@ def blocking_recv(worker, ep, msg, tag=0):
...
+        ep,
+        msg,
+        msg.nbytes,
+        tag=tag,
+        cb_func=non_blocking_handler,
+        cb_args=(completed_cb,),
+    )
+    if req is None:
+        completed_cb()
+    # print(f"send: {count[0]}")
```
madsbk, line 75:
```suggestion
```

ucp/_libs/utils_test.py
```diff
@@ -51,6 +51,49 @@ def blocking_recv(worker, ep, msg, tag=0):
             worker.progress()
 
 
+def non_blocking_handler(request, exception, completed_cb):
+    if exception is not None:
+        print(exception)
+    assert exception is None
+    completed_cb()
+    # print(f"handler: {count[0]}")
```
madsbk, line 59:
```suggestion
```

madsbk (changes requested):
Thanks @pentschev, very interesting! Maybe it is time to investigate possible strategics for use of `ucp_worker_progress` in Dask. 

The PR LGTM, I only have some minor suggestions.

benchmarks/send-recv-core.py
pentschev: Good idea, added a new `--no-detailed-report` to suppress that.

pentschev:
Thanks @madsbk for the review. I was working on removing some code duplication here but the changes ended up being too large, so I'll leave that for a follow-up PR.

pentschev:
I updated the performance numbers from https://github.com/rapidsai/ucx-py/pull/787#issuecomment-934912877 and https://github.com/rapidsai/ucx-py/pull/787#issuecomment-935612139 . It seems the DGX-1 I was running on had some lingering processes that had a significant negative impact on absolute performance numbers, but the relative performance numbers remain the same.

PR #791, opened 2021-10-11
Fix AM wireup message exchange in core benchmark

pentschev:
Wireup message was introduced in https://github.com/rapidsai/ucx-py/pull/787 but that change missed AM, this fixes that case.


madsbk (approved):
LGTM, thanks @pentschev

PR #796, opened 2021-10-19
Update CI default to UCX 1.11

pentschev:
RAPIDS 21.10 and above require UCX 1.11 and UCX 1.9 is therefore not tested anymore. Updates ensure always setting UCX configurations to match UCX 1.11+.


pentschev:
rerun tests

pentschev:
Thanks @Ethyling for the review/approval!

PR #798, opened 2021-10-20
Do not set UCX_MEMTYPE_CACHE for UCX >= 1.12

pentschev:
`UCX_MEMTYPE_CACHE=n` is no longer needed with UCX 1.12 and above, and raises the warning below when set:

```
parser.c:1907 UCX  WARN  unused env variable: UCX_MEMTYPE_CACHE (set UCX_WARN_UNUSED_ENV_VARS=n to suppress this warning)
```


quasiben (approved):
I'm glad this is going away for future UCX versions

PR #804, opened 2021-11-01
Update UCP configuration validation tests for UCX >= 1.12

pentschev:
Starting from UCX 1.12, it's not possible to validate non-UCP configuration options, update tests accordingly.

See https://github.com/openucx/ucx/issues/7519 .

PR #805, opened 2021-11-03
Fix core config tests for UCX 1.12

pentschev:
Config tests for UCX 1.12 with the async API were already fixed in https://github.com/rapidsai/ucx-py/pull/804, but the core tests missed the same fixes.

Additionally, skip `TopologicalDistance` tests when UCX-Py is built using `UCXPY_DISABLE_HWLOC=1`, as well as skipping endpoint close callback tests for UCX < 1.11.

### DIFF
--- a/tests/test_endpoint.py
+++ b/tests/test_endpoint.py
@@ -4,9 +4,12 @@ import ucp
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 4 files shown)


quasiben (approved):

pentschev:
Thanks @quasiben !

PR #807, opened 2021-11-04
Reorder and reenable UCX_MEMTYPE_CACHE=n for all UCX versions

pentschev:
UCX_MEMTYPE_CACHE=n is still required even for UCX 1.12 (see
https://github.com/openucx/ucx/issues/7575), so we reenable it.
Reordering the import is also required to prevent UCS warnings when
Cython code is imported.

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -6,19 +6,20 @@
 import logging
 import os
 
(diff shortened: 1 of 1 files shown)


pentschev:
Thanks @madsbk for the quick review!

PR #810, opened 2021-11-09
Ensure am_recv waiting futures are canceled

pentschev:
When am_recv is being awaited and an endpoint error, the future needs to be canceled to prevent an indefinite hang.

### DIFF
--- a/ucp/_libs/ucx_worker.pyx
+++ b/ucp/_libs/ucx_worker.pyx
@@ -96,7 +96,7 @@ cdef class UCXWorker(UCXObject):
         ucp_worker_h _handle
         UCXContext _context
         set _inflight_msgs
-        set _inflight_msgs_to_cancel
+        dict _inflight_msgs_to_cancel
         IF CY_UCP_AM_SUPPORTED:
             dict _am_recv_pool
             dict _am_recv_wait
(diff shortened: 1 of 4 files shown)


pentschev:
rerun tests

ucp/_libs/ucx_worker.pyx
```diff
@@ -120,7 +120,7 @@ cdef class UCXWorker(UCXObject):
         status = ucp_worker_create(context._handle, &worker_params, &self._handle)
         assert_ucs_status(status)
         self._inflight_msgs = set()
-        self._inflight_msgs_to_cancel = set()
+        self._inflight_msgs_to_cancel = dict({"tag": set()})
```
madsbk, line 123:
```suggestion
        self._inflight_msgs_to_cancel = {"tag": set()}
```

madsbk: Do we need to use `dict` here explicitly?

madsbk (approved):
LGTM, only a very minor suggestion.

ucp/_libs/ucx_worker.pyx
pentschev: You're right, that's redundant, applied the suggestion now.

PR #813, opened 2021-11-10
Prevent endpoint canceling AM messages multiple times

pentschev:
This is important because UCX may reuse the same endpoint handle, and if a message is canceled during then endpoint finalizer, a message received on the same (new) endpoint handle may be canceled incorrectly.

Due to this, `test_close_after_n_recv[am]` would fail non-deterministically.

### DIFF
--- a/tests/test_endpoint.py
+++ b/tests/test_endpoint.py
@@ -34,3 +34,32 @@ async def test_close_callback(server_close_callback):
     listener = ucp.create_listener(server_node,)
     await client_node(listener.port)
     assert closed[0] is True
+
+
+@pytest.mark.asyncio
+@pytest.mark.skipif(
+    ucp.get_ucx_version() < (1, 11, 0),
+    reason="Endpoint error handling is unreliable in UCX releases prior to 1.11.0",
+)
+@pytest.mark.parametrize("transfer_api", ["am", "tag"])
+async def test_cancel(transfer_api):
+    async def server_node(ep):
+        await ep.close()
+
+    async def client_node(port):
+        ep = await ucp.create_endpoint(ucp.get_address(), port,)
+        if transfer_api == "am":
+            with pytest.raises(
+                ucp.exceptions.UCXCanceled, match="am_recv",
+            ):
+                await ep.am_recv()
+        else:
+            with pytest.raises(
+                ucp.exceptions.UCXCanceled, match="Recv.*tag",
+            ):
+                msg = bytearray(1)
+                await ep.recv(msg)
+        await ep.close()
+
+    listener = ucp.create_listener(server_node,)
+    await client_node(listener.port)
--- a/tests/test_endpoint.py
+++ b/tests/test_endpoint.py
@@ -47,7 +47,7 @@ async def test_cancel(transfer_api):
         await ep.close()
 
     async def client_node(port):
-        ep = await ucp.create_endpoint(ucp.get_address(), port,)
+        ep = await ucp.create_endpoint(ucp.get_address(), port)
         if transfer_api == "am":
             with pytest.raises(
                 ucp.exceptions.UCXCanceled, match="am_recv",
@@ -61,5 +61,5 @@ async def test_cancel(transfer_api):
                 await ep.recv(msg)
         await ep.close()
 
-    listener = ucp.create_listener(server_node,)
+    listener = ucp.create_listener(server_node)
     await client_node(listener.port)
--- a/ucp/_libs/tests/test_cancel.py
+++ b/ucp/_libs/tests/test_cancel.py
@@ -40,7 +40,7 @@ def _server_cancel(queue, transfer_api):
         )
 
     listener = ucx_api.UCXListener(worker=worker, port=0, cb_func=_listener_handler)
-    queue.put(listener.port),
+    queue.put(listener.port)
 
     while ep[0] is None:
         worker.progress()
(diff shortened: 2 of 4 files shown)


ucp/_libs/ucx_endpoint.pyx
```diff
@@ -54,6 +54,13 @@ cdef int _cancel_am_recv(UCXWorker worker, uintptr_t handle_as_int):
                 **cb_kwargs
             )
 
+        # Prevent endpoint canceling AM messages multiple times.  This is important
+        # because UCX may reuse the same endpoint handle, and if a message is canceled
+        # during then endpoint finalizer, a message received on the same (new) endpoint
+        # handle may be canceled incorrectly.
+        del worker._am_recv_wait[handle_as_int]
+        worker._inflight_msgs_to_cancel["am"].discard(handle_as_int)
```
madsbk, line 62:
```suggestion
        # Prevent endpoint canceling AM messages multiple times. This is important
        # because UCX may reuse the same endpoint handle, and if a message is canceled
        # during then endpoint finalizer, a message received on the same (new) endpoint
        # handle may be canceled incorrectly.
        del worker._am_recv_wait[handle_as_int]
        worker._inflight_msgs_to_cancel["am"].discard(handle_as_int)
```

madsbk (approved):
LGTM, only a very minor suggestion :P

ucp/_libs/ucx_endpoint.pyx
pentschev: Ha! Good catch!

pentschev:
I just realized I'm changing the size of `self._inflight_msgs_to_cancel["am"]` set during a loop with this change. I'll need to rework it a little to prevent such warnings and will push a fix to this shortly.

pentschev:
@madsbk I now fixed the issue I mentioned above, could you take another look if you have the chance?

madsbk:
> @madsbk I now fixed the issue I mentioned above, could you take another look if you have the chance?

Looks good but it is possible to write a  regression test of the mentioned issue?

pentschev:
> Looks good but it is possible to write a regression test of the mentioned issue?

I've added a couple of tests now, please let me know if you think they're satisfactory. It's difficult to test the exact behavior of `UCXWorker.cancel_inflight_messages` from the async API because of the lack of control to progress the worker, but I added a test for that in the core library.

ucp/_libs/ucx_endpoint.pyx
```diff
@@ -54,6 +65,27 @@ cdef int _cancel_am_recv(UCXWorker worker, uintptr_t handle_as_int):
...
+    cdef size_t len_wait = 0
+
+    if _is_am_enabled(worker):
+        if handle_as_int == 0:
+            for handle_as_int in worker._inflight_msgs_to_cancel["am"]:
+                len_wait += _cancel_am_recv_single(worker, handle_as_int)
+
+            # Prevent endpoint canceling AM messages multiple times. This is important
+            # because UCX may reuse the same endpoint handle, and if a message is
+            # canceled # during then endpoint finalizer, a message received on the same
```
madsbk, line 82:
```suggestion
            # canceled during then endpoint finalizer, a message received on the same
```

tests/test_endpoint.py
```diff
@@ -34,3 +34,32 @@ async def client_node(port):
...
+                await ep.am_recv()
+        else:
+            with pytest.raises(
+                ucp.exceptions.UCXCanceled, match="Recv.*tag",
+            ):
+                msg = bytearray(1)
+                await ep.recv(msg)
+        await ep.close()
+
+    listener = ucp.create_listener(server_node,)
```
madsbk, line 64:
```suggestion
    listener = ucp.create_listener(server_node)
```

tests/test_endpoint.py
```diff
@@ -34,3 +34,32 @@ async def client_node(port):
...
+    ucp.get_ucx_version() < (1, 11, 0),
+    reason="Endpoint error handling is unreliable in UCX releases prior to 1.11.0",
+)
+@pytest.mark.parametrize("transfer_api", ["am", "tag"])
+async def test_cancel(transfer_api):
+    async def server_node(ep):
+        await ep.close()
+
+    async def client_node(port):
+        ep = await ucp.create_endpoint(ucp.get_address(), port,)
```
madsbk, line 50:
```suggestion
        ep = await ucp.create_endpoint(ucp.get_address(), port)
```

ucp/_libs/tests/test_cancel.py
```diff
@@ -0,0 +1,108 @@
...
+    # Keep endpoint to be used from outside the listener callback
+    ep = [None]
+
+    def _listener_handler(conn_request):
+        ep[0] = ucx_api.UCXEndpoint.create_from_conn_request(
+            worker, conn_request, endpoint_error_handling=True,
+        )
+
+    listener = ucx_api.UCXListener(worker=worker, port=0, cb_func=_listener_handler)
+    queue.put(listener.port),
```
madsbk, line 43:
```suggestion
    queue.put(listener.port)
```

madsbk (approved):
> I've added a couple of tests now, please let me know if you think they're satisfactory. It's difficult to test the exact behavior of UCXWorker.cancel_inflight_messages from the async API because of the lack of control to progress the worker, but I added a test for that in the core library.

Looks great @pentschev, only have some cosmetic suggestions.

pentschev:
Nice catches @madsbk , I'm actually surprised `black` didn't fix those, I often just ignore them in the expectation `black` will fix them, but it doesn't seem that it does. Thanks for the review and approval!

PR #814, opened 2021-11-16
Add new monitor line for nvlink and deprecate old one

quasiben:
cc @pentschev


pentschev (approved):
LGTM, thanks @quasiben .

PR #811, opened 2021-11-09
Update `RAPIDS_VERSION` var and `update-version.sh` script

ajschmidt8:
This PR updates the `RAPIDS_VERSION` variable and also includes a command in `update-version.sh` to ensure that it gets updated during releases.

PR #819, opened 2021-11-30
Revert `update-version.sh` changes

ajschmidt8:
I recently added a `sed` command (#811) to update the `RAPIDS_VERSION` in `ci/gpu/build.sh`. However, I now realize that the `NEXT_MAJOR` and `NEXT_MINOR` variables that I used correspond to the `ucx-py` versions (i.e. `0.22`) and not the RAPIDS versions (i.e. `22.12`). Therefore my changes would result in an incorrect string replacement.

I've commented out the line instead of removing it because we have some `update-version.sh` script improvements that are coming that will allow us to update both RAPIDS and `ucx-py` versions throughout the repositories.

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -30,4 +30,4 @@ function sed_runner() {
 }
 
 # cpp update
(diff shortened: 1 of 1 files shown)


AyodeAwe (approved):
Ok

pentschev (approved):
Thanks @ajschmidt8

PR #824, opened 2022-01-03
Set UCX_CUDA_COPY_MAX_REG_RATIO=1.0 when GPUs permit

pentschev:
This is necessary to maintain good performance on GPUs with large BAR1 size (greater or equal to the total GPU memory size). Starting with UCX 1.12 this value is set to 0.1 by default to ensure registration doesn't fail with GPUs that have a BAR1 size lower than the total GPU memory size (e.g., T4).

Currently we're setting to 1.0 when all GPUs have a large BAR1, in the future it may be useful to compute the real BAR1/TotalGPUMemory ratio to set this number. This is also a global setting, for fine-grained settings the application should take care of changing this value appropriately as UCX-Py doesn't know which GPUs are going to be used by the application at import time.

Additionally, setup logger before environment variables are set and log what's being set.

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -23,17 +23,23 @@ from .core import get_ucx_version  # noqa
 from .utils import get_ucxpy_logger  # noqa
 from ._libs.ucx_api import get_address  # noqa
 
+# Setup UCX-Py logger
+logger = get_ucxpy_logger()
+
+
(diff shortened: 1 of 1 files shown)


quasiben:
Thank you tracking this down @pentschev !

quasiben (approved):

pentschev:
rerun tests

quasiben:
Tested and confirmed this fixed a performance issue -- merging in. 

Thank you again @pentschev !

PR #825, opened 2022-01-05
xfail AM `test_from_worker_address_error` on UCX 1.11

pentschev:
Only TAG messages were covered by the xfail condition in `test_from_worker_address_error`, but AM should also xfail on UCX 1.11.


quasiben (approved):

quasiben:
Thanks @pentschev

PR #826, opened 2022-01-14
Set `UCX_MAX_RNDV_RAILS=1` on UCX >= 1.12.0 and update configurations docs

pentschev:
Add `UCX_MAX_RNDV_RAILS=1` to prevent crashes in UCX 1.12.0 caused by recent changes, more specifically when `UCX_CUDA_COPY_MAX_REG_RATIO=1.0` to maintain better performance on GPUs with large BAR1.

Add new subsection on UCX-Py default environment variables and further improve UCX-Py configuration docs.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -15,16 +15,41 @@ UCX/UCX-Py either with environment variables or programmatically during initiali
 .. note::
     When programmatically configuring UCX-Py, the ``UCX`` prefix is not used.
 
-For novice users we recommend the following settings:
+For novice users we recommend using UCX-Py defaults, see the next section for details.
+
+UCX-Py vs UCX Defaults
+----------------------
+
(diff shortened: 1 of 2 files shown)


quasiben:
This looks really good!  My only question is why remove the section on `UCX_TLS=all` ?

pentschev:
> This looks really good! My only question is why remove the section on `UCX_TLS=all` ?

Because that's the default. Best to leave to UCX actual default rather than telling the user to overwrite UCX's default with UCX's default. 😄

quasiben (approved):

pentschev:
rerun tests

jakirkham:
> > This looks really good! My only question is why remove the section on `UCX_TLS=all` ?
> 
> Because that's the default. Best to leave to UCX actual default rather than telling the user to overwrite UCX's default with UCX's default. 😄

It is pretty cool that we are that point that `UCX_TLS=all` is the default and we don't even need to mention it. Took a lot of hard work to get here 🙂

PR #829, opened 2022-01-18
Remove support for UCX < 1.11.1

pentschev:
About ~1000 fewer lines of fun!


quasiben:
Woot!

quasiben:
We'll need @rapidsai/admins to merge due to the README changes

pentschev:
Marking as draft while working on Dask-CUDA/Distributed PRs.

pentschev:
> We'll need @rapidsai/admins to merge due to the README changes

I think you mean @rapidsai/ops-codeowners , they only need to approve AFAIK, we can merge once they and @rapidsai/ucxpy-python-codeowners both approve then.

pentschev:
rerun tests

conda/environments/builddocs_py37.yml
```diff
@@ -16,6 +16,5 @@ dependencies:
 - pandoc=<2.0.0
 - pip
 - psutil
-- libhwloc
```
ajschmidt8, line 19: 1. Can this filename be updated? We don't currently support Python 3.7 any longer
2. If `libhwloc` is being removed here, can we also remove it from the `integration` repo below? It doesn't seem that any other RAPIDS libraries depend on it.

- https://github.com/rapidsai/integration/blob/887942b09b72e081da56ca057197d1b55d5780c6/conda/recipes/rapids-build-env/meta.yaml#L91
- https://github.com/search?q=org%3Arapidsai+libhwloc&type=code

pentschev: > 1. Can this filename be updated? We don't currently support Python 3.7 any longer

Filename updated in https://github.com/rapidsai/ucx-py/pull/829/commits/37c14fb540759d75cd9ea84cc95732320b39edb2 .

> 2. If `libhwloc` is being removed here, can we also remove it from the `integration` repo below? It doesn't seem that any other RAPIDS libraries depend on it.
> 
> * https://github.com/rapidsai/integration/blob/887942b09b72e081da56ca057197d1b55d5780c6/conda/recipes/rapids-build-env/meta.yaml#L91
> * https://github.com/search?q=org%3Arapidsai+libhwloc&type=code

Yes, it absolutely can be removed. I see a few other cases there, namely `cloud-ml-examples` and `node`, I'm not familiar with either of those repositories, but I believe the reason for them listing `libhwloc` was also due to the UCX-Py dependency. Should those be removed as well?

pentschev:
rerun tests

conda/environments/builddocs_py37.yml
ajschmidt8: > Yes, it absolutely can be removed. I see a few other cases there, namely cloud-ml-examples and node, I'm not familiar with either of those repositories, but I believe the reason for them listing libhwloc was also due to the UCX-Py dependency. Should those be removed as well?

I'm not sure if `cloud-ml-examples` is still maintained. I'll leave that up to you for those repos. But at the very least, we should remove it from the `integration` repo.

jakirkham:
Would be good to update these for 3.8 & 3.9 as well

https://github.com/rapidsai/ucx-py/blob/933b373a73d45f7d0a0b91043a6386b396eaebb8/setup.py#L112

Maybe we could change this latter one just to 3 so we don't need to keep updating it

https://github.com/rapidsai/ucx-py/blob/933b373a73d45f7d0a0b91043a6386b396eaebb8/setup.py#L131-L132

.readthedocs.yml
```diff
@@ -6,13 +6,13 @@ build:
...
 python:
-  version: 3.7
+  version: 3.8
   install:
     - method: setuptools
       path: .
 
 conda:
-  environment: conda/environments/builddocs_py37.yml
+  environment: conda/environments/builddocs_py38.yml
```
jakirkham, line 15: Maybe we can just rename to drop the Python version for simplicity?

```suggestion
  environment: conda/environments/builddocs.yml
```

conda/environments/builddocs_py37.yml
pentschev: Since I don't know anything about those projects, I'll leave them as is. Filed https://github.com/rapidsai/integration/pull/437 to remove `libhwloc` support from the integration repository.

pentschev:
> Would be good to update these for 3.8 & 3.9 as well

We have people using Python 3.6/3.7 still. We don't test it anymore, but it does still work and is needed by some projects so it's best to not completely block users from installing UCX-Py on older Python for now.

.readthedocs.yml
pentschev: Good idea, done in https://github.com/rapidsai/ucx-py/pull/829/commits/8cb2dd3c7ec446098eb21a942c85041321cc2a97 .

pentschev:
> Maybe we could change this latter one just to 3 so we don't need to keep updating it

Good idea as well, done in https://github.com/rapidsai/ucx-py/pull/829/commits/69afa101c77d25c488438772cfca25f80a8ddbdf .

jakirkham:
> > Would be good to update these for 3.8 & 3.9 as well
> 
> We have people using Python 3.6/3.7 still. We don't test it anymore, but it does still work and is needed by some projects so it's best to not completely block users from installing UCX-Py on older Python for now.

Interesting, who is still using UCX-Py on these older Python versions?

FWIW Python 3.6 is [EOL as of 2021-12-23]( https://www.python.org/dev/peps/pep-0494/#lifespan ). So it may be worth dropping that at least.

jakirkham:
Other than that LGTM

pentschev:
> Interesting, who is still using UCX-Py on these older Python versions?
> 
> FWIW Python 3.6 is [EOL as of 2021-12-23](https://www.python.org/dev/peps/pep-0494/#lifespan). So it may be worth dropping that at least.

I followed up with you on that offline. Anyway, this is change is not relevant for the problem being treated in this PR, I filed https://github.com/rapidsai/ucx-py/issues/834 to do that later on.

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Peter! 😄

PR #836, opened 2022-04-01
Add RMM managed memory argument to benchmarks

pentschev:
With https://github.com/openucx/ucx/pull/8057, UCX now supports CUDA IPC with CUDA managed memory through bounce buffers. This change adds an argument to benchmarks allowing selection of CUDA managed memory via RMM.

It is additionally necessary to remove the override of `UCX_RNDV_SCHEME=get_zcopy` as that won't work with bounce buffer.

Finally, separate setting CUDA device and instantiating allocator is also needed. The CUDA device must be set _before_ initializing UCX, but any memory allocations (such as an RMM pool) must be called _after_ initializing UCX. Therefore, CUDA device setting and allocator initialization must be split and reordered accordingly.


wence-:
> From UCX-Py it kinda is, but performance-wise it was really not great at all. From the time I was working on this, see performance below:

Confirmed with openucx/ucx@1c27876111ee77aa5cbf40083d163586b5353f31 that performance still suffers with managed memory on.

pentschev:
For completeness here, I had a brief chat with @Akshay-Venkatesh last week, he acknowledged that's still an issue and will check with his team when this could be addressed, but no ETA as of now.

pentschev:
I have finally figured out the issues with managed memory.

The first part is due to the ordering of CUDA tasks. We must create a CUDA context _before_ initializing UCX, but we must only initialize the RMM memory resource (and any CUDA allocations, for that matter) _after_ UCX is initialized. This involved a bit of refactoring that is now reflected in this PR.

The second part was due to `UCX_RNDV_SCHEME=get_zcopy` being set by default, also `UCX_RNDV_SCHEME=put_zcopy` won't work. In the past we had issues with schemes other than `get_zcopy`, so I'm attempting to remove the override so that UCX will just use its own default which is currently `UCX_RNDV_SCHEME=auto` and the bounce buffer for UVM works in that situation. Because of this change, I would rather leave it for RAPIDS 21.12 to avoid breaking anything just before the release.

cc @Akshay-Venkatesh for awareness of the issue above.

Updated results:

<details><summary>Without Managed Memory</summary>

```
$ UCX_RNDV_FRAG_MEM_TYPE=cuda python -m ucp.benchmarks.send_recv --backend ucp-core --reuse-alloc -b 0 -c 1 -d 0 -e 3 --n-iter 10 -n 1MiB -o rmm
Server Running at 10.33.225.163:48138
Client connecting to server at 10.33.225.163:48138
Roundtrip benchmark
================================================================================
Iterations                | 10
Bytes                     | 1.00 MiB
Object type               | rmm
Reuse allocation          | True
Transfer API              | TAG
Delay progress            | False
UCX_TLS                   | all
UCX_NET_DEVICES           | all
================================================================================
Device(s)                 | 0, 3
================================================================================
Bandwidth (average)       | 17.01 GiB/s
Bandwidth (median)        | 17.10 GiB/s
Latency (average)         | 57405 ns
Latency (median)          | 57098 ns
================================================================================
Iterations                | Bandwidth, Latency
--------------------------------------------------------------------------------
0                         | 17.14 GiB/s, 56977ns
1                         | 17.23 GiB/s, 56670ns
2                         | 16.39 GiB/s, 59594ns
3                         | 17.01 GiB/s, 57405ns
4                         | 17.05 GiB/s, 57263ns
5                         | 16.53 GiB/s, 59085ns
6                         | 17.07 GiB/s, 57218ns
7                         | 17.23 GiB/s, 56685ns
8                         | 17.22 GiB/s, 56708ns
```

<details><summary>With Managed Memory</summary>

```
$ UCX_RNDV_FRAG_MEM_TYPE=cuda python -m ucp.benchmarks.send_recv --backend ucp-core --reuse-alloc -b 0 -c 1 -d 0 -e 3 --n-iter 10 -n 1MiB -o rmm --rmm-managed-memory
Server Running at 10.33.225.163:33626
Client connecting to server at 10.33.225.163:33626
Roundtrip benchmark
================================================================================
Iterations                | 10
Bytes                     | 1.00 MiB
Object type               | rmm
Reuse allocation          | True
Transfer API              | TAG
Delay progress            | False
UCX_TLS                   | all
UCX_NET_DEVICES           | all
================================================================================
Device(s)                 | 0, 3
================================================================================
Bandwidth (average)       | 10.14 GiB/s
Bandwidth (median)        | 10.45 GiB/s
Latency (average)         | 96285 ns
Latency (median)          | 93424 ns
================================================================================
Iterations                | Bandwidth, Latency
--------------------------------------------------------------------------------
0                         | 10.46 GiB/s, 93377ns
1                         | 10.45 GiB/s, 93447ns
2                         | 8.70 GiB/s, 112300ns
3                         | 9.48 GiB/s, 103014ns
4                         | 10.46 GiB/s, 93401ns
5                         | 10.41 GiB/s, 93779ns
6                         | 10.50 GiB/s, 93025ns
7                         | 10.51 GiB/s, 92897ns
8                         | 10.36 GiB/s, 94242ns
9                         | 10.46 GiB/s, 93366ns
```

pentschev:
It seems like CI isn't happy with `UCX_RNDV_SCHEME=auto`, I'll have to check tomorrow what can be done there.

benchmarks/send-recv-core.py
```diff
@@ -0,0 +1,541 @@
...
+    if args.object_type == "numpy":
+        import numpy as xp
+    elif args.object_type == "cupy":
+        import cupy as xp
+
+        xp.cuda.runtime.setDevice(args.client_dev)
+    else:
+        import cupy as xp
+
+        import rmm
```
madsbk, line 218: Could the imports go to the top of the file?

ucp/benchmarks/utils.py
```diff
@@ -22,6 +22,43 @@ def _ensure_cuda_device(devs, rank):
...
+        otherwise.
+    """
+    if object_type == "numpy":
+        import numpy as xp
+    elif object_type == "cupy":
+        import cupy as xp
+    else:
+        import cupy as xp
+
+        import rmm
```
madsbk, line 50: ditto

madsbk (commented):
LGTM, if you can get CI to work with `UCX_RNDV_SCHEME=auto`

ucp/benchmarks/utils.py
pentschev: They could, but we need to test for their availability anyway, for example, a UCX-Py user may not have any CUDA libraries and importing `cupy` or `rmm` would then fail, or just `rmm` may fail if it's not a RAPIDS user. Therefore, I would have to import at the top and still handle the cases here when the import was unsuccessful, which will probably making the code even more complex, so I would prefer to keep them like this. What do you think?

benchmarks/send-recv-core.py
pentschev: Sorry, this file is here by accident, it's the old file that was previously removed after the benchmarks were refactored, but I accidentally reintroduced it after merging the new branch due to my original changes to this file in the PR. This has been removed again in https://github.com/rapidsai/ucx-py/pull/836/commits/0fae0b696f33fe66217cf3d40735de166849a0e0 .

pentschev:
We must now use `UCX_RNDV_FRAG_MEM_TYPE=cuda`, which is done in https://github.com/rapidsai/ucx-py/pull/836/commits/674b43dbb8e01f334ee5d2872c481c0d4ab1b711 . That resolves previous CI issues.

Could you take another look when you have a chance @madsbk ?

ucp/benchmarks/utils.py
```diff
@@ -22,6 +22,43 @@ def _ensure_cuda_device(devs, rank):
...
+    object_type: str
+        The type of object the allocator should return. Options are: "numpy", "cupy"
+        or "rmm".
+    rmm_init_pool_size: int
+        If the object type is "rmm" (implies usage of RMM pool), define the initial
+        pool size.
+    rmm_managed_memory: bool
+        If the object type is "rmm", use managed memory if `True`, or default memory
+        otherwise.
+    """
```
wence-, line 42:
```suggestion
    Returns
    -------
    A handle to a module, one of ``numpy`` or ``cupy`` (if device memory is requested). 
    If the object type is ``rmm``, then ``cupy`` is configured to use RMM as an allocator.
    """
```

?

ucp/__init__.py
```diff
@@ -34,9 +34,9 @@
     logger.info("Setting UCX_RNDV_THRESH=8192")
     os.environ["UCX_RNDV_THRESH"] = "8192"
 
-if "UCX_RNDV_SCHEME" not in os.environ:
-    logger.info("Setting UCX_RNDV_SCHEME=get_zcopy")
-    os.environ["UCX_RNDV_SCHEME"] = "get_zcopy"
+if "UCX_RNDV_FRAG_MEM_TYPE" not in os.environ:
+    logger.info("Setting UCX_RNDV_FRAG_MEM_TYPE=cuda")
+    os.environ["UCX_RNDV_FRAG_MEM_TYPE"] = "cuda"
```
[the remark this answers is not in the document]
pentschev: Running the tests we have on a Docker container with no CUDA support seems to behave safely, so I guess we're good and probably will find out soon if there are any issues with it, thus why we should leave this PR for 21.12.

ucp/__init__.py
```diff
@@ -34,9 +34,9 @@
     logger.info("Setting UCX_RNDV_THRESH=8192")
     os.environ["UCX_RNDV_THRESH"] = "8192"
 
-if "UCX_RNDV_SCHEME" not in os.environ:
-    logger.info("Setting UCX_RNDV_SCHEME=get_zcopy")
-    os.environ["UCX_RNDV_SCHEME"] = "get_zcopy"
```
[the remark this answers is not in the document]
pentschev: Yup, that is my understanding too, selecting `auto` should be good enough. But this is another change that's better left for 21.12, which will give us time to understand whether there were any regressions.

ucp/benchmarks/backends/tornado.py
```diff
@@ -82,27 +80,23 @@ async def run(self) -> bool:
...
 
             await stream.write(send_msg.data)
             await stream.read_into(recv_msg.data)
 
             stop = monotonic()
             if i >= self.args.n_warmup_iter:
                 times.append(stop - start)
-        if self.args.cuda_profile:
-            self.xp.cuda.profiler.stop()
         self.queue.put(times)
```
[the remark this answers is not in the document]
pentschev: That's right, it originally made it there by accident.

ucp/benchmarks/utils.py
```diff
@@ -22,6 +22,43 @@ def _ensure_cuda_device(devs, rank):
     numba.cuda.current_context()
 
 
+def get_allocator(
+    object_type: str, rmm_init_pool_size: int, rmm_managed_memory: bool
+) -> object:
```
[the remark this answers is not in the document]
pentschev: Yes, `types.ModuleType` is exactly what I wanted but have overlooked. I fixed it now in https://github.com/rapidsai/ucx-py/pull/836/commits/f2b955cea6bcd5b15d29e43ea89d32d142c7e5ca .

wence-:
From f6fd8c80acba9608d03821230b8eccdc243c80fc

> The CUDA device must be set _before_ initializing UCX, but any memory allocations (such as an RMM pool) must be called _after_ initializing UCX. Therefore, splitting CUDA device setting and allocator initialization must be split and reorder accordingly.

This information seems important enough that we should document it (and possibly make it easy for users to achieve it): if you can point at where in the code we initialize UCX I can have a go at splitting things and writing up some docs.

I guess this requirement is generically true so that UCX has a chance to insert its malloc hooks in cudaMalloc, etc...

pentschev:
> This information seems important enough that we should document it (and possibly make it easy for users to achieve it): if you can point at where in the code we initialize UCX I can have a go at splitting things and writing up some docs.

This is a good idea. We initialize UCX upon the first call we make, technically it happens in [`ucp.init()`](https://github.com/rapidsai/ucx-py/blob/4a64c892dfb15c13a5d28c51a2a1852cbd54a1db/ucp/core.py#L887-L926), but unless the user is attempting to override UCX context defaults, this is called implicitly when calling [any other `ucp` functions](https://github.com/rapidsai/ucx-py/blob/4a64c892dfb15c13a5d28c51a2a1852cbd54a1db/ucp/core.py#L963-L1083) that call `_get_ctx()` as well, for example when creating a listener or endpoint.

> I guess this requirement is generically true so that UCX has a chance to insert its malloc hooks in cudaMalloc, etc...

That is correct. For reference, in Dask we do this before anything else in [`distributed.comm.ucx.init_once`](https://github.com/dask/distributed/blob/37fd8b501a4b0fe5c528fb394513b35e0d4ead3c/distributed/comm/ucx.py#L94-L123), and comms in Dask happens before preload plugins, like those [plugins we have in Dask-CUDA](https://github.com/rapidsai/dask-cuda/blob/23460f6f7300202a2d3a25ffedf6884755fe6fbc/dask_cuda/cuda_worker.py#L215-L228) to initialize, among other things, RMM.

wence- (approved):

raydouglass (approved):

pentschev:
Thanks @wence- and @madsbk for reviews and @raydouglass for approving. This is now in.

PR #838, opened 2022-04-13
Ignore MIG instances when overriding `UCX_CUDA_COPY_MAX_REG_RATIO`

pentschev:
Importing UCX-Py currently fails when trying to query memory for memory with PyNVML. Since it's not clear and not tested what would be safe `UCX_CUDA_COPY_MAX_REG_RATIO` values for MIG instances, just skip them and rely on UCX's default for now.


pentschev:
rerun tests

jakirkham (approved):

PR #844, opened 2022-04-29
Use patch.dict to modify environment state

### LINKED ISSUE #842: Testsuite statefully modifies `os.environ`
A number of tests in the testsuite modify `os.environ` to control some aspect of UCX for the purpose of a specific test. They should restore the original value afterwards in case of potential bad interactions with later tests. This could be done by hand in each test that modifies the environment, but this is prone to being missed when new tests are introduced.

I would therefore suggest to save a copy of the environment in `pytest_setup` and then restore it in `pytest_teardown` in `conftest.py`: if this is not too magic, I will take this approach.

wence-:
To avoid changing the environment for later tests, patch the os.environ
dict and restore it after test exit. Fixes #842.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,4 +1,5 @@
 import os
+from unittest.mock import patch
 
 import pytest
 from utils import captured_logger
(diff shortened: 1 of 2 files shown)


pentschev (approved):
Looks real good, thanks @wence- !

PR #845, opened 2022-04-29
Fix environment override logic in ucp.init

wence-:
In the config dict, UCX variables are not prefixed with UCX_, so we must
take that into account when comparing to the environment.

While we're here, take a copy of the user-provided config dict so that
we don't modify it behind their back.

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -909,11 +909,20 @@ def init(options={}, env_takes_precedence=False, blocking_progress_mode=None):
             "UCX is already initiated. Call reset() and init() "
             "in order to re-initate UCX with new options."
         )
-    if env_takes_precedence:
-        options = options.copy()
-        for k, v in options.items():
-            options[k] = os.environ.get(f"UCX_{k}", v)
-
+    options = options.copy()
+    for k, v in options.items():
+        env_k = f"UCX_{k}"
+        env_v = os.environ.get(env_k)
+        if env_v is not None:
+            if env_takes_precedence:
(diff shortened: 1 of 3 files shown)


wence-:
This is based on #844 so that probably wants to merge first and then I can rebase.

pentschev:
#844 is in now.

ucp/core.py
```diff
@@ -910,9 +910,9 @@ def init(options={}, env_takes_precedence=False, blocking_progress_mode=None):
             "in order to re-initate UCX with new options."
         )
     if env_takes_precedence:
-        for k in os.environ.keys():
-            if k in options:
-                del options[k]
+        options = options.copy()
+        for k, v in options.items():
+            options[k] = os.environ.get(f"UCX_{k}", v)
```
pentschev, line 915: Maybe we should add a warning or a `logger.debug` entry if both are set and have values?

pentschev (commented):
Looks good, I have only one suggestion, what do you think?

ucp/core.py
wence-: Not a warning, I think, because IME chatty warnings that always appear are swiftly ignored by end users.

Logging at debug level is reasonable. I think then it should occur in both branches (i.e. log whenever the environment and config overlap irrespective of which is shadowing).

wence-: Done.

pentschev (approved):
LGTM, thanks @wence- !

PR #851, opened 2022-05-17
Add dockerfile using librdmacm to provide infiniband interfaces

### LINKED ISSUE #846: Improve Dockerfile for cluster deployment
With the great work from @wence- in https://github.com/rapidsai/ucx-py/pull/840, we now have a reference Dockerfile + run script that supports IB and CUDA IPC and runs tests and benchmarks. This is a great way for users to get started with UCX-Py deployment and test that the system is in good state.

Besides that, we could think getting that recipe even more useful. I imagine we could make that more general to replace recipes that people have out there, for example, I know @ayushdg has one that's being used for deployment in some clusters (which we actually copied from for building MOFED in https://github.com/rapidsai/ucx-py/pull/840), I think @quasiben also had some in the past without MOFED.

The idea here is to get information on how we can make this useful for different needs, which people can hopefully just copy and do minimal changes to get started. We would then maintain the Dockerfile and keep it always up-to-date, releasing users from figuring out what's new and what they should update their current environments.

@ayushdg @beckernick @VibhuJawa @randerzander @quasiben when you have a chance, could you please let us know if this is something that would be generally useful? I think the best way to do so would be reading https://github.com/rapidsai/ucx-py/tree/branch-0.26/docker#infinibandnvlink-enabled-docker-file and checking files referred there to see what we already have and what could be improved.

In the future we may have it all within RAPIDS docker containers, but it's not clear when this will happen as we still need to figure out `rdma-core>=28` as a conda-forge package, and we also want to be mindful of UCX-Py users that may want to use it without RAPIDS.

wence-:
Rather than relying on exact matching MOFED versions, use rdma-core
which is forward-compatible with MOFED version >= 5.x.

To facilitate downstream users adopting the build setup, split the
building and installing of UCX and UCX-Py into scripted steps for which
the conda environment name and UCX version to use can be configured.
Adapt the existing MOFED-based docker file to use this setup too.

Closes #846.

- [x] Update readme on the different docker file versions
- [x] Test on more than a single host MOFED version.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Contains reference dockerfile and build script to run UCX-Py tests and benchmarks. This is a minimal setup, without support for CUDA, MOFED or rdma-core.
+Contains reference dockerfile and build script to run UCX-Py tests and benchmarks. This is a minimal setup, without support for CUDA, MOFED, or rdma-core.
 
 ## Building Docker image
 
@@ -26,18 +26,18 @@ The container above will run UCX-Py tests and benchmarks.
 ## Infiniband/NVLink-enabled docker file
 
 In addition to the reference Docker image, there are two further docker
-files which (respectively) have support for CUDA devices and
-infiniband/nvlink-enabled communications using either
+files which have support for CUDA devices and
+InfiniBand/NVLink-enabled communications using either
 [rdma-core](https://github.com/linux-rdma/rdma-core) or
 [MOFED](https://network.nvidia.com/products/infiniband-drivers/linux/mlnx_ofed/).
-image with support for CUDA and MOFED. In both cases, the default base image is
-[nvidia/cuda:11.5.2-devel-ubuntu20.04](https://hub.docker.com/layers/cuda/nvidia/cuda/11.5.2-devel-ubuntu20.04/images/sha256-fed73168f35a44f5ff53d06d61a1c55da7c26e7ca5a543efd78f35d98f29fd4a?context=explore).
+In both cases, the default base image is
+[nvidia/cuda:11.5.2-devel-ubuntu20.04](https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=11.5.2-devel-ubuntu20.04).
 
-The rdma-core image should work as long as the host system has MOFED >= 5.x.
+The rdma-core image should work as long as the host system has MOFED >= 5.0.
 If you use the MOFED image, then the host version (reported by `ofed_info
 -s`) should match that used when building the container.
 
-To use these images, first build it
+To use one of these images, first build it
 ```bash
 docker build -t ucx-py-mofed -f UCXPy-MOFED.dockerfile .
 # or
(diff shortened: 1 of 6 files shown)


wence-:
I still need to test on more than one host MOFED version, but this setup seems to work.

Refactored the dockerfile to hopefully make it easier to use downstream, cc @ayushdg and @jacobtomlinson.

wence-:
This works well on both MOFED 5.3 and 5.2 on systems with A100 and V100 respectively (same performance as version-matched MOFED in the container).

I've updated the readme description as well.

docker/README.md
```diff
@@ -25,17 +25,37 @@ The container above will run UCX-Py tests and benchmarks.
...
-image, and uses MOFED version 5.3-1.0.5.0. To function successfully the OFED
-version reported by `ofed_info -s` on the host should match this version.
-
-To use this image, first build it
+In addition to the reference Docker image, there are two further docker
+files which (respectively) have support for CUDA devices and
+infiniband/nvlink-enabled communications using either
+[rdma-core](https://github.com/linux-rdma/rdma-core) or
+[MOFED](https://network.nvidia.com/products/infiniband-drivers/linux/mlnx_ofed/).
+image with support for CUDA and MOFED. In both cases, the default base image is
```
pentschev, line 33:
```suggestion
In both cases, the default base image is
```

docker/README.md
```diff
@@ -25,17 +25,37 @@ The container above will run UCX-Py tests and benchmarks.
...
+[nvidia/cuda:11.5.2-devel-ubuntu20.04](https://hub.docker.com/layers/cuda/nvidia/cuda/11.5.2-devel-ubuntu20.04/images/sha256-fed73168f35a44f5ff53d06d61a1c55da7c26e7ca5a543efd78f35d98f29fd4a?context=explore).
+
+The rdma-core image should work as long as the host system has MOFED >= 5.x.
```
pentschev, line 36:
```suggestion
The rdma-core image should work as long as the host system has MOFED >= 5.0.
```

docker/UCXPy-MOFED.dockerfile
```diff
@@ -4,18 +4,37 @@ FROM nvidia/cuda:${CUDA_VERSION}-devel-${DISTRIBUTION_VERSION}
...
+ARG CONDA_ENV_SPEC=ucx-py-cuda11.5.yml
+
+ENV CONDA_ENV="${CONDA_ENV}"
+ENV CONDA_HOME="${CONDA_HOME}"
 
+# Where cuda is installed
 ENV CUDA_HOME="/usr/local/cuda"
-ENV PATH="${CUDA_HOME}/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+
+# This can go once the nvidia images have updated the signing key?
```
pentschev, line 24: Not quite sure what to make of this comment. Are current signing keys invalid?

pentschev (changes requested):
This is looking amazing, thanks so much for the work @wence- ! I've requested a few changes and made some suggestions only.

docker/UCXPy-MOFED.dockerfile
wence-: https://developer.nvidia.com/blog/updating-the-cuda-linux-gpg-repository-key/ but I realise this is actually not needed (I had forgotten to `docker pull` the base image before testing).

docker/build-ucx.sh
```diff
@@ -0,0 +1,25 @@
...
+source ${CONDA_HOME}/etc/profile.d/conda.sh
+conda activate ${CONDA_ENV}
+
+git clone https://github.com/openucx/ucx.git
+
+pushd ucx
+git checkout ${UCX_VERSION_TAG}
+./autogen.sh
+mkdir build-linux
+pushd build-linux
```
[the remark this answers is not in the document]
wence-: The pushes were kind of unnecessary (I just copied from my terminal transcript). I've moved them into `cd` s (and reduced the number).

docker/UCXPy-MOFED.dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# will be used to build the conda environment.
+ARG CONDA_ENV_SPEC=ucx-py-cuda11.5.yml
+
+ENV CONDA_ENV="${CONDA_ENV}"
+ENV CONDA_HOME="${CONDA_HOME}"
+
+# Where cuda is installed
+ENV CUDA_HOME="/usr/local/cuda"
+
+ADD https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh /miniconda.sh
```
jacobtomlinson, line 24: `ADD` does a lot of magic and is generally discouraged these days. It would probably be preferable to curl this instead as part of the same `RUN` that calls and removes it.

docker/UCXPy-MOFED.dockerfile
```diff
@@ -0,0 +1,64 @@
...
+        libnuma-dev \
+        libtool \
+        make \
+        pkg-config \
+        udev \
+        wget \
+    && apt-get autoremove -y \
+    && apt-get clean
+
+ADD https://content.mellanox.com/ofed/MLNX_OFED-${MOFED_VERSION}/MLNX_OFED_LINUX-${MOFED_VERSION}-${DISTRIBUTION_VERSION}-x86_64.tgz /MLNX_OFED_LINUX-${MOFED_VERSION}-${DISTRIBUTION_VERSION}-x86_64.tgz
```
jacobtomlinson, line 48: Same here. I think in this case `ADD` is also extracting these files for you so the next `RUN` command might be redundant.

Again it's probably best to `curl`, `tar xzf`, etc all in one command.

jacobtomlinson (commented):
This looks great!

I left a few comments about the use of `ADD`. I see a lot of folks discouraging the use of it these days because it does a lot of magic under the hood that can catch folks by surprise and creates unnecessary layers.

Best practice is to use `COPY` or `RUN curl ...`. Explicit is better than implicit.

jacobtomlinson (approved):

jacobtomlinson:
Thanks for the update here @wence-. Just a heads up force-pushing here made it hard for me to see what changes you made after the review. It's common for us to squash merge in RAPIDS and Dask anyway so I wouldn't worry about squashing yourself.

pentschev (approved):
Thanks so much @wence- for the work here, and @jacobtomlinson for reviewing as well.

wence-:
> Just a heads up force-pushing here made it hard for me to see what changes you made after the review.

Apologies. FWIW I only squashed commits after your previous review (but I can see that it is painful).

wence-:
wrt performance, I've tested both the basic ucx-py benchmarks and dask-cuda merge benchmarks on baremetal, rdma-core (mofed 5.2 and 5.30 on the host), and mofed version-matched containers, and the results look the same. So I think I am happy that the performance will not suffer in the container.

pentschev:
rerun tests

PR #843, opened 2022-04-29
Add conda recipe

jakirkham:
Includes a Conda recipe for building UCX-Py as part of the CI here. Also configures uploads so that happen as part of the build process here.

Issue #797, opened 2021-10-19
Possible flaky AM test


pentschev:
I've seen `tests/test_send_recv_am.py::test_send_recv_bytes` fail at least a couple of times, [this is the latest failure](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-gpu-build/detail/ucx-py-gpu-build/3313/pipeline/) from https://github.com/rapidsai/ucx-py/pull/796:

<details>

```python
__________________ test_send_recv_bytes[data2-True-True-256] ___________________

size = 256, blocking_progress_mode = True, recv_wait = True
data = {'allocator': functools.partial(<function ones at 0x7fa87acfa830>, dtype=<class 'numpy.uint8'>), 'generator': functool...pe=<class 'numpy.int64'>), 'memory_type': 'cuda', 'validator': <function get_data.<locals>.<lambda> at 0x7fa7ddf98320>}

    @pytest.mark.skipif(
        not ucp._libs.ucx_api.is_am_supported(), reason="AM only supported in UCX >= 1.11"
    )
    @pytest.mark.asyncio
    @pytest.mark.parametrize("size", msg_sizes)
    @pytest.mark.parametrize("blocking_progress_mode", [True, False])
    @pytest.mark.parametrize("recv_wait", [True, False])
    @pytest.mark.parametrize("data", get_data())
    async def test_send_recv_bytes(size, blocking_progress_mode, recv_wait, data):
        rndv_thresh = 8192
        ucp.init(
            options={"RNDV_THRESH": str(rndv_thresh)},
            blocking_progress_mode=blocking_progress_mode,
        )
    
        ucp.register_am_allocator(data["allocator"], data["memory_type"])
        msg = data["generator"](size)
    
        recv = []
        listener = ucp.create_listener(simple_server(size, recv))
        num_clients = 1
        clients = [
            await ucp.create_endpoint(ucp.get_address(), listener.port)
            for i in range(num_clients)
        ]
        for c in clients:
            if recv_wait:
                # By sleeping here we ensure that the listener's
                # ep.am_recv call will have to wait, rather than return
                # immediately as receive data is already available.
                await asyncio.sleep(1)
            await c.am_send(msg)
        for c in clients:
            await c.close()
        listener.close()
    
        if data["memory_type"] == "cuda" and msg.nbytes < rndv_thresh:
            # Eager messages are always received on the host, if no host
[...]

PR #853, opened 2022-05-23
Add `VERSION_SUFFIX` to nightly packages

charlesbluca:
Looks like with #843 we removed the version suffix from the published nightly packages - this PR adds this back in.

### DIFF
--- a/ci/cpu/build.sh
+++ b/ci/cpu/build.sh
@@ -15,6 +15,11 @@ export HOME=$WORKSPACE
 # Switch to project root; also root of repo checkout
 cd $WORKSPACE
 
+# If nightly build, append current YYMMDD to version
+if [[ "$BUILD_MODE" = "branch" && "$SOURCE_BRANCH" = branch-* ]] ; then
+  export VERSION_SUFFIX=`date +%y%m%d`
+fi
+
 # Get latest tag and number of commits since tag
 export GIT_DESCRIBE_TAG=`git describe --abbrev=0 --tags`
 export GIT_DESCRIBE_NUMBER=`git rev-list ${GIT_DESCRIBE_TAG}..HEAD --count`


ci/cpu/build.sh
```diff
@@ -15,6 +15,11 @@ export HOME=$WORKSPACE
...
 cd $WORKSPACE
 
+# If nightly build, append current YYMMDD to version
+if [[ "$BUILD_MODE" = "branch" && "$SOURCE_BRANCH" = branch-* ]] ; then
+  export VERSION_SUFFIX=`date +%y%m%d`
+fi
+
 # Get latest tag and number of commits since tag
 export GIT_DESCRIBE_TAG=`git describe --abbrev=0 --tags`
 export GIT_DESCRIBE_NUMBER=`git rev-list ${GIT_DESCRIBE_TAG}..HEAD --count`
```
charlesbluca, line 25: Unrelated question, do we need to explicitly export these? I was under the impression that these variables are already populated for conda builds of git repos.

ci/cpu/build.sh
```diff
@@ -15,6 +15,11 @@ export HOME=$WORKSPACE
 # Switch to project root; also root of repo checkout
 cd $WORKSPACE
 
+# If nightly build, append current YYMMDD to version
+if [[ "$BUILD_MODE" = "branch" && "$SOURCE_BRANCH" = branch-* ]] ; then
+  export VERSION_SUFFIX=`date +%y%m%d`
+fi
```
jakirkham, line 21: Analogous to [this code in cuDF]( https://github.com/rapidsai/cudf/blob/b926f51dde17db1c591fa5d25c41a9f6b989233b/ci/cpu/build.sh#L32-L35 )

`VERSION_SUFFIX` is already in the recipe version (see below). It just wasn't being set before

https://github.com/rapidsai/ucx-py/blob/0caf1f2abf5475777dd3fc962e81697eec9f180a/conda/recipes/ucx-py/meta.yaml#L1

Ethyling (approved):

raydouglass (approved):

ci/cpu/build.sh
jakirkham: Yeah both of these are set by Conda. Some repos still seem to [set these]( https://github.com/rapidsai/rapids-cmake/blob/9a1beebf9966145615f3b13f9bf3452ba580d5c4/ci/gpu/build.sh#L26-L27 ). Perhaps we should raise a new thread to discuss this)

jakirkham (approved):

jakirkham:
rerun tests

Note: CI failure appears to be due to a flaky test. Mentioned that in comment ( https://github.com/rapidsai/ucx-py/issues/797#issuecomment-1134978403 )

pentschev (approved):

PR #859, opened 2022-05-30
Use mamba rather than conda in docker files

wence-:
This significantly speeds up building images, and makes installing
packages in the image much more pleasant, since the solver is so much
faster.

This is low-priority, but a reasonable quality of life improvement when building the containers regularly.


pentschev (approved):
This is very nice, I actually didn't know there was a "Mambaforge" installer. Thanks @wence- !

PR #863, opened 2022-06-23
Improve benchmarks results output and features

pentschev:
The following improvements are introduced:

- More consistency formatting of results output
- Latency results
- `--n-warmup-iter` argument
- `--no-detailed-report` argument to `send-recv.py`
- Remove Dask dependency on benchmarks when unavailable

### DIFF
--- a/benchmarks/send-recv-core.py
+++ b/benchmarks/send-recv-core.py
@@ -26,7 +26,12 @@ from time import perf_counter as clock
 import ucp
 from ucp._libs import ucx_api
 from ucp._libs.arr import Array
-from ucp._libs.utils import format_bytes, parse_bytes
+from ucp._libs.utils import (
+    format_bytes,
+    parse_bytes,
+    print_key_value,
+    print_separator,
+)
 from ucp._libs.utils_test import (
     blocking_am_recv,
     blocking_am_send,
@@ -39,14 +44,6 @@ from ucp._libs.utils_test import (
 mp = mp.get_context("spawn")
 
 WireupMessage = bytearray(b"wireup")
-
-
-def print_separator(separator="-", length=80):
-    print(separator * length)
-
-
-def print_key_value(key, value, key_length=25):
-    print(f"{key: <{key_length}} | {value}")
 
 
 def register_am_allocators(args, worker):
--- a/benchmarks/send-recv.py
+++ b/benchmarks/send-recv.py
@@ -24,17 +24,14 @@ import os
 from time import perf_counter as clock
 
 import ucp
-from ucp._libs.utils import format_bytes, parse_bytes
+from ucp._libs.utils import (
+    format_bytes,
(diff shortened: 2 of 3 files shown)


benchmarks/send-recv-core.py
```diff
@@ -293,32 +302,39 @@ def op_completed():
...
             worker.progress()
 
     if args.cuda_profile:
         xp.cuda.profiler.stop()
 
     assert len(times) == args.n_iter
+    bw_avg = format_bytes(2 * args.n_iter * args.n_bytes / sum(times))
+    bw_med = format_bytes(2 * args.n_bytes / np.median(times))
+    lat_avg = int(sum(times) * 1e9 / (2 * args.n_iter))
+    lat_med = int(np.median(times) * 1e9 / 2)
```
wence-, line 319: Could use `dask.utils.format_time` here instead.

benchmarks/send-recv-core.py
```diff
@@ -329,22 +345,23 @@ def op_completed():
...
-        print("--------------------------")
+        print_separator(separator="=")
+        print_key_value(key="Iterations", value="Bandwidth, Latency")
+        print_separator(separator="-")
         for i, t in enumerate(times):
             ts = format_bytes(2 * args.n_bytes / t)
-            ts = (" " * (9 - len(ts))) + ts
-            print("%03d         |%s/s" % (i, ts))
+            lat = int(t * 1e9 / 2)
+            print_key_value(key=i, value=f"{ts}/s, {lat}ns")
```
wence-, line 364: `dask.utils.format_time` again.

wence- (changes requested):
Overall this looks good with the nitpick to do with `format_time`

benchmarks/send-recv-core.py
pentschev: We can't have Dask dependencies in UCX-Py, it is used by projects who do not use Dask.

benchmarks/send-recv-core.py
pentschev: Ditto.

benchmarks/send-recv-core.py
wence-: The `format_bytes` call needs to go then (since that comes from `dask.utils` AFAICS)

pentschev: You're right, we need to remove that. I'll work on that later today or next week.

pentschev: This is done now in https://github.com/rapidsai/ucx-py/pull/863/commits/8741b275d2c69cec2b27289bfa819cea2301fdff .

wence- (approved):

wence-:
Looks good.

pentschev:
Thanks for the review @wence- !

quasiben (approved):

PR #864, opened 2022-07-06
Add `get_event_loop()` util function

pentschev:
Since `asyncio.get_event_loop()` is deprecated starting on Python 3.10 and will eventually become an alias to `asyncio.get_running_loop()`, implement a new function to emulate the old behavior to be used where appropriate.

In https://github.com/dask/distributed/pull/6231, some of the fixtures were updated to match `asyncio`'s updates, which also caused some problems with the event loop being created in `ucp.ApplicationContext`, which can be seen in [this log](https://raw.githack.com/pentschev/ucx-py-ci/test-results/assets/ucx-master-dask-cuda-202207052200.html#0161d81a-ac60-4f1b-87ac-9b4bde5931e5) of nightly runs on a DGX-1, this fixes that.

### DIFF
--- a/tests/test_multiple_nodes.py
+++ b/tests/test_multiple_nodes.py
@@ -4,7 +4,6 @@ import numpy as np
 import pytest
 
 import ucp
-from ucp.utils import get_event_loop
 
 
 def get_somaxconn():
@@ -52,4 +51,4 @@ async def test_many_servers_many_clients(num_servers, num_clients):
         clients = []
         for __ in range(i, min(i + somaxconn, num_clients * num_servers)):
             clients.append(client_node(listeners[__ % num_servers].port))
-        await asyncio.gather(*clients, loop=get_event_loop())
+        await asyncio.gather(*clients)
--- a/tests/test_multiple_nodes.py
+++ b/tests/test_multiple_nodes.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 
 import ucp
+from ucp.utils import get_event_loop
 
 
 def get_somaxconn():
@@ -51,4 +52,4 @@ async def test_many_servers_many_clients(num_servers, num_clients):
         clients = []
         for __ in range(i, min(i + somaxconn, num_clients * num_servers)):
             clients.append(client_node(listeners[__ % num_servers].port))
-        await asyncio.gather(*clients, loop=asyncio.get_event_loop())
+        await asyncio.gather(*clients, loop=get_event_loop())
--- a/ucp/comm.py
+++ b/ucp/comm.py
@@ -23,7 +23,7 @@ def _call_ucx_api(event_loop, func, *args, **kwargs):
     Basically, all the communication functions have the
     same structure, which this wrapper implements.
     """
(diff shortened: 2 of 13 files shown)


tests/test_multiple_nodes.py
```diff
@@ -51,4 +52,4 @@ async def test_many_servers_many_clients(num_servers, num_clients):
         clients = []
         for __ in range(i, min(i + somaxconn, num_clients * num_servers)):
             clients.append(client_node(listeners[__ % num_servers].port))
-        await asyncio.gather(*clients, loop=asyncio.get_event_loop())
+        await asyncio.gather(*clients, loop=get_event_loop())
```
wence-, line 55: Python 3.10 doesn't have the loop keyword argument to `asyncio.gather` (and it is deprecated since 3.8, so probably just remove this keyword argument). The relevant event loop is grabbed from the coroutines/futures being passed in.

For coroutines, the currently running event loop is used to run the coroutines. For futures, the loop that created the future is used to run things (but all futures must come from the same loop).

So perhaps:

```
from itertools import cycle, islice

...
    somaxconn = get_somaxconn()

    listeners = cycle((ucp.create_listener(server_node) for _ in range(num_servers)))

    # We ensure no more than `somaxconn` connections are submitted
    # at once. Doing otherwise can block and hang indefinitely.
    for _ in range(0, num_clients * num_servers, somaxconn):
        await asyncio.gather(
            *(client_node(listener.port) for listener in islice(listeners, somaxconn))
        )
```

ucp/comm.py
```diff
@@ -110,7 +111,7 @@ def am_recv(
     event_loop=None,
 ) -> asyncio.Future:
 
-    event_loop = event_loop if event_loop else asyncio.get_event_loop()
+    event_loop = event_loop if event_loop else get_event_loop()
```
wence-, line 114: Style nitpick `event_loop or get_event_loop()`

ucp/core.py
```diff
@@ -397,9 +397,9 @@ def continuous_ucx_progress(self, event_loop=None):
         ----------
         event_loop: asyncio.event_loop, optional
             The event loop to evoke UCX progress. If None,
-            `asyncio.get_event_loop()` is used.
+            `ucp.utils.get_event_loop()` is used.
         """
-        loop = event_loop if event_loop is not None else asyncio.get_event_loop()
+        loop = event_loop if event_loop is not None else get_event_loop()
```
wence-, line 402: Slightly different style to the previous code, but probably `event_loop or get_event_loop()` here too.

tests/test_multiple_nodes.py
wence-: (This is the only call to `asyncio.gather` that uses the loop keyword argument in the codebase, so no need to fix anywhere else)

wence- (changes requested):
Minor comments (feel free to ignore the style ones), and one issue with deprecated keyword arg to asyncio.gather

wence- (approved):

wence-:
Thanks!

PR #867, opened 2022-08-04
Update cuDF merge benchmark

pentschev:
This adds several bugfixes and improvements:

- Correctly assign `CUDA_VISIBLE_DEVICES`;
- Create CUDA contexts before initializing UCX;
- Ensure merge uses the `"key"` column and asserts expected result size;
- Allow running the benchmark for multiple iterations;
- Allow running multiple warmup iterations;
- Prints results for each iteration;
- Allow enabling garbage collection after each itearation.

### DIFF
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -7,7 +7,6 @@ import cProfile
 import io
 import pickle
 import pstats
-import sys
 from time import perf_counter as clock
 
 import cupy
@@ -20,6 +19,12 @@ import rmm
 
 import ucp
 from ucp.utils import run_on_local_network
+
+
+def sizeof_cudf_dataframe(df):
+    return int(
+        sum(col.memory_usage for col in df._data.columns) + df._index.memory_usage()
+    )
 
 
 async def send_df(ep, df):
@@ -80,7 +85,12 @@ async def exchange_and_concat_bins(rank, eps, bins, timings=None):
     if timings is not None:
         t2 = clock()
         timings.append(
-            (t2 - t1, sum([sys.getsizeof(b) for i, b in enumerate(bins) if i != rank]))
+            (
+                t2 - t1,
+                sum(
+                    [sizeof_cudf_dataframe(b) for i, b in enumerate(bins) if i != rank]
+                ),
+            )
         )
     return cudf.concat(ret)
 
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -5,6 +5,7 @@ import argparse
 import asyncio
 import cProfile
 import io
+import os
 import pickle
 import pstats
 from time import perf_counter as clock
@@ -14,11 +15,15 @@ import numpy as np
 
 from dask.utils import format_bytes, format_time
 
-import cudf
-import rmm
-
 import ucp
 from ucp.utils import run_on_local_network
+
+# Must be set _before_ importing RAPIDS libraries (cuDF, RMM)
+os.environ["RAPIDS_NO_INITIALIZE"] = "True"
+
+
+import cudf  # noqa
+import rmm  # noqa
 
 
 def sizeof_cudf_dataframe(df):
@@ -167,11 +172,7 @@ def generate_chunk(i_chunk, local_size, num_chunks, chunk_type, frac_match):
 
 async def worker(rank, eps, args):
     # Setting current device and make RMM use it
-    dev_id = args.devs[rank % len(args.devs)]
-    cupy.cuda.runtime.setDevice(dev_id)
-    rmm.reinitialize(
-        pool_allocator=True, devices=dev_id, initial_pool_size=args.rmm_init_pool_size
-    )
+    rmm.reinitialize(pool_allocator=True, initial_pool_size=args.rmm_init_pool_size)
 
     # Make cupy use RMM
     cupy.cuda.set_allocator(rmm.rmm_cupy_allocator)
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -366,10 +366,10 @@ def main():
     print_multi(values=["Row matching fraction", f"{args.frac_match}"])
     print_separator(separator="=", length=110)
     print_multi(values=["Wall-clock", f"{format_time(wc)}"])
-    print_multi(values=["Avg Bandwidth / GPU", f"{format_bytes(bw)}/s"])
-    print_multi(values=["Throughput / GPU", f"{format_bytes(tp)}/s"])
+    print_multi(values=["Bandwidth", f"{format_bytes(bw)}/s"])
+    print_multi(values=["Throughput", f"{format_bytes(tp)}/s"])
     print_separator(separator="=", length=110)
-    print_multi(values=["Run", "Wall-clock", "Avg Bandwidth / GPU", "Throughput / GPU"])
+    print_multi(values=["Run", "Wall-clock", "Bandwidth", "Throughput"])
     for i in range(args.iter):
         iter_results = stats[0]["iter_results"]
 
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -361,10 +361,10 @@ def main():
     print_multi(values=["Row matching fraction", f"{args.frac_match}"])
     print_separator(separator="=", length=110)
     print_multi(values=["Wall-clock", f"{format_time(wc)}"])
-    print_multi(values=["Bandwidth", f"{format_bytes(bw)}/s"])
-    print_multi(values=["Throughput", f"{format_bytes(tp)}/s"])
+    print_multi(values=["Avg Bandwidth / GPU", f"{format_bytes(bw)}/s"])
+    print_multi(values=["Throughput / GPU", f"{format_bytes(tp)}/s"])
     print_separator(separator="=", length=110)
-    print_multi(values=["Run", "Wall-clock", "Bandwidth", "Throughput"])
+    print_multi(values=["Run", "Wall-clock", "Avg Bandwidth / GPU", "Throughput / GPU"])
     for i in range(args.iter):
         iter_results = stats[0]["iter_results"]
 
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -111,7 +111,7 @@ async def distributed_join(args, rank, eps, left_table, right_table, timings=Non
 
     left_df = await exchange_and_concat_bins(rank, eps, left_bins, timings)
     right_df = await exchange_and_concat_bins(rank, eps, right_bins, timings)
-    return left_df.merge(right_df)
+    return left_df.merge(right_df, on="key")
 
 
 def generate_chunk(i_chunk, local_size, num_chunks, chunk_type, frac_match):
@@ -206,9 +206,16 @@ async def worker(rank, eps, args):
         iter_timings = []
 
         iter_t = clock()
-        await distributed_join(args, rank, eps, df1, df2, iter_timings)
+        ret = await distributed_join(args, rank, eps, df1, df2, iter_timings)
         await barrier(rank, eps)
         iter_took = clock() - iter_t
+
+        # Ensure the number of matches falls within `args.frac_match` +/- 1%
+        expected_len = args.chunk_size * args.frac_match
+        expected_len_err = expected_len * 0.01
+        assert len(ret) in range(
+            expected_len - expected_len_err, expected_len + expected_len_err
+        )
 
         if args.collect_garbage:
             gc.collect()
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -17,6 +17,7 @@ import numpy as np
 from dask.utils import format_bytes, format_time
 
 import ucp
+from ucp._libs.utils import print_multi, print_separator
 from ucp.utils import run_on_local_network
 
 # Must be set _before_ importing RAPIDS libraries (cuDF, RMM)
@@ -195,13 +196,32 @@ async def worker(rank, eps, args):
         pr = cProfile.Profile()
         pr.enable()
 
+    iter_results = {"bw": [], "wallclock": [], "throughput": [], "data_processed": []}
     timings = []
     t1 = clock()
     for i in range(args.iter):
-        await distributed_join(args, rank, eps, df1, df2, timings)
+        iter_timings = []
+
+        iter_t = clock()
+        await distributed_join(args, rank, eps, df1, df2, iter_timings)
         await barrier(rank, eps)
+        iter_took = clock() - iter_t
+
         if args.collect_garbage:
             gc.collect()
+
+        iter_bw = sum(t[1] for t in iter_timings) / sum(t[0] for t in iter_timings)
+        iter_data_processed = len(df1) * sum([t.itemsize for t in df1.dtypes])
+        iter_data_processed += len(df2) * sum([t.itemsize for t in df2.dtypes])
+        iter_throughput = args.n_chunks * iter_data_processed / iter_took
+
+        iter_results["bw"].append(iter_bw)
+        iter_results["wallclock"].append(iter_took)
+        iter_results["throughput"].append(iter_throughput)
+        iter_results["data_processed"].append(iter_data_processed)
+
+        timings += iter_timings
+
     took = clock() - t1
 
     if args.profile:
@@ -221,6 +241,7 @@ async def worker(rank, eps, args):
         "wallclock": took,
         "throughput": args.n_chunks * data_processed / took,
         "data_processed": data_processed,
+        "iter_results": iter_results,
     }
 
 
@@ -325,18 +346,37 @@ def main():
     bw = sum(s["bw"] for s in stats) / len(stats)
     tp = stats[0]["throughput"]
     dp = sum(s["data_processed"] for s in stats)
+    dp_iter = sum(s["iter_results"]["data_processed"][0] for s in stats)
 
-    print("cudf merge benchmark")
-    print("----------------------------")
-    print(f"device(s)      | {args.devs}")
-    print(f"chunks-per-dev | {args.chunks_per_dev}")
-    print(f"rows-per-chunk | {args.chunk_size}")
-    print(f"data-processed | {format_bytes(dp)}")
-    print(f"frac-match     | {args.frac_match}")
-    print("============================")
-    print(f"Wall-clock     | {format_time(wc)}")
-    print(f"Bandwidth      | {format_bytes(bw)}/s")
-    print(f"Throughput     | {format_bytes(tp)}/s")
+    print("cuDF merge benchmark")
+    print_separator(separator="-", length=110)
+    print_multi(values=["Device(s)", f"{args.devs}"])
+    print_multi(values=["Chunks per device", f"{args.chunks_per_dev}"])
+    print_multi(values=["Rows per chunk", f"{args.chunk_size}"])
+    print_multi(values=["Total data processed", f"{format_bytes(dp)}"])
+    print_multi(values=["Data processed per iter", f"{format_bytes(dp_iter)}"])
+    print_multi(values=["Row matching fraction", f"{args.frac_match}"])
+    print_separator(separator="=", length=110)
+    print_multi(values=["Wall-clock", f"{format_time(wc)}"])
+    print_multi(values=["Bandwidth", f"{format_bytes(bw)}/s"])
+    print_multi(values=["Throughput", f"{format_bytes(tp)}/s"])
+    print_separator(separator="=", length=110)
+    print_multi(values=["Run", "Wall-clock", "Bandwidth", "Throughput"])
+    for i in range(args.iter):
+        iter_results = stats[0]["iter_results"]
+
+        iter_wc = iter_results["wallclock"][i]
+        iter_bw = sum(s["iter_results"]["bw"][i] for s in stats) / len(stats)
+        iter_tp = iter_results["throughput"][i]
+
+        print_multi(
+            values=[
+                i,
+                f"{format_time(iter_wc)}",
+                f"{format_bytes(iter_bw)}/s",
+                f"{format_bytes(iter_tp)}/s",
+            ]
+        )
 
 
 if __name__ == "__main__":
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -4,6 +4,7 @@ Benchmark send receive on one machine
 import argparse
 import asyncio
 import cProfile
+import gc
 import io
 import os
 import pickle
@@ -184,6 +185,8 @@ async def worker(rank, eps, args):
     for i in range(args.warmup_iter):
         await distributed_join(args, rank, eps, df1, df2)
         await barrier(rank, eps)
+        if args.collect_garbage:
+            gc.collect()
 
     if args.cuda_profile:
         cupy.cuda.profiler.start()
@@ -197,6 +200,8 @@ async def worker(rank, eps, args):
     for i in range(args.iter):
         await distributed_join(args, rank, eps, df1, df2, timings)
         await barrier(rank, eps)
+        if args.collect_garbage:
+            gc.collect()
     took = clock() - t1
 
     if args.profile:
@@ -272,6 +277,12 @@ def parse_args():
         default=None,
         type=int,
         help="Initial RMM pool size (default  1/2 total GPU memory)",
+    )
+    parser.add_argument(
+        "--collect-garbage",
+        default=False,
+        action="store_true",
+        help="Trigger Python garbage collection after each iteration.",
     )
     parser.add_argument(
         "--iter",
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -213,9 +213,7 @@ async def worker(rank, eps, args):
         # Ensure the number of matches falls within `args.frac_match` +/- 1%
         expected_len = args.chunk_size * args.frac_match
         expected_len_err = expected_len * 0.01
-        assert len(ret) in range(
-            expected_len - expected_len_err, expected_len + expected_len_err
-        )
+        assert abs(len(ret) - expected_len) <= expected_len_err
 
         if args.collect_garbage:
             gc.collect()
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -307,6 +307,7 @@ def main():
         worker,
         worker_args=args,
         server_address=args.server_address,
+        ensure_cuda_device=True,
     )
 
     wc = stats[0]["wallclock"]
--- a/benchmarks/cudf-merge.py
+++ b/benchmarks/cudf-merge.py
@@ -18,7 +18,7 @@ from dask.utils import format_bytes, format_time
 
 import ucp
 from ucp._libs.utils import print_multi, print_separator
-from ucp.utils import run_on_local_network
+from ucp.utils import hmean, run_on_local_network
 
 # Must be set _before_ importing RAPIDS libraries (cuDF, RMM)
 os.environ["RAPIDS_NO_INITIALIZE"] = "True"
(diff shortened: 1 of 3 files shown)


pentschev:
Since this benchmark is broken due to changes in CUDA context handling, it would be good if this still makes 0.27 (22.08), despite the ongoing burndown. As this is only a benchmark it should not impact release in any way, as this is not executed anywhere as part of that process.

wence-:
> Since this benchmark is broken due to changes in CUDA context handling

Can you explain what you mean by this?

benchmarks/cudf-merge.py
```diff
@@ -4,23 +4,35 @@
...
 import ucp
+from ucp._libs.utils import print_multi, print_separator
 from ucp.utils import run_on_local_network
 
+# Must be set _before_ importing RAPIDS libraries (cuDF, RMM)
+os.environ["RAPIDS_NO_INITIALIZE"] = "True"
+
+
+import cudf  # noqa
+import rmm  # noqa
```
wence-, line 28: Why do we need the noqa? Is it because flake8 complains about imports not at top-level?

benchmarks/cudf-merge.py
```diff
@@ -80,7 +92,12 @@ async def exchange_and_concat_bins(rank, eps, bins, timings=None):
     if timings is not None:
         t2 = clock()
         timings.append(
-            (t2 - t1, sum([sys.getsizeof(b) for i, b in enumerate(bins) if i != rank]))
+            (
+                t2 - t1,
+                sum(
+                    [sizeof_cudf_dataframe(b) for i, b in enumerate(bins) if i != rank]
```
wence-, line 98: So we're measuring the amount of data we've received here?

ucp/utils.py
```diff
@@ -75,10 +75,28 @@ def filter(self, record):
     return logger
 
 
+def _ensure_cuda_device(devs, rank):
+    import numba.cuda
+
+    dev_id = devs[rank % len(devs)]
+    os.environ["CUDA_VISIBLE_DEVICES"] = str(dev_id)
+    numba.cuda.current_context()
```
wence-, line 83: It _should_ be sufficient to use `numba.cuda.select_device(dev_id)` without resorting to `os.environ` munging, but that needs https://github.com/NVIDIA/cuda-python/issues/24 to be fixed first.

wence- (changes requested):
Minor nitpicks, otherwise looks good.

pentschev:
> > Since this benchmark is broken due to changes in CUDA context handling

> Can you explain what you mean by this?

cuDF creates a CUDA context at import time, which wasn't the case in the past, and is the reason we have https://github.com/rapidsai/dask-cuda/pull/379 in Dask-CUDA, if we don't set `CUDA_VISIBLE_DEVICES` then a context will always be created on device 0. Added to that, we now rely on UCX to ensure IB<->GPU mapping automatically, but that requires the CUDA context to be created before `import ucp`.

benchmarks/cudf-merge.py
```diff
@@ -180,10 +196,32 @@ async def worker(rank, eps, args):
...
+        iter_t = clock()
+        await distributed_join(args, rank, eps, df1, df2, iter_timings)
+        await barrier(rank, eps)
+        iter_took = clock() - iter_t
+
+        if args.collect_garbage:
+            gc.collect()
+
+        iter_bw = sum(t[1] for t in iter_timings) / sum(t[0] for t in iter_timings)
+        iter_data_processed = len(df1) * sum([t.itemsize for t in df1.dtypes])
```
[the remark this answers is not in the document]
pentschev: It is part of the `dask_cudf` repo, we don't want an implicit Dask dependency for this benchmark.

benchmarks/cudf-merge.py
```diff
@@ -282,24 +339,44 @@ def main():
...
+    print_multi(values=["Wall-clock", f"{format_time(wc)}"])
+    print_multi(values=["Bandwidth", f"{format_bytes(bw)}/s"])
+    print_multi(values=["Throughput", f"{format_bytes(tp)}/s"])
+    print_separator(separator="=", length=110)
+    print_multi(values=["Run", "Wall-clock", "Bandwidth", "Throughput"])
+    for i in range(args.iter):
+        iter_results = stats[0]["iter_results"]
+
+        iter_wc = iter_results["wallclock"][i]
+        iter_bw = sum(s["iter_results"]["bw"][i] for s in stats) / len(stats)
```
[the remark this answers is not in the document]
pentschev: Done in https://github.com/rapidsai/ucx-py/pull/867/commits/b385060af5dd6495bc2fd3d0bb7f90b79a6babde.

benchmarks/cudf-merge.py
```diff
@@ -282,24 +339,44 @@ def main():
         worker,
         worker_args=args,
         server_address=args.server_address,
+        ensure_cuda_device=True,
     )
 
     wc = stats[0]["wallclock"]
     bw = sum(s["bw"] for s in stats) / len(stats)
```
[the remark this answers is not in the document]
pentschev: Also done in https://github.com/rapidsai/ucx-py/pull/867/commits/b385060af5dd6495bc2fd3d0bb7f90b79a6babde.

benchmarks/cudf-merge.py
```diff
@@ -282,24 +339,44 @@ def main():
...
     tp = stats[0]["throughput"]
```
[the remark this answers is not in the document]
pentschev: The way the code is written today adds a barrier as a synchronization point: https://github.com/rapidsai/ucx-py/pull/867/files#diff-0955423813905f007f223fcbec713336bcee9fdea5c2dc766e93c4006020e266R205-R208 . I guess this is sort of incomplete, but should probably provide a decent overview I guess.

In general, given the load is balanced, I think looking at one worker should probably be accurate enough. The one issue I can immediately think is if we have interconnects that do not provide the same bandwidth in both directions. Any other things that could truly cause results to be completely different when comparing the results of any worker pair?

benchmarks/cudf-merge.py
pentschev: It actually is `isort` rather than `flake8`, but yes.

benchmarks/cudf-merge.py
pentschev: That is correct.

ucp/utils.py
pentschev: Yes, that's right. This isn't critical here as well, so let's wait for that and then we can do it properly once that's completed.

benchmarks/cudf-merge.py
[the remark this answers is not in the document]
wence-: this pr introduces a version in this file, no?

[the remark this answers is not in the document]
pentschev: Ah yes, good catch. Resolved now in https://github.com/rapidsai/ucx-py/pull/867/commits/f8969f2c907e2bc91a9cc762ca8098188688a28a .

benchmarks/cudf-merge.py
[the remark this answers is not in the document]
wence-: Given there's a barrier on every iteration, I think the numbers you get out will be similar. But note that this is then "throughput/GPU" I think (so as you add more workers you expect it to stay constant).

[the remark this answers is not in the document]
wence-: Ah wait, no `args.n_chunks` is multiplied by the number of devices, so this ends up measuring the total throughput.

benchmarks/cudf-merge.py
```diff
@@ -361,10 +361,10 @@ def main():
     print_multi(values=["Row matching fraction", f"{args.frac_match}"])
     print_separator(separator="=", length=110)
     print_multi(values=["Wall-clock", f"{format_time(wc)}"])
-    print_multi(values=["Bandwidth", f"{format_bytes(bw)}/s"])
-    print_multi(values=["Throughput", f"{format_bytes(tp)}/s"])
+    print_multi(values=["Avg Bandwidth / GPU", f"{format_bytes(bw)}/s"])
+    print_multi(values=["Throughput / GPU", f"{format_bytes(tp)}/s"])
     print_separator(separator="=", length=110)
-    print_multi(values=["Run", "Wall-clock", "Bandwidth", "Throughput"])
+    print_multi(values=["Run", "Wall-clock", "Avg Bandwidth / GPU", "Throughput / GPU"])
```
wence-, line 367: Sorry, I screwed up. I hadn't spotted that `n_chunks` is multiplied by the total number of GPUs, so these are aggregate rather than per-GPU numbers.

benchmarks/cudf-merge.py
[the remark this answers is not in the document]
pentschev: Yes, you're right. The same applies for bandwidth, given we compute the average of them all. I made the description clearer in https://github.com/rapidsai/ucx-py/pull/867/commits/2159d1de54eb4376a0274dd137bbd3ee3e027f94, alternatively we could multiple that by the number of ranks and provide the overall value instead.

wence-:
FWIW, the only niggle I have left here is that the `--chunks-per-dev` argument is a bit misnamed. The total number of workers started up is `chunks-per-dev * num_devices` (so when running with all GPUs and more than `chunks-per-dev > 1` we get errors because we try and build 2 (or more) RMM pools per device and things break).

benchmarks/cudf-merge.py
pentschev: Right, but the throughput is a single worker multiplied by `n_chunks`, i.e., only that worker's throughput. So that's not an aggregate of the full cluster?

pentschev:
> FWIW, the only niggle I have left here is that the `--chunks-per-dev` argument is a bit misnamed. The total number of workers started up is `chunks-per-dev * num_devices` (so when running with all GPUs and more than `chunks-per-dev > 1` we get errors because we try and build 2 (or more) RMM pools per device and things break).

The RMM pool being a problem is expected with the default value, that's when the user will have to adjust `--rmm-init-pool-size`. So `--chunks-per-dev` still sounds reasonable, it would be misnamed if it was `--chunks-per-process` or `--chunks-per-worker`. Do you have a proposal on how to make this better?

benchmarks/cudf-merge.py
wence-: `parse_args` does:

```
    args.n_chunks = len(args.devs) * args.chunks_per_dev
```

So I think that this measures the total data (certainly if I run the code, the
data volume is consistent with more GPUs -> higher throughput), no?

wence-:
> > FWIW, the only niggle I have left here is that the `--chunks-per-dev` argument is a bit misnamed. The total number of workers started up is `chunks-per-dev * num_devices` (so when running with all GPUs and more than `chunks-per-dev > 1` we get errors because we try and build 2 (or more) RMM pools per device and things break).
> 
> The RMM pool being a problem is expected with the default value, that's when the user will have to adjust `--rmm-init-pool-size`. So `--chunks-per-dev` still sounds reasonable, it would be misnamed if it was `--chunks-per-process` or `--chunks-per-worker`. Do you have a proposal on how to make this better?

I had naively thought that we would get `num_devices` workers and `chunks-per-dev` would say "how many chunks does each worker have". But actually, each worker only ever has a single chunk, and we always have `num_devices * chunks_per_device` workers (i.e. we oversubscribe the GPUs).

Since this isn't a substantive change from the current behaviour, let's not pollute this PR with too many additional pieces (sorry, this is my fault for noticing!) and leave as-is for now.

benchmarks/cudf-merge.py
```diff
@@ -206,10 +206,17 @@ async def worker(rank, eps, args):
...
+        ret = await distributed_join(args, rank, eps, df1, df2, iter_timings)
         await barrier(rank, eps)
         iter_took = clock() - iter_t
 
+        # Ensure the number of matches falls within `args.frac_match` +/- 1%
+        expected_len = args.chunk_size * args.frac_match
+        expected_len_err = expected_len * 0.01
+        assert len(ret) in range(
+            expected_len - expected_len_err, expected_len + expected_len_err
+        )
```
wence-, line 218: I think that the dataframe generation is such that `allreduce(len(ret), op=SUM)` is exactly equal to `expected_len`, so perhaps do that? A bad hash could result in this assertion failing even if the collective result is correct.

pentschev: In this particular benchmark we don't do `allreduce`, maybe we could introduce this in the future but I'll try not to tackle everything in this one PR. IMO, if we have a bad hash and the assertion fails, this is probably worth causing an outright failure to be investigated and the benchmarks made more robust, what do you think?

wence-: Yes, I think that's reasonable.

benchmarks/cudf-merge.py
```diff
@@ -180,10 +199,39 @@ async def worker(rank, eps, args):
...
+        ret = await distributed_join(args, rank, eps, df1, df2, iter_timings)
+        await barrier(rank, eps)
+        iter_took = clock() - iter_t
+
+        # Ensure the number of matches falls within `args.frac_match` +/- 1%
+        expected_len = args.chunk_size * args.frac_match
+        expected_len_err = expected_len * 0.01
+        assert len(ret) in range(
+            expected_len - expected_len_err, expected_len + expected_len_err
+        )
```
wence-, line 218:
```suggestion
        assert abs(len(ret) - expected_len) <= expected_len_err
```
Since `expected_len_err` is a float

benchmarks/cudf-merge.py
wence-: I think this is the final issue with this change for me. Due to the way `args.n_chunks` is updated in `parse_args`, the printouts here are really the aggregate bandwidth (not per GPU):

```
$ python cudf-merge.py --warmup-iter 2 --frac-match 0.3 -c 10_000_000 -d 0,1 --chunks-per-dev 1 --iter 10 
cuDF merge benchmark
--------------------------------------------------------------------------------------------------------------
Device(s)                 | [0, 1]
Chunks per device         | 1
Rows per chunk            | 10000000
Total data processed      | 5.96 GiB
Data processed per iter   | 610.35 MiB
Row matching fraction     | 0.3
==============================================================================================================
Wall-clock                | 391.33 ms
Avg Bandwidth / GPU       | 13.71 GiB/s
Throughput / GPU          | 15.23 GiB/s
==============================================================================================================
Run                       | Wall-clock                | Avg Bandwidth / GPU       | Throughput / GPU
0                         | 38.73 ms                  | 14.04 GiB/s               | 15.39 GiB/s
1                         | 40.32 ms                  | 12.46 GiB/s               | 14.78 GiB/s
2                         | 38.30 ms                  | 13.90 GiB/s               | 15.56 GiB/s
3                         | 39.93 ms                  | 14.14 GiB/s               | 14.93 GiB/s
4                         | 37.84 ms                  | 13.49 GiB/s               | 15.75 GiB/s
5                         | 38.26 ms                  | 13.92 GiB/s               | 15.58 GiB/s
6                         | 38.08 ms                  | 13.83 GiB/s               | 15.65 GiB/s
7                         | 38.00 ms                  | 13.85 GiB/s               | 15.69 GiB/s
8                         | 37.89 ms                  | 13.93 GiB/s               | 15.73 GiB/s
9                         | 38.15 ms                  | 13.74 GiB/s               | 15.63 GiB/s
```
vs.
```
$ python cudf-merge.py --warmup-iter 2 --frac-match 0.3 -c 10_000_000 -d 0,1,2,3 --chunks-per-dev 1 --iter 10 
cuDF merge benchmark
--------------------------------------------------------------------------------------------------------------
Device(s)                 | [0, 1, 2, 3]
Chunks per device         | 1
Rows per chunk            | 10000000
Total data processed      | 11.92 GiB
Data processed per iter   | 1.19 GiB
Row matching fraction     | 0.3
==============================================================================================================
Wall-clock                | 391.67 ms
Avg Bandwidth / GPU       | 24.57 GiB/s
Throughput / GPU          | 30.44 GiB/s
==============================================================================================================
Run                       | Wall-clock                | Avg Bandwidth / GPU       | Throughput / GPU
0                         | 36.86 ms                  | 27.46 GiB/s               | 32.34 GiB/s
1                         | 39.28 ms                  | 25.39 GiB/s               | 30.35 GiB/s
2                         | 37.59 ms                  | 23.88 GiB/s               | 31.71 GiB/s
3                         | 39.75 ms                  | 22.00 GiB/s               | 29.99 GiB/s
4                         | 37.14 ms                  | 25.90 GiB/s               | 32.10 GiB/s
5                         | 37.70 ms                  | 25.11 GiB/s               | 31.62 GiB/s
6                         | 36.74 ms                  | 26.41 GiB/s               | 32.44 GiB/s
7                         | 37.22 ms                  | 25.15 GiB/s               | 32.02 GiB/s
8                         | 39.57 ms                  | 22.13 GiB/s               | 30.13 GiB/s
9                         | 43.67 ms                  | 23.50 GiB/s               | 27.30 GiB/s
```

So I think just removing the ` / GPU` is correct.

pentschev: I'm normally skeptical of comparing different GPUs in a DGX-1 because of the different interconnects they have available. Part of the interconnects are NV1 (25GB/s) and the remaining are NV2 (50GB/s) as seen with `nvidia-smi topo -m`, and given GPUs `0,1` are NV1, that is at least partially responsible for the bandwidth increase.

A DGX-2 is more reliable in that matter, as bandwidths should be almost equal due to NVSwitch. In this case we still see `0,1,2,3` having higher average bandwidth than `0,1`, but not as significant as in DGX-1 which showed pretty much linear scaling. See below:

```
$ python cudf-merge.py --warmup-iter 2 --frac-match 0.3 -c 10_000_000 -d 0,1 --chunks-per-dev 1 --iter 10
cuDF merge benchmark
--------------------------------------------------------------------------------------------------------------
Device(s)                 | [0, 1]
Chunks per device         | 1
Rows per chunk            | 10000000
Total data processed      | 5.96 GiB
Data processed per iter   | 610.35 MiB
Row matching fraction     | 0.3
==============================================================================================================
Wall-clock                | 280.93 ms
Avg Bandwidth / GPU       | 31.95 GiB/s
Throughput / GPU          | 21.22 GiB/s
==============================================================================================================
Run                       | Wall-clock                | Avg Bandwidth / GPU       | Throughput / GPU
0                         | 31.78 ms                  | 20.59 GiB/s               | 18.75 GiB/s
1                         | 27.74 ms                  | 33.78 GiB/s               | 21.48 GiB/s
2                         | 27.58 ms                  | 33.25 GiB/s               | 21.61 GiB/s
3                         | 26.92 ms                  | 35.33 GiB/s               | 22.14 GiB/s
4                         | 27.00 ms                  | 35.76 GiB/s               | 22.07 GiB/s
5                         | 27.22 ms                  | 32.73 GiB/s               | 21.89 GiB/s
6                         | 26.70 ms                  | 35.41 GiB/s               | 22.33 GiB/s
7                         | 27.10 ms                  | 34.20 GiB/s               | 22.00 GiB/s
8                         | 27.41 ms                  | 32.37 GiB/s               | 21.74 GiB/s
9                         | 26.92 ms                  | 33.85 GiB/s               | 22.14 GiB/s

$ python cudf-merge.py --warmup-iter 2 --frac-match 0.3 -c 10_000_000 -d 0,1,2,3 --chunks-per-dev 1 --iter 10
cuDF merge benchmark
--------------------------------------------------------------------------------------------------------------
Device(s)                 | [0, 1, 2, 3]
Chunks per device         | 1
Rows per chunk            | 10000000
Total data processed      | 11.92 GiB
Data processed per iter   | 1.19 GiB
Row matching fraction     | 0.3
==============================================================================================================
Wall-clock                | 300.60 ms
Avg Bandwidth / GPU       | 42.57 GiB/s
Throughput / GPU          | 39.66 GiB/s
==============================================================================================================
Run                       | Wall-clock                | Avg Bandwidth / GPU       | Throughput / GPU
0                         | 29.62 ms                  | 45.72 GiB/s               | 40.24 GiB/s
1                         | 29.15 ms                  | 47.19 GiB/s               | 40.90 GiB/s
2                         | 29.37 ms                  | 44.94 GiB/s               | 40.60 GiB/s
3                         | 29.07 ms                  | 46.73 GiB/s               | 41.00 GiB/s
4                         | 28.88 ms                  | 48.47 GiB/s               | 41.28 GiB/s
5                         | 31.43 ms                  | 34.46 GiB/s               | 37.93 GiB/s
6                         | 31.60 ms                  | 32.29 GiB/s               | 37.72 GiB/s
7                         | 28.41 ms                  | 48.06 GiB/s               | 41.96 GiB/s
8                         | 28.27 ms                  | 49.47 GiB/s               | 42.16 GiB/s
9                         | 30.27 ms                  | 37.62 GiB/s               | 39.38 GiB/s
```

For a DGX-1, I think a more fair comparison would be to set `UCX_TLS=^cuda_ipc` to only use InfiniBand, providing equal bandwidth between all endpoints, in which case we get:

```
$ UCX_TLS=^cuda_ipc python cudf-merge.py --warmup-iter 2 --frac-match 0.3 -c 10_000_000 -d 0,1 --chunks-per-dev 1 --iter 10
cuDF merge benchmark
--------------------------------------------------------------------------------------------------------------
Device(s)                 | [0, 1]
Chunks per device         | 1
Rows per chunk            | 10000000
Total data processed      | 5.96 GiB
Data processed per iter   | 610.35 MiB
Row matching fraction     | 0.3
==============================================================================================================
Wall-clock                | 696.12 ms
Avg Bandwidth / GPU       | 4.88 GiB/s
Throughput / GPU          | 8.56 GiB/s
==============================================================================================================
Run                       | Wall-clock                | Avg Bandwidth / GPU       | Throughput / GPU
0                         | 68.34 ms                  | 4.93 GiB/s                | 8.72 GiB/s
1                         | 69.05 ms                  | 4.89 GiB/s                | 8.63 GiB/s
2                         | 68.63 ms                  | 4.89 GiB/s                | 8.69 GiB/s
3                         | 68.08 ms                  | 4.95 GiB/s                | 8.76 GiB/s
4                         | 71.25 ms                  | 4.87 GiB/s                | 8.37 GiB/s
5                         | 68.88 ms                  | 4.87 GiB/s                | 8.65 GiB/s
6                         | 68.80 ms                  | 4.88 GiB/s                | 8.66 GiB/s
7                         | 69.08 ms                  | 4.84 GiB/s                | 8.63 GiB/s
8                         | 69.02 ms                  | 4.85 GiB/s                | 8.64 GiB/s
9                         | 69.26 ms                  | 4.84 GiB/s                | 8.61 GiB/s

$ UCX_TLS=^cuda_ipc python cudf-merge.py --warmup-iter 2 --frac-match 0.3 -c 10_000_000 -d 0,1,2,3 --chunks-per-dev 1 --iter 10
cuDF merge benchmark
--------------------------------------------------------------------------------------------------------------
Device(s)                 | [0, 1, 2, 3]
Chunks per device         | 1
Rows per chunk            | 10000000
Total data processed      | 11.92 GiB
Data processed per iter   | 1.19 GiB
Row matching fraction     | 0.3
==============================================================================================================
Wall-clock                | 919.79 ms
Avg Bandwidth / GPU       | 5.03 GiB/s
Throughput / GPU          | 12.96 GiB/s
==============================================================================================================
Run                       | Wall-clock                | Avg Bandwidth / GPU       | Throughput / GPU
0                         | 91.15 ms                  | 5.00 GiB/s                | 13.08 GiB/s
1                         | 91.24 ms                  | 5.04 GiB/s                | 13.07 GiB/s
2                         | 91.64 ms                  | 5.02 GiB/s                | 13.01 GiB/s
3                         | 91.26 ms                  | 5.05 GiB/s                | 13.06 GiB/s
4                         | 91.15 ms                  | 5.05 GiB/s                | 13.08 GiB/s
5                         | 91.94 ms                  | 5.02 GiB/s                | 12.97 GiB/s
6                         | 91.45 ms                  | 5.02 GiB/s                | 13.04 GiB/s
7                         | 91.11 ms                  | 5.06 GiB/s                | 13.08 GiB/s
8                         | 91.00 ms                  | 5.06 GiB/s                | 13.10 GiB/s
9                         | 91.53 ms                  | 5.00 GiB/s                | 13.02 GiB/s
```

In the above case results are not as significant as well, just like the DGX-2. In other words, I'm not sure if this alone proves/disproves if we should write `/ GPU` or not. What do you think @wence- ?

wence-: The throughput/GPU should be at best constant (since the partition by hash and local merge are perfectly data parallel, whereas the the communication only costs something that increases with the number of GPUs). So I don't know how the throughput/GPU  can almost double going from 2 to 4 GPUs.

pentschev: On the DGX-2 case for example, the wall-clock is just slightly lower (~10%) for 2 GPUs when compared to the 4 GPUs sample, even though we increased the problem size with more GPUs. That would cause throughput/GPU to go up, given the problem size is now larger, no?

wence-: > On the DGX-2 case for example, the wall-clock is just slightly lower (~10%) for 2 GPUs when compared to the 4 GPUs sample, even though we increased the problem size with more GPUs. That would cause throughput/GPU to go up, given the problem size is now larger, no?

I don't think so. Here's my explanation:

On 2 GPUs, the benchmark processes a total of 615MB data in 70 ms (or about 8.8GB/s). So the processing rate per GPU is 4.4 GB/s;

On 4 GPUs, the benchmark processes a total of 1.2GB data in 92 ms (or about 13GB/s). So the processing rate per GPU is 3.25 GB/s.

The _overall_ throughput is higher (as we would hope). But for throughput/GPU to go up, the wallclock would have to decrease as we add more GPUs: we're doing a weak scaling experiment here where we are increasing the amount of data to be processed linearly and linearly adding more compute to deal with it.

pentschev: Yeah, ok, based on the total processed data I agree that your explanation looks correct. I reverted that change in https://github.com/rapidsai/ucx-py/pull/867/commits/ae884a6dbfaa282ba27d2a6d3a419d5025b3dd5d .

However, in the future we may want to refactor a bit pieces of this code to make sure that it's more immediately evident. I think the fact that we are discussing whether this is right or not feels like something could be improved. For now, let's leave it as is.

pentschev:
I think all changes are in now, could you check one last time/approve the PR @wence- ?

wence- (approved):

wence-:
@gpucibot merge

wence-:
I am happy to merge this, though does it need further approval to go to 0.27 (rather than 0.28) ?

pentschev:
> I am happy to merge this, though does it need further approval to go to 0.27 (rather than 0.28) ?

UCX-Py doesn't follow exactly the same conditions as the remaining of RAPIDS, and given there are no other required reviewers, we should be good to merge. Also gpucibot has no power here. :)

pentschev:
For the very small benchmark in CI, the assertion failed:

```
$ python cudf-merge.py --chunks-per-dev 4 --chunk-size 10000 --rm
m-init-pool-size 2097152
Process SpawnProcess-1:
  File "/datasets/pentschev/miniconda3/envs/rn-220803/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/datasets/pentschev/miniconda3/envs/rn-220803/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/datasets/pentschev/miniconda3/envs/rn-220803/lib/python3.8/site-packages/ucp/utils.py", line 131, in _worker_process
    ret = loop.run_until_complete(run())
  File "/datasets/pentschev/miniconda3/envs/rn-220803/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/datasets/pentschev/miniconda3/envs/rn-220803/lib/python3.8/site-packages/ucp/utils.py", line 126, in run
    return await func(rank, eps, args)
  File "/datasets/pentschev/src/ucx-py/benchmarks/cudf-merge.py", line 216, in worker
    assert abs(len(ret) - expected_len) <= expected_len_err
```

Maximum error was 30, and the actual value was 31. Since this is not critical, I increased the tolerance to 2% now.

pentschev:
And it seems I was wrong. We have to ask for approval.

msadang (approved):
LGTM

quasiben (approved):

PR #872, opened 2022-08-25
Set `RAPIDS_NO_INITIALIZE` in `conftest.py`

pentschev:
Without this variable set, some tests that need to initialize the CUDA context on the appropriate device will fail, as importing a library such as `cudf` in another test will cause CUDA context to be created.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,16 @@
 import asyncio
+import os
 
 import pytest
 
 import ucp
+
+# Prevent calls such as `cudf = pytest.importorskip("cudf")` from initializing
+# a CUDA context. Such calls may cause tests that must initialize the CUDA
+# context on the appropriate device to fail.
+# For example, without `RAPIDS_NO_INITIALIZE=True`, `test_benchmark_cluster`
+# will succeed if running alone, but fails when all tests are run in batch.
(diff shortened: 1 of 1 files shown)


wence- (approved):
I hope that with cuda-python 11.8 this problem will be gone (I am confirming that the device attribute querying that is done in cudf no longer initializes a device context...), but this looks fine, please go ahead and merge.

pentschev:
> I hope that with cuda-python 11.8 this problem will be gone (I am confirming that the device attribute querying that is done in cudf no longer initializes a device context...), but this looks fine, please go ahead and merge.

That would be great indeed. Thanks @wence- !

wence-:
> I hope that with cuda-python 11.8 this problem will be gone (I am confirming that the device attribute querying that is done in cudf no longer initializes a device context...)

This does appear to be the case, so modulo all the other palaver around cuda-toolkit conda packages, we should be able to remove `RAPIDS_NO_INITIALIZE` when we depend on cuda-python 11.8

pentschev:
> This does appear to be the case, so modulo all the other palaver around cuda-toolkit conda packages, we should be able to remove `RAPIDS_NO_INITIALIZE` when we depend on cuda-python 11.8

That will probably take a while until we drop all pre-11.8 support though, but it's nice anyway to know this will be resolved eventually!

PR #873, opened 2022-09-01
Reorganize `send_recv` benchmarks backends and unify them

pentschev:
A substantial part of the `send_recv` and `send_recv_core` benchmarks consist of common code. By reorganizing the implementation into separate backends it's possible to unify them, thus reducing code duplication and opening up the possibility to introduce additional backends in more structured way.


ucp/benchmarks/backends/ucp_core.py
```diff
@@ -0,0 +1,287 @@
...
+                blocking_am_recv(worker, ep)
+            else:
+                if not self.args.reuse_alloc:
+                    recv_msg = self.xp.zeros(self.args.n_bytes, dtype="u1")
+
+                if self.args.delay_progress:
+                    maybe_progress()
+                    non_blocking_send(worker, ep, send_msg, op_started, op_completed)
+                    maybe_progress()
+                    non_blocking_recv(worker, ep, recv_msg, op_started, op_completed)
```
wence-, line 255: Usual convention for non-blocking point to point is to post receives before sends.

wence-: So:
```
non_blocking_recv(...)
non_blocking_send(...)
maybe_progress()
```
?

ucp/benchmarks/backends/ucp_core.py
```diff
@@ -0,0 +1,287 @@
...
+                else:
+                    blocking_send(worker, ep, send_msg)
+                    blocking_recv(worker, ep, recv_msg)
+
+            stop = monotonic()
+            if i >= self.args.n_warmup_iter:
+                times.append(stop - start)
```
wence-, line 262: In the delay progress case, pretty sure this doesn't actually measure the time for messages to round trip

ucp/benchmarks/backends/ucp_core.py
```diff
@@ -0,0 +1,287 @@
...
+
+        if self.args.delay_progress:
+            while finished[0] != 2 * (self.args.n_iter + self.args.n_warmup_iter):
+                worker.progress()
```
wence-, line 266: Because your messages have only completed after this loop is done...

ucp/benchmarks/send_recv.py
```diff
@@ -77,43 +63,16 @@ def server(queue, args):
...
-        while not lf.closed():
-            await asyncio.sleep(0.5)
-
-    loop = get_event_loop()
-    loop.run_until_complete(run())
+    if asyncio.iscoroutinefunction(server.run):
+        loop = get_event_loop()
+        loop.run_until_complete(server.run())
+    else:
+        server.run()
```
wence-, line 75: Maybe merge the conditions (since `server.run` is a coroutine iff `args.backend == "ucxpy-async"`) ?

```
if args.backend == "ucxpy-async":
    server = UCXPyAsyncServer(...)
    get_event_loop().run_until_complete(server.run())
elif ...:
    server = UCXPyCoreServer(...)
    server.run()
```
?

ucp/benchmarks/send_recv.py
```diff
@@ -142,53 +101,19 @@ def client(queue, port, server_address, args):
...
+    if args.backend == "ucp-async":
+        client = UCXPyAsyncClient(args, xp, queue, server_address, port)
+    elif args.backend == "ucp-core":
+        client = UCXPyCoreClient(args, xp, queue, server_address, port)
+
+    if asyncio.iscoroutinefunction(client.run):
+        loop = get_event_loop()
+        loop.run_until_complete(client.run())
+    else:
+        client.run()
```
wence-, line 113: Again merge conditions?

wence- (changes requested):
Overall looks great, I had a bunch of comments/queries but they are all very minor.

ucp/benchmarks/send_recv.py
pentschev: This iff is only temporary, in https://github.com/rapidsai/ucx-py/pull/874 we introduce another backend that makes this a more logical condition: https://github.com/rapidsai/ucx-py/pull/874/files#diff-e2b415b95132fb557a1c441b2959d7da1b114a8164eac1156161d3101eb376a3R85-R91 . So if you don't mind, I'll disregard this suggestion and we can make any changes in that follow-up PR.

ucp/benchmarks/backends/ucp_core.py
```diff
@@ -0,0 +1,287 @@
...
+        listener = ucx_api.UCXListener(
+            worker=worker,
+            port=self.args.port or 0,
+            cb_func=_listener_handler,
+            cb_args=(msg,),
+        )
+        self.queue.put(listener.port)
+
+        while outstanding[0] == 0:
+            worker.progress()
```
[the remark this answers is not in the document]
pentschev: Not really, progressing the worker ensures they eventually complete, requests are only cleaned up when they go out-of-scope, but perhaps your comment is not exactly in the context of what I'm thinking of in my head. Please let me know if any further clarification is required here.

ucp/benchmarks/backends/ucp_core.py
pentschev: For the Tag API, that's ok. I'm generally hesitant to make this kind of change because the stream API doesn't guarantee the ordering: https://github.com/rapidsai/ucx-py/blob/branch-0.28/ucp/core.py#L48-L55 .

ucp/benchmarks/backends/base.py
```diff
@@ -0,0 +1,69 @@
...
+
+        Parameters
+        ----------
+        args: argparse.Namespace
+            Parsed command-line arguments that will be used as parameters during
+            the `run` method.
+        xp: module
+            Module implementing the NumPy API to use for data generation.
+        queue: Queue
+            Queue object where server will put the port it is listening at.
```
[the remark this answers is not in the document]
pentschev: Done in https://github.com/rapidsai/ucx-py/pull/873/commits/0dfe25305e893086472e1438cdd19ad10cc4531d .

ucp/benchmarks/backends/ucp_core.py
pentschev: In any case applied that change in https://github.com/rapidsai/ucx-py/pull/873/commits/9a88d1cf5324a9ecce86270eb1c0a6ad384fc418 .

pentschev (commented):
Thanks @wence- . I believe I've addressed all your comments/suggestions, but please let me know if I missed anything. Could you also take another look when you have a chance?

ucp/benchmarks/backends/ucp_core.py
```diff
@@ -0,0 +1,287 @@
...
+        outstanding = [0]
+
+        def op_started():
+            with op_lock:
+                outstanding[0] += 1
+
+        def op_completed():
+            with op_lock:
+                outstanding[0] -= 1
+                finished[0] += 1
```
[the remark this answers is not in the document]
wence-: Ah yeah, this doesn't work, but `nonlocal` does. I think it's ok to have it in a list, I'm just recalling with horror the last time I wrote java at which time you could only close over `final` (read-only) variables, so had to do this same trick.

I think it's fine to leave as is.

ucp/benchmarks/backends/ucp_core.py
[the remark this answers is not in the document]
wence-: So `tag_send_nb :: stuff -> Optional[UCXRequest]`. So I thought that if you don't grab the request handles from the return value then you can never wait on them:

```
req = tag_send_nb(...)

...
while not req.closed():
    worker.progress()
```

But IIUC, the way you handle this is that the `_tag_recv_handle` callback is called every time a request completes and that decrements the outstanding message count. Internally somehow you're holding on to these requests until they complete (ah yes, the endpoint tracks all the in flight messages in a `set`).

ucp/benchmarks/backends/ucp_core.py
```diff
@@ -0,0 +1,287 @@
...
+                endpoint_error_handling=True,
+            )
+
+            # Wireup before starting to transfer data
+            if self.args.enable_am is True:
+                ucx_api.am_recv_nb(ep, cb_func=_am_recv_handle, cb_args=(ep,))
+            else:
+                wireup = Array(bytearray(len(WireupMessage)))
+                op_started()
+                ucx_api.tag_recv_nb(
```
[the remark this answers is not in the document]
wence-: I meant that should we not wait on the request to complete (possibly before starting the rest of the ping-pong messages).

The client does a blocking send/recv to exchange the wireup information, but on the server we do a nonblocking wireup.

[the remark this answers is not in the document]
pentschev: I see what you're saying now. I think this is not a problem to have the synchronization only on the client, given this is also a Tag message as we have a guarantee of message ordering. The case where this may be problematic is if the message fails, but then it would anyway eventually be reported in order as well.

Do you have any other concerns that I'm perhaps not thinking about?

ucp/benchmarks/backends/ucp_core.py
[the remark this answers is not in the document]
pentschev: That is all absolutely correct.

ucp/benchmarks/backends/ucp_core.py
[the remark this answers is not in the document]
wence-: Nope.

wence- (approved):
Looks great, thanks!

pentschev:
Thanks @wence- for the review!

ajschmidt8 (approved):
Approving `ops-codeowner` file changes

pentschev:
Thanks @ajschmidt8 for the approval!

PR #874, opened 2022-09-01
Add Tornado `send_recv` benchmark

pentschev:
Extends `send_recv` benchmark adding Tornado backend.


ucp/benchmarks/send_recv.py
```diff
@@ -101,10 +117,9 @@ def client(queue, port, server_address, args):
         xp.cuda.runtime.setDevice(args.client_dev)
         xp.cuda.set_allocator(rmm.rmm_cupy_allocator)
 
-    if args.backend == "ucp-async":
-        client = UCXPyAsyncClient(args, xp, queue, server_address, port)
-    elif args.backend == "ucp-core":
-        client = UCXPyCoreClient(args, xp, queue, server_address, port)
+    client = _get_backend_implementation(args.backend)["client"](
+        args, xp, queue, server_address, port
+    )
```
pentschev, line 122: @wence- regarding https://github.com/rapidsai/ucx-py/pull/873/files#r961566376 and https://github.com/rapidsai/ucx-py/pull/873/files#r961566555 , I hope this block makes it clear why we didn't merge the conditions there, we have multiple backends now that have different requirements.

ucp/benchmarks/backends/tornado.py
```diff
@@ -0,0 +1,121 @@
...
+
+                for i in range(args.n_iter + args.n_warmup_iter):
+                    try:
+                        await stream.read_into(msg_recv_list[i].data)
+                        await stream.write(msg_recv_list[i].data)
+                    except StreamClosedError as e:
+                        print(e)
+                        break
+
+                event.set()
```
wence-, line 64: I think you only need `nonlocal event`, but perhaps instead:
```suggestion
        class TransferServer(TCPServer):
            def __init__(self, event, *args, **kwargs):
                super().__init__(*args, **kwargs)
                self.event = event

            async def handle_stream(self, stream, address):
                global event
                msg_recv_list = []
                if not args.reuse_alloc:
                    for _ in range(args.n_iter + args.n_warmup_iter):
                        msg_recv_list.append(xp.zeros(args.n_bytes, dtype="u1"))
                else:
                    t = xp.zeros(args.n_bytes, dtype="u1")
                    for _ in range(args.n_iter + args.n_warmup_iter):
                        msg_recv_list.append(t)

                assert msg_recv_list[0].nbytes == args.n_bytes

                for i in range(args.n_iter + args.n_warmup_iter):
                    try:
                        await stream.read_into(msg_recv_list[i].data)
                        await stream.write(msg_recv_list[i].data)
                    except StreamClosedError as e:
                        print(e)
                        break

                self.event.set()
```

ucp/benchmarks/backends/tornado.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Set max_buffer_size to 1 GiB for now
+        server = TransferServer(max_buffer_size=1024**3)
```
wence-, line 67: And then here
```suggestion
        server = TransferServer(event, max_buffer_size=1024**3)
```

ucp/benchmarks/backends/tornado.py
```diff
@@ -0,0 +1,121 @@
...
+            server.listen(port)
+        else:
+            for i in range(10000, 60000):
+                try:
+                    server.listen(i)
+                except OSError:
+                    continue
+                else:
+                    port = i
+                    break
```
wence-, line 31: You can `server.listen(0)` to get a random port, but then it appears very difficult to actually determine what the port you got was, so meh I guess.

ucp/benchmarks/backends/tornado.py
```diff
@@ -0,0 +1,121 @@
...
+                    t = xp.zeros(args.n_bytes, dtype="u1")
+                    for _ in range(args.n_iter + args.n_warmup_iter):
+                        msg_recv_list.append(t)
+
+                assert msg_recv_list[0].nbytes == args.n_bytes
+
+                for i in range(args.n_iter + args.n_warmup_iter):
+                    try:
+                        await stream.read_into(msg_recv_list[i].data)
+                        await stream.write(msg_recv_list[i].data)
```
wence-, line 59: I know this mimics the other benchmarks but in the "alloc each time case" I would have though a more realistic setup would be to allocate as we go (rather than allocate niter buffers up front). WDYT?

wence- (changes requested):
Looks good overall, some mostly minor nitpicks and a question about the "allocate every time" case

ucp/benchmarks/backends/tornado.py
pentschev: Hmm, I wouldn't mind, but it seems using `read_bytes` instead of `read_into` makes it much slower, see below.

<details><summary>`read_into` (current)</summary>

```
$ python -m ucp.benchmarks.send_recv -l tornado
Server Running at 10.33.225.163:10000
Client connecting to server at 10.33.225.163:10000
Roundtrip benchmark
================================================================================
Iterations                | 10
Bytes                     | 9.54 MiB
Object type               | numpy
Reuse allocation          | False
================================================================================
Device(s)                 | CPU-only
Server CPU                | affinity not set
Client CPU                | affinity not set
================================================================================
Bandwidth (average)       | 2.10 GiB/s
Bandwidth (median)        | 2.10 GiB/s
Latency (average)         | 4427092 ns
Latency (median)          | 4433597 ns
================================================================================
Iterations                | Bandwidth, Latency
--------------------------------------------------------------------------------
0                         | 2.10 GiB/s, 4432572ns
1                         | 2.09 GiB/s, 4446343ns
2                         | 2.09 GiB/s, 4447852ns
3                         | 2.10 GiB/s, 4434623ns
4                         | 2.11 GiB/s, 4414575ns
5                         | 2.11 GiB/s, 4419028ns
6                         | 2.11 GiB/s, 4418157ns
7                         | 2.08 GiB/s, 4472294ns
8                         | 2.10 GiB/s, 4435407ns
9                         | 2.14 GiB/s, 4350074ns
```

</details>

<details><summary>`read_bytes` (proposed)</summary>

```
$ python -m ucp.benchmarks.send_recv -l tornado
Server Running at 10.33.225.163:10000
Client connecting to server at 10.33.225.163:10000
Roundtrip benchmark
================================================================================
Iterations                | 10
Bytes                     | 9.54 MiB
Object type               | numpy
Reuse allocation          | False
================================================================================
Device(s)                 | CPU-only
Server CPU                | affinity not set
Client CPU                | affinity not set
================================================================================
Bandwidth (average)       | 1.27 GiB/s
Bandwidth (median)        | 1.27 GiB/s
Latency (average)         | 7312802 ns
Latency (median)          | 7327963 ns
================================================================================
Iterations                | Bandwidth, Latency
--------------------------------------------------------------------------------
0                         | 1.27 GiB/s, 7342915ns
1                         | 1.27 GiB/s, 7331363ns
2                         | 1.27 GiB/s, 7361351ns
3                         | 1.27 GiB/s, 7324030ns
4                         | 1.27 GiB/s, 7324563ns
5                         | 1.26 GiB/s, 7362499ns
6                         | 1.27 GiB/s, 7321720ns
7                         | 1.27 GiB/s, 7348971ns
8                         | 1.27 GiB/s, 7314609ns
9                         | 1.31 GiB/s, 7096002ns
```

</details>

To avoid giving UCX-Py a potentially unfair advantage, it seems better to keep it as is. Also, not committing it yet, but below is diff for the changes I've made, in case you can spot something I did wrong and created an unfair advantage for us:

<details><summary>`read_bytes` patch</summary>

```diff
diff --git a/ucp/benchmarks/backends/tornado.py b/ucp/benchmarks/backends/tornado.py
index 3076d99..8bd90f1 100644
--- a/ucp/benchmarks/backends/tornado.py
+++ b/ucp/benchmarks/backends/tornado.py
@@ -42,20 +42,21 @@ class TornadoServer(BaseServer):
             async def handle_stream(self, stream, address):
                 global event
                 msg_recv_list = []
-                if not args.reuse_alloc:
-                    for _ in range(args.n_iter + args.n_warmup_iter):
-                        msg_recv_list.append(xp.zeros(args.n_bytes, dtype="u1"))
-                else:
+                if args.reuse_alloc:
                     t = xp.zeros(args.n_bytes, dtype="u1")
                     for _ in range(args.n_iter + args.n_warmup_iter):
                         msg_recv_list.append(t)

-                assert msg_recv_list[0].nbytes == args.n_bytes
+                    assert msg_recv_list[0].nbytes == args.n_bytes

                 for i in range(args.n_iter + args.n_warmup_iter):
                     try:
-                        await stream.read_into(msg_recv_list[i].data)
-                        await stream.write(msg_recv_list[i].data)
+                        if args.reuse_alloc:
+                            await stream.read_into(msg_recv_list[i].data)
+                            await stream.write(msg_recv_list[i].data)
+                        else:
+                            msg_recv_list.append(await stream.read_bytes(args.n_bytes))
+                            await stream.write(msg_recv_list[i])
                     except StreamClosedError as e:
                         print(e)
                         break
@@ -89,18 +90,18 @@ class TornadoClient(BaseClient):

         msg_send_list = []
         msg_recv_list = []
-        if not self.args.reuse_alloc:
-            for i in range(self.args.n_iter + self.args.n_warmup_iter):
-                msg_send_list.append(self.xp.arange(self.args.n_bytes, dtype="u1"))
-                msg_recv_list.append(self.xp.zeros(self.args.n_bytes, dtype="u1"))
-        else:
+        if self.args.reuse_alloc:
             t1 = self.xp.arange(self.args.n_bytes, dtype="u1")
             t2 = self.xp.zeros(self.args.n_bytes, dtype="u1")
             for i in range(self.args.n_iter + self.args.n_warmup_iter):
                 msg_send_list.append(t1)
                 msg_recv_list.append(t2)
+            assert msg_recv_list[0].nbytes == self.args.n_bytes
+        else:
+            for i in range(self.args.n_iter + self.args.n_warmup_iter):
+                msg_send_list.append(self.xp.arange(self.args.n_bytes, dtype="u1"))
+
         assert msg_send_list[0].nbytes == self.args.n_bytes
-        assert msg_recv_list[0].nbytes == self.args.n_bytes

         if self.args.cuda_profile:
             self.xp.cuda.profiler.start()
@@ -109,7 +110,10 @@ class TornadoClient(BaseClient):
             start = monotonic()

             await stream.write(msg_send_list[i].data)
-            await stream.read_into(msg_recv_list[i].data)
+            if self.args.reuse_alloc:
+                await stream.read_into(msg_recv_list[i].data)
+            else:
+                msg_recv_list.append(await stream.read_bytes(self.args.n_bytes))

             stop = monotonic()
             if i >= self.args.n_warmup_iter:
```

</details>

wence-: I think I meant something like:
```
if self.args.reuse_alloc:
   await stream.read_into(msg_recv_list[i].data)
else:
   buf = self.xp.empty(self.args.n_bytes, dtype="u1")
   await stream.read_into(buf.data)
   msg_recv_list.append(buf)
```

pentschev: I see your point, I agree this makes more sense, we are not necessarily timing the allocation this way. If you don't mind I'll do it in a follow-up PR, then I'll do a similar change to the UCX-Py benchmarks together with this, that will ensure that all of the backends are timing things in an equivalent manner.

pentschev: Opened https://github.com/rapidsai/ucx-py/pull/880 to cover this, but this needs to go in first.

ucp/benchmarks/backends/tornado.py
```diff
@@ -0,0 +1,121 @@
...
+                    break
+
+        return port
+
+    async def run(self):
+        args = self.args
+        xp = self.xp
+
+        global event
+        event = asyncio.Event()
```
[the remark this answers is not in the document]
wence-: If I apply this patch on top:
```patch
diff --git a/ucp/benchmarks/backends/tornado.py b/ucp/benchmarks/backends/tornado.py
index 8bccb51..8238e2c 100644
--- a/ucp/benchmarks/backends/tornado.py
+++ b/ucp/benchmarks/backends/tornado.py
@@ -35,16 +35,10 @@ class TornadoServer(BaseServer):
         args = self.args
         xp = self.xp
 
-        global event
         event = asyncio.Event()
 
         class TransferServer(TCPServer):
-            def __init__(self, event, *args, **kwargs):
-                super().__init__(*args, **kwargs)
-                self.event = event
-
             async def handle_stream(self, stream, address):
-                global event
                 msg_recv_list = []
                 if not args.reuse_alloc:
                     for _ in range(args.n_iter + args.n_warmup_iter):
@@ -64,10 +58,10 @@ class TornadoServer(BaseServer):
                         print(e)
                         break
 
-                self.event.set()
+                event.set()
 
         # Set max_buffer_size to 1 GiB for now
-        server = TransferServer(event, max_buffer_size=1024**3)
+        server = TransferServer(max_buffer_size=1024**3)
         port = self._start_listener(server, args.port)
 
         self.queue.put(port)
```

I seem to be able to run with `python -m ucp.benchmarks.send_recv -l tornado -o numpy` without problems?

wence- (approved):
Apart from the `global` annotation, I think this looks good now. Thanks!

ucp/benchmarks/backends/tornado.py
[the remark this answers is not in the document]
pentschev: You're right, I don't know what I was doing. Nevertheless, this is now applied in https://github.com/rapidsai/ucx-py/pull/874/commits/70438edf265be6e49959db3b512912a4a10cf68f . Thanks for the patch!

pentschev:
Thanks @wence- for the thorough review!

PR #878, opened 2022-09-21
Correctly parse bytes argument when dask unavailable

wence-:
We need to ensure the value we read is converted to int.


pentschev (approved):
LGTM, thanks @wence- !

PR #889, opened 2022-10-18
Fix Cython exceptions

pentschev:
A crash was reported after `tmpfile` in `create_text_fd()` raised an exception, causing a segmentation fault. This was due to the missing `except *` Cython handle for that function. The `_err_cb` also used `except *` handle which is unnecessary, that is now removed.


jakirkham (approved):

jakirkham:
Thanks Peter! 🙏

pentschev:
Thanks John for quick review/approval!

PR #893, opened 2022-11-02
Replace tests_require with extras_require.

vyasr:
`tests_require` was deprecated by setuptools over three years ago now. `extras_require["test"]` is the standard replacement.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,10 +69,12 @@ install_requires = [
     "pynvml",
 ]
 
-tests_require = [
-    "pytest",
-    "pytest-asyncio",
-]
+extras_require = {
+    "test": [
+        "pytest",
+        "pytest-asyncio",
+    ],
+}
 
 setup(
     name="ucx-py",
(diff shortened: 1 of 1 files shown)


madsbk (approved):
LGTM, thanks @vyasr

jakirkham:
Generally agree with this change, but do you have a doc reference where they recommend this syntax?

For example could see someone saying this should be `"tests"` instead of `"test"`. Here someone went with `"testing"` ( https://github.com/shimming-toolbox/shimming-toolbox/commit/c9bab401abc4e64b36026052b3d90f987ee04d20 ).

Spent a bit looking for one myself and outside of issue ( https://github.com/pypa/setuptools/issues/931 ) and this suggestion ( https://github.com/pypa/pip/issues/1197#issuecomment-228939212 ) am not seeing much in the way of guidance, which is a bit disappointing (hope I'm just overlooking something though).

vyasr:
No, unfortunately I'm not aware of any official recommendation for this either. I went with the key name that's used consistently in RAPIDS projects since that seems like the most consistent choice we could make.

jakirkham (approved):

jakirkham:
Thanks Vyas! 🙏

PR #896, opened 2022-11-07
Remove `psutil` dependency

pentschev:
In the past `psutil` was used to generate a random port number for the listener, but this is not anymore used as now UCX's random port generation is utilized, but the dependency to `psutil` was not removed. This PR removes the not anymore used dependency.


raydouglass (approved):

jakirkham:
Thanks all! 🙏

PR #901, opened 2022-11-16
Add wheel builds

vyasr:
This PR adds support for building wheels of ucx-py.


setup.py
```diff
@@ -56,7 +43,10 @@ def get_ucp_version():
 cmdclass = versioneer.get_cmdclass(cmdclass)
 
 setup(
+    name="ucx-py" + os.getenv("RAPIDS_PY_WHEEL_CUDA_SUFFIX", default=""),
```
jakirkham, line 46: Am curious if this works given it is also in `pyproject.toml`. If so, great!

FWIW had tried adding `name` to the [`dynamic` attribute list]( https://github.com/rapidsai/ucx-py/blob/branch-0.29/pyproject.toml#L26-L28 ), which didn't work. AIUI this lines up with [the `dynamic` specification in PEP 621]( https://peps.python.org/pep-0621/#dynamic ):

> A build back-end MUST raise an error if the metadata specifies the `name` in `dynamic`.

jakirkham: Should add if this doesn't work we may want to consider:

* Templating `pyproject.toml`
* Having multiple `pyproject.toml`s and using one based on CUDA version
* Some other solution that doesn't depend on changing `name` (for example using [local version identifiers]( https://peps.python.org/pep-0440/#local-version-identifiers ), altering other metadata, including a metapackage dependency, etc.)

vyasr: Ah good point I forgot about that. I do not think that the package name is being modified by the suffix at present. I'm waiting to get wheels built to verify everything though. 

I would be very happy if we could come up with a scheme that did not require this environment variable-based approach. I don't know how well anything plays with multiple or templated pyproject.toml files, though, and my understanding of local version identifiers is that they are meant to denote versions that are ABI compatible. I am not familiar with how else this might be done.

vyasr: I will say that the expedient solution would be to simply remove the name from pyproject.toml altogether. Certainly not my favorite solution, but I would be OK with it temporarily to get this release done while we search for a longer-term solution. We definitely need one across all of RAPIDS.

vyasr: I've temporarily removed the name from pyproject.toml because I want to get an up-to-date ucx build for testing raft wheels, but then I can revert until we can come up with a consensus on the best solution for this problem.

vyasr: It looks like that isn't valid either, the name must exist in that section. I'll undo this change and modify the downstream library's dependencies for now.

jakirkham: Possibly silly question, but could we just hardcode the value? Know we are looking to handle other versions in the future, but right now it is just one value correct? Maybe we can skip customizing it until we need that

vyasr: We could. Will it affect anything with the conda build if the pyproject.toml name does not match the meta.yaml name?

jakirkham: My guess is probably not, but let's try it and see 🙂

setup.py
```diff
@@ -52,11 +39,18 @@ def get_ucp_version():
     ),
 ]
 
-cmdclass = dict(build_ext=build_ext)
-cmdclass = versioneer.get_cmdclass(cmdclass)
+cmdclass = dict()
+cmdclass.update(versioneer.get_cmdclass())
+cmdclass["build_ext"] = build_ext
```
jakirkham, line 44: Why is this needed? The former should already use `build_ext`. Please see [this doc]( https://github.com/python-versioneer/python-versioneer/blob/83f20fdf886df7a6089fca3bb7fc3c33198bd629/INSTALL.md?plain=1#L132 )

setup.py
```diff
@@ -52,11 +39,18 @@ def get_ucp_version():
...
+    # TODO: At present the ucx-py naming scheme will not match the rest of the
+    # RAPIDS wheels. It is hardcoded in pyproject.toml, so overriding it here
+    # will not work and wheels will always just be called ucx_py.
+    # name="ucx-py" + os.getenv("RAPIDS_PY_WHEEL_CUDA_SUFFIX", default=""),
     ext_modules=ext_modules,
     cmdclass=cmdclass,
-    version=versioneer.get_version(),
+    version=os.getenv(
+        "RAPIDS_PY_WHEEL_VERSIONEER_OVERRIDE", default=versioneer.get_version()
+    ),
```
jakirkham, line 55: Maybe we should check this environment variable above and use it (skipping versioneer and skipping `cmdclass` updates)

setup.py
vyasr: Hmm I'm not sure why I did this. May have been a merge from an old commit. I'll undo this change.

setup.py
vyasr: I've switched to the versioneer patch that we're now using in all the libraries to ensure that we get a consistent version in nightlies as well. Let me know if that addresses your comment here sufficiently or if you also want to try to avoid cmdclass changes (I don't think I'm doing special for the cmdclass when building wheels anymore, that code is identical to before now that I've undone the above change).

setup.py
vyasr: Trying it now!

vyasr: OK it seems like it worked fine. When I look at the build logs ([here's one example](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fucx-py%2Fprb%2Fucx-py-cpu-build-arm64/detail/ucx-py-cpu-build-arm64/552/pipeline)) I see a few references to the `cu11` suffix, mostly coming from metadata files (egg-info, PKG-INFO, etc). @jakirkham does any of that look concerning to you in terms of how it will get packaged up by conda, or do you think the hardcoded solution should be fine?

wence- (commented):
Minor nits

.github/build-scripts/patch-wheel.sh
```diff
@@ -0,0 +1,58 @@
...
+hash=$(sha256sum ${src} | awk '{print substr($1, 0, 8)}')
+target=$(basename $(readlink -f ${src}))
+
+mv ${target} ${target/libcudart/libcudart-${hash}}
+while readlink ${src} > /dev/null; do
+    target=$(readlink ${src})
+    ln -s ${target/libcudart/libcudart-${hash}} ${src/libcudart/libcudart-${hash}}
+    rm -f ${src}
+    src=${target}
+done
```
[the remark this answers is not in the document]
vyasr: Good call, I'll add a comment. I wouldn't call this rebuilding symbols, it's just renaming the libraries.

setup.py
```diff
@@ -6,30 +6,31 @@
...
 
-include_dirs = [os.path.dirname(get_python_inc())]
-library_dirs = [get_config_var("LIBDIR")]
-libraries = ["ucp", "uct", "ucm", "ucs"]
-extra_compile_args = ["-std=c99", "-Werror"]
+# Patch versioneer version for wheel builds.
+if "RAPIDS_PY_WHEEL_VERSIONEER_OVERRIDE" in os.environ:
+    orig_get_versions = versioneer.get_versions
+
+    version_override = os.environ["RAPIDS_PY_WHEEL_VERSIONEER_OVERRIDE"]
```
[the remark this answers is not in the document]
vyasr: Done.

setup.py
```diff
@@ -6,30 +6,31 @@
...
+    def get_versions():
+        data = orig_get_versions()
+        data["version"] = version_override
+        return data
 
-def get_ucp_version():
-    with open(include_dirs[0] + "/ucp/api/ucp_version.h") as f:
-        ftext = f.read()
-        major = re.findall("^#define.*UCP_API_MAJOR.*", ftext, re.MULTILINE)
-        minor = re.findall("^#define.*UCP_API_MINOR.*", ftext, re.MULTILINE)
```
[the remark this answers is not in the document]
vyasr: Just so we're on the same page, I've already removed this in this PR because it was unused. I assume you're just explaining why it can go, not requesting any particular changes?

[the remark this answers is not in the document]
wence-: yep

jakirkham (commented):
Thanks Vyas for all the hard work here! 🙏

Had some thoughts on simplifying the version handling below

setup.py
jakirkham: Yeah I don't think it should be an issue

Maybe there's some value in having this metadata match the wheel somewhat as that might protect against `pip` clobbering the Conda package. Could be worth testing this though

setup.py
```diff
@@ -56,6 +63,10 @@ def get_ucp_version():
 cmdclass = versioneer.get_cmdclass(cmdclass)
 
 setup(
+    # TODO: At present the ucx-py naming scheme is not dynamic and will not
+    # support different CUDA major versions if we need to build wheels. It is
+    # hardcoded in pyproject.toml, so overriding it here will not work.
+    # name="ucx-py" + os.getenv("RAPIDS_PY_WHEEL_CUDA_SUFFIX", default=""),
     ext_modules=ext_modules,
     cmdclass=cmdclass,
     version=versioneer.get_version(),
```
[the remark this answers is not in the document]
wence-: The rationale for this is as for the approach we took in [dask-cuda](https://github.com/rapidsai/dask-cuda/pull/948). I copy the description:

> The GIT_DESCRIBE_TAG and VERSION_SUFFIX environment variables are used
to control the name and version of the created conda/pypi package.
They should, however, not be used to control the version of the
installed package by overriding the versioneer cmdclass since that
leaves an unmodified _version.py file in the installed package
directory. A consequence is that the version reported by
dask_cuda.__version__ is "0+unknown".

> We cannot always use the versioneer-provided cmdclass unmodified since
PEP440 specifically forbids PyPI from accepting packages that have local
version identifiers (as used by versioneer). To get around this, when setup.py
detects it is building a PyPI package (GIT_DESCRIBE_TAG is in the environment),
patch the version returned from versioneer with a PyPI-compatible one.

Specifically. If we don't use versioneer here, then the dynamic code in `__init__.py` which imports `_version` and calls `get_versions()` will report nonsense for wheel packages.

pyproject.toml
```diff
@@ -53,6 +53,10 @@ classifiers = [
 test = [
     "pytest",
     "pytest-asyncio",
+    "cloudpickle",
+    "dask",
+    "distributed",
+    "cupy-cuda11x",
```
wence-, line 59: This is presumably making explicit what was previously implicit

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,61 @@
...
+      # copy custom auditwheel script to known location with '{package}' placeholder
+      # in the auditwheel repair command, the '{package}' placeholder can't be used
+      cibw-before-build: "cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh"
+
+      auditwheel-repair-command-amd64: "auditwheel --verbose repair -w {dest_dir} --plat manylinux_2_17_x86_64 {wheel} && /patch-wheel.sh {dest_dir}/*.whl"
+      auditwheel-repair-command-arm64: "auditwheel --verbose repair -w {dest_dir} --plat manylinux_2_31_aarch64 {wheel} && /patch-wheel.sh {dest_dir}/*.whl"
+
+      test-docker-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g"
+
+      test-before-arm64: "pip install cupy-cuda11x -f https://pip.cupy.dev/aarch64"
```
wence-, line 58: I guess this is not required on x86 because pypi hosts wheels in that case.

wence- (approved):
Thanks!

pyproject.toml
vyasr: Yes exactly, and it's important for the user experience if they do `pip install ucx-py` to get all the dependencies pulled in (and we leverage it in the tests of the wheels now as well).

.github/workflows/wheels.yml
vyasr: Correct. There is unfortunately no way to encode a dependency of this form (`-f` basically says "search all links on a page") into a requirements list. It is possible to specify an exact wheel `$dependency @ $URL_OF_WHEEL_FILE`) but that's not equivalent. The old way to do this was with `dependency_links`, but that was deprecated years ago, so manual installation like this is the best that we can do.

setup.py
```diff
@@ -6,30 +6,37 @@
...
+        data = orig_get_versions()
+        data["version"] = version_override
+        return data
 
-def get_ucp_version():
-    with open(include_dirs[0] + "/ucp/api/ucp_version.h") as f:
-        ftext = f.read()
-        major = re.findall("^#define.*UCP_API_MAJOR.*", ftext, re.MULTILINE)
-        minor = re.findall("^#define.*UCP_API_MINOR.*", ftext, re.MULTILINE)
+    versioneer.get_versions = get_versions
```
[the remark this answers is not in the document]
vyasr: I'll defer to @wence- here, but the answer here is related to his answer to your other query about why not directly use the version: this hack ensures that we get the right version encoded in `_version.py` so that `ucp.__version__` is correct when users import and use it.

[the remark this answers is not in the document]
wence-: Yes, it's all part of the same dance.

The other approach would be to rewrite `_version.py` in the wheel-building process to fill in the version by hand. But this is kind of what versioneer does any, so this monkeypatch is simpler.

.github/workflows/wheels.yml
```diff
@@ -48,7 +48,9 @@ jobs:
 
       # copy custom auditwheel script to known location with '{package}' placeholder
       # in the auditwheel repair command, the '{package}' placeholder can't be used
-      cibw-before-build: "cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh"
+      # Also rewrite the name in the pyproject.toml file.
+      # TODO: Eventually this needs to be done more generically.
+      cibw-before-build: "cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh && sed -i 's/name = \"ucx-py\"/name = \"ucx-py-cu11\"/g'  {package}/pyproject.toml"
```
jakirkham, line 53: Was thinking we could also capture this in a patch we apply as part of the build. No strong feelings about that or this approach though

setup.py
vyasr: @jakirkham in order to avoid this problem entirely, I've changed the name back in pyproject.toml and I've modified the wheels workflow to modify the pyproject.toml file at build time. I think this approach is cleaner for this release and avoids these concerns. For 23.02 I'm going to try to figure out the appropriate long-term solution, but I think this is suitable for now.

setup.py
```diff
@@ -56,6 +63,10 @@ def get_ucp_version():
 cmdclass = versioneer.get_cmdclass(cmdclass)
```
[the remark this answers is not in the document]
vyasr: Given the discussions around `versioneer.get_version` I think it's more appropriate to leave all of this unconditional and make the versioneer monkey patch the only conditional change.

.github/workflows/wheels.yml
vyasr: Let's stick to this approach for now and punt determining an "optimal" long-term solution for the next release. I don't know if we want to store multiple patches in the repo since that will create multiple sources of truth (the GHA workflow will also need to have access to the current CUDA version, for instance). I think it will be cleaner in the long run to use this sort of scripting approach that reacts to the values of inputs to the workflow.

setup.py
vyasr: [I've opened a discussion on the Python forums](https://discuss.python.org/t/dynamic-project-names-and-pep-621/21359) to continue this further. I'd like to see what solutions have been considered for this in the past before we try and come up with one of our own to apply across RAPIDS. The one that I've implemented in ucx-py seems fine for now.

jolorunyomi (approved):

vyasr:
@jakirkham do you have further concerns here about the version management (or anything else)?

vyasr:
In the interest of getting things in for code freeze I'm going to go ahead and merge this. If there are any outstanding concerns about the version handling we can address that in the next release.

vyasr:
@gpucibot merge

vyasr:
Ah I guess I don't have write access to this repository. @wence- or @pentschev could you please merge this? Thanks!

jakirkham:
@gpucibot merge

PR #903, opened 2022-11-28
Move `get_address` and improve compatibility and error handling

pentschev:
Move `get_address` into `utils.py`, which is a pure-Python function that does not need to live in a Cython source file.

Optionally attempt to query for `docker*` network interface, if other known interfaces fail in a Docker container. If still no valid IP address could be determined, raise an exception with information on how the user can manually setup to proceed.

### DIFF
--- a/ucp/_libs/utils.py
+++ b/ucp/_libs/utils.py
@@ -85,6 +85,11 @@ def get_address(ifname=None):
     address : str
         The inet addr associated with an interface.
 
+    Raises
+    ------
+    RuntimeError
+        If a network address could not be determined.
+
     Examples
     --------
     >>> get_address()
@@ -119,6 +124,11 @@ def get_address(ifname=None):
                     return _get_address(i)
                 except OSError:
                     pass
+
+        raise RuntimeError(
+            "A network address could not be determined, an interface that has a valid "
+            "IP address with the environment variable `UCXPY_IFNAME`."
+        )
 
     if ifname is None:
         ifname = os.environ.get("UCXPY_IFNAME")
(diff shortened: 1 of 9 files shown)


pentschev:
Waiting for confirmation from @quasiben this works in his setup.

ucp/_libs/utils.py
```diff
@@ -60,3 +66,64 @@ def print_multi(values, key_length=25):
...
+    def _get_address(ifname):
+        ifname = ifname.encode()
+        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+            return socket.inet_ntoa(
+                fcntl.ioctl(
+                    s.fileno(), 0x8915, struct.pack("256s", ifname[:15])  # SIOCGIFADDR
+                )[20:24]
+            )
+
+    def _try_interfaces():
```
jakirkham, line 106: Do these need to be closures? Or could they be top-level functions that just get called here?

pentschev: They could, but they're exclusively meant to be part of the `get_address()` implementation, we don't want people to be calling them directly by accident. Are there any specific reasons to suggest this change?

Note: I'm just moving the code around, not changing the implementation in this PR.

wence- (approved):

ucp/_libs/utils.py
jakirkham: No more thoughts from me

jakirkham:
@gpucibot merge

jakirkham:
Thanks Peter! 🙏

pentschev:
Thanks for reviews @jakirkham and @wence- !

PR #904, opened 2022-11-29
Upgrade Python version to 3.8 in docs

pentschev:
Python 3.6 and above _should_ still work, but are not actively tested, thus upgrading to 3.8 which is actually tested should be encouraged.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -27,14 +27,14 @@ With GPU support:
 ::
 
     conda create -n ucx -c conda-forge -c rapidsai \
-      cudatoolkit=<CUDA version> ucx-proc=*=gpu ucx ucx-py python=3.7
+      cudatoolkit=<CUDA version> ucx-proc=*=gpu ucx ucx-py python=3.8
 
 Without GPU support:
 
 ::
 
     conda create -n ucx -c conda-forge -c rapidsai \
-      ucx-proc=*=cpu ucx ucx-py python=3.7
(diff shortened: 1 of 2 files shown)


wence-:
Looks good, should we also update the versions in `ci/local/build.sh` and `ci/local/README.md` ?

wence-:
I guess 3.6 is still around for colab (it's now 11 months since it was end-of-lifed)?

pentschev:
> Looks good, should we also update the versions in `ci/local/build.sh` and `ci/local/README.md` ?

I thought about that, but AFAIK `ci/local/build.sh` is deprecated and maybe we should just drop it entirely, I'll check with ops.

> I guess 3.6 is still around for colab (it's now 11 months since it was end-of-lifed)?

It's there for one particular team who still uses that, but hopefully we'll get them to upgrade by EOY (or at least that's their goal).

pentschev:
> > Looks good, should we also update the versions in `ci/local/build.sh` and `ci/local/README.md` ?
> 
> I thought about that, but AFAIK `ci/local/build.sh` is deprecated and maybe we should just drop it entirely, I'll check with ops.

I've confirmed internally with the ongoing migration to GH Actions that script is now obsolete and will be removed along with other changes from this repo soon. Therefore I'll not update that.

Thanks for the review @wence- !

pentschev:
@gpucibot merge

pentschev:
@wence- actually, could you approve? 😄

wence- (approved):

PR #909, opened 2022-12-30
Update wheel workflows to incorporate PR builds

sevagh:
This creates new GHA workflow files to build wheels in more situations (e.g. PR builds).


.github/workflows/build.yaml
```diff
@@ -18,43 +17,49 @@ on:
...
+      wheel_versioneer_override:
+        type: string
+        default: ''
+      wheel_build_tag:
+        type: string
+        default: ''
 
 concurrency:
-  group: "ucx-py-${{ github.workflow }}-${{ github.ref }}"
+  group: ${{ github.workflow }}-${{ github.ref }}
```
vyasr, line 31: Hmm I didn't notice this difference in the other PRs, but here because wheels.yaml->build.yaml is treated like a file rename I see that the hardcoding is removed from the group. IIRC we originally added this because using the same shared action from different repos resulted in them all being part of the same concurrency group and therefore canceling each other, right? Is there a reason that isn't necessary any more? I assume it's working correctly or we'd be seeing conda CI from different repos conflicting all over the place, I'm just surprised.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,40 @@
...
+    needs: wheel-build-ucx-py
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-test.yml@upkeep/manylinux-improvements
+    with:
+
+      build_type: pull-request
+      package-name: ucx_py
+      test-docker-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g"
+      test-before-arm64: "pip install cupy-cuda11x -f https://pip.cupy.dev/aarch64"
+      test-unittest: "pytest -k 'not test_send_recv_am' --cache-clear -vs ./ucp/_libs/tests/ && pytest --cache-clear -vs ./tests"
```
vyasr, line 39: Let's link to https://github.com/rapidsai/ucx-py/issues/797 here and in test.yaml so we know why we're skipping this test and when we can remove the filter.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,40 @@
...
+      test-smoketest: "" #"python ci/wheel_smoke_test_ucx_py.py"
```
vyasr, line 40: Given that UCX unit tests only take 2 minutes and that in fact the majority of the test run time is just spinning up the citestwheel container, I think we can just run the entire unit test suite as the smoke test in this case. Also, given the nature of the communication APIs in UCX, I feel like regularly testing on different architectures is even more valuable here than for other RAPIDS projects.

vyasr (changes requested):

.github/workflows/build.yaml
sevagh: Our org use of GitHub Actions has evolved since then, with the following improvements:
(@ajschmidt8 gave me this info through slack):
> solved by adding the nightly- prefix to the top-level pipeline as opposed to the lower-level pipelines
https://github.com/rapidsai/actions/blob/main/.github/workflows/nightly-pipeline.yaml#L12
  group: nightly-${{ github.workflow }}-${{ github.ref }}

>however, this isn't needed anymore since we've switched to using workflow_dispatch events instead of workflow_call for the repo pipelines via the trigger-workflow-and-wait composite action that we chatted about yesterday
https://github.com/rapidsai/actions/blob/main/.github/workflows/nightly-pipeline.yaml#L35

ajschmidt8:
rerun tests

.github/workflows/build.yaml
vyasr: This change is nice!

vyasr (approved):

vyasr:
I think this is good to go

PR #911, opened 2023-01-10
Update GH Action Workflows

ajschmidt8:
This PR updates the workflows to change some of the events used to trigger the `build.yaml` and `test.yaml` workflows.

These changes are necessary due to a recent change in how we call these workflows from the nightly pipeline in the https://github.com/rapidsai/actions.

See https://github.com/rapidsai/actions/pull/7 for more details.


ajschmidt8:
Admin merging since these changes are benign and not tested in CI.

PR #914, opened 2023-01-13
Add GitHub Actions Workflows

ajschmidt8:
This PR adds GitHub Actions workflows to `ucx-py`.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,7 +5,7 @@ on:
     branches:
       - "branch-*"
     tags:
-      - v[0-9][0-9].[0-9][0-9].[0-9][0-9]
+      - v[0-9].[0-9][0-9].[0-9][0-9]
   workflow_dispatch:
     inputs:
       branch:
@@ -26,6 +26,23 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  conda-python-build:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@main
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      branch: ${{ inputs.branch }}
+      date: ${{ inputs.date }}
+      sha: ${{ inputs.sha }}
+  upload-conda:
+    needs: [conda-python-build]
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-upload-packages.yaml@main
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      branch: ${{ inputs.branch }}
+      date: ${{ inputs.date }}
+      sha: ${{ inputs.sha }}
   wheel-build:
     secrets: inherit
     uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@main
@@ -40,7 +57,7 @@ jobs:
       # in the auditwheel repair command, the '{package}' placeholder can't be used
       # Also rewrite the name in the pyproject.toml file.
       # TODO: Eventually this needs to be done more generically.
-      cibw-before-build: "cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh && sed -i 's/name = \"ucx-py\"/name = \"ucx-py-cu11\"/g'  {package}/pyproject.toml"
+      cibw-before-build: 'cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh && sed -i ''s/name = "ucx-py"/name = "ucx-py-cu11"/g''  {package}/pyproject.toml'
       post-repair-amd64: "/patch-wheel.sh {dest_dir}/*.whl"
       post-repair-arm64: "/patch-wheel.sh {dest_dir}/*.whl"
   wheel-publish:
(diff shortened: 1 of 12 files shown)


raydouglass (approved):

.github/workflows/build.yaml
```diff
@@ -40,7 +57,7 @@ jobs:
       # in the auditwheel repair command, the '{package}' placeholder can't be used
       # Also rewrite the name in the pyproject.toml file.
       # TODO: Eventually this needs to be done more generically.
-      cibw-before-build: "cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh && sed -i 's/name = \"ucx-py\"/name = \"ucx-py-cu11\"/g'  {package}/pyproject.toml"
+      cibw-before-build: 'cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh && sed -i ''s/name = "ucx-py"/name = "ucx-py-cu11"/g''  {package}/pyproject.toml'
```
pentschev, line 60: Sorry for my ignorance of the yaml structure for GH, but is this duplicated single-quotes correct? Specifically around the sed expression. I would normally write it as:

sed -i "s/name = \"ucx-py\"/name = \"ucx-py-cu11\"/g"  {package}/pyproject.toml

.github/workflows/pr.yaml
```diff
@@ -12,26 +12,44 @@ concurrency:
...
   wheel-build:
+    needs: checks
     secrets: inherit
     uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@main
     with:
       build_type: pull-request
       package-name: ucx_py
       package-dir: .
-      cibw-before-build: "cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh && sed -i 's/name = \"ucx-py\"/name = \"ucx-py-cu11\"/g'  {package}/pyproject.toml"
+      cibw-before-build: 'cp {package}/.github/build-scripts/patch-wheel.sh /patch-wheel.sh && sed -i ''s/name = "ucx-py"/name = "ucx-py-cu11"/g''  {package}/pyproject.toml'
```
pentschev, line 45: Same question about the duplicate single-quotes here.

dependencies.yaml
```diff
@@ -0,0 +1,83 @@
...
+    output: none
+    includes:
+      - checks
+      - py_version
+channels:
+  - rapidsai
+  - rapidsai-nightly
+  - dask/label/dev
+  - conda-forge
+  - nvidia
```
pentschev, line 27: Are these channels here for testing as well, or just for building/distributing? In the latter case both `dask/label/dev` and `nvidia` can be removed, as UCX-Py doesn't (or shouldn't) have any hard dependencies on either.

sevagh (approved):

pentschev (changes requested):
Thanks @ajschmidt8 , I've left a few questions there and I've noticed several of the tests are complaining about missing `pytest-asyncio`, could we have it added? Besides that, everything looks good to me.

.github/workflows/build.yaml
ajschmidt8: This wasn't actually a change I made manually. VS Code did this automatically when I saved the file.

After some Googling, I found the article below which talks about escaping the `'` character by doubling it within a larger single-quoted string:

- https://docs.octoprint.org/en/master/configuration/yaml.html#:~:text=In%20single%20quoted%20strings%20the,just%20indent%20the%20following%20lines

Ultimately, I'm not sure if the outer quotes here are necessary at all, since in YAML the following string values are equivalent:

```yaml
somekey: a value
anotherkey: "a value"
```

Probably not worth dwelling on too much unless you have a strong preference one way or the other.

pentschev: No, I don't have a preference, just thought it was unusual and not sure if that was even correct. If it works I don't mind keeping it as is.

dependencies.yaml
ajschmidt8: These channels are used when we generated the `environment.yaml` files used for our test environments in CI. (e.g. [here](https://github.com/rapidsai/ucx-py/actions/runs/3914114616/jobs/6691144626#step:6:18)).

I just copy/pasted the default channels we use from other libraries but I can remove `dask` and `nvidia` if they're extraneous.

pentschev: The reason I commented is really because I don't know exactly what they're used for. If they're not going to pull any potential hard dependencies into a build (which probably won't happen indeed) then we can keep them. IOW, users may not want CUDA, RAPIDS or Dask at all, and thus we don't want to introduce any such dependencies.

ajschmidt8: Gotcha. Yeah, they're just used for the files that get generated by the new [rapids-dependency-file-generator](https://github.com/rapidsai/dependency-file-generator) tool and `dependencies.yaml` file.

For example, here's an environment file in `cudf` that was generated by the tool:

- https://github.com/rapidsai/cudf/blob/ec7f8c6b070e421f4efa748a2da08465651701aa/conda/environments/all_cuda-118_arch-x86_64.yaml#L1-L9

The tool is also used to generate testing environments for CI (e.g. [here](https://github.com/ajschmidt8/ucx-py/blob/55c8cb02c48d567506e7250471a2e19e722f899e/ci/test_python.sh#L8-L14) and corresponding [log output](https://github.com/rapidsai/ucx-py/actions/runs/3914114616/jobs/6691144626#step:6:18)).

This lets us avoid having to repeat dependencies in several areas.

In the long term, we'd like to continue iterating on the tool so that `dependencies.yaml` can be a single source of truth for all dependencies of a given repository, but for now it does just enough to support keeping things DRY for GH Actions.

pentschev:
LGTM, thanks @ajschmidt8 !

pentschev (approved):

PR #930, opened 2023-02-23
Consolidate metadata in `pyproject.toml`

jakirkham:
Also use this metadata to fill out the Conda recipe.


pyproject.toml
```diff
@@ -35,7 +35,7 @@ license = { text = "BSD-3-Clause" }
 requires-python = ">=3.6"
 dependencies = [
     "numpy",
-    "pynvml",
+    "pynvml >=11.4.1",
```
wence-, line 38: Why this change?

conda/recipes/ucx-py/conda_build_config.yaml
```diff
@@ -1,12 +1,7 @@
 c_compiler_version:
   - 9
-
 cxx_compiler_version:
   - 9
 
-cython:
-  - " >=0.29.14,<3.0.0a0"
-pynvml:
-  - " >=11.4.1"
```
jakirkham, line 10: We have moved this version constraint to `pyproject.toml` where it is now used for both conda & wheel packages

pyproject.toml
jakirkham: Maybe this comment ( https://github.com/rapidsai/ucx-py/pull/930#discussion_r1115918109 ) helps explain it?

ajschmidt8 (approved):

pyproject.toml
wence-: Aha, thanks.

conda/recipes/ucx-py/meta.yaml
```diff
@@ -41,10 +44,12 @@ test:
...
 
 about:
-  home: https://rapids.ai/
-  license: BSD-3-Clause
-  license_family: BSD
-  license_file: ../../../LICENSE
-  summary: Python bindings for UCX
-  dev_url: https://github.com/rapidsai/ucx-py
-  doc_url: https://ucx-py.readthedocs.io
+  home: {{ data.get("project", {}).get("urls", {}).get("Homepage", "") }}
```
wence-, line 47: Should the homepage point at the docs site? I don't really mind one way or the other, WDYT?

pyproject.toml
```diff
@@ -35,7 +35,7 @@ license = { text = "BSD-3-Clause" }
 requires-python = ">=3.6"
```
wence-, line 35: Note for a separate discussion, I think this is still true, but we may want to consider moving to 3.8+ (I don't recall if we had some downstream users who were still on 3.6 though).

wence- (approved):
Thanks @jakirkham!

conda/recipes/ucx-py/meta.yaml
vyasr: pyproject.toml supports storing arbitrary URLs in the table, and the repo is already stored under the Source key as well as the Homepage key. As long as all the relevant pages are somewhere under URLs, I don't have a strong preference as to which we designate as the "Homepage". I do think that we'd want to be consistent in what the conda page refers to as the homepage and what the PyPI index will show as the homepage, so any change should be made to pyproject.toml's Homepage so that it automatically propagates to here.

jakirkham: Yep agree with Vyas

Only think I'd add is we now also include `doc_url` to cover the documentation page. So that is referenced as well

conda/recipes/ucx-py/meta.yaml
```diff
@@ -21,18 +25,17 @@ requirements:
...
+    - pip
     - ucx
-    - versioneer >=0.24
+    {% for r in data.get("build-system", {}).get("requires", []) %}
+    - {{ r }}
+    {% endfor %}
   run:
-    - numpy
     - python
-    - ucx
```
jakirkham, line 33: Would also note that [`ucx` has `run_exports`]( https://github.com/conda-forge/ucx-split-feedstock/blob/dd05d5af3a6c4902093ecff980564b6605633124/recipe/meta.yaml#L38-L39 ) as of `1.12.1` ( https://github.com/conda-forge/ucx-split-feedstock/pull/92 ). As we are using `ucx` version `1.13.0`, we can rely on `run_exports` to add `ucx` as a dependency in `run` (with proper version constraints) based on what was used in `host`

pyproject.toml
pentschev: We still have some users running Python 3.6, but they're also using an older UCX-Py version. I checked and they're ok with us moving to Python 3.8+ on current development version. This is now being addressed in https://github.com/rapidsai/ucx-py/pull/933 .

pentschev (approved):
There are a couple conflicts to be resolved but otherwise LGTM, thanks @jakirkham !

pentschev:
Thanks John!

pentschev:
merge

jakirkham:
Thanks all! 🙏

PR #931, opened 2023-02-24
Stop using versioneer to manage versions and migrate wheel building for different CUDA versions

vyasr:
This change aligns ucx-py with the rest of RAPIDS plan for package and CUDA version management for conda and wheel builds going forward.


wence- (approved):
Thanks (you may wish to coordinate with @jakirkham's changes in #930)

ajschmidt8 (approved):

vyasr:
Thanks for that pointer @wence- I'm discussing analogous changes to rmm with John at the moment so will coordinate on making this a consistent cross-repo effort.

vyasr:
It looks like I don't have write permissions on the repo. @wence- when you get a chance would you mind merging this for me? Thanks!

pyproject.toml
```diff
@@ -18,14 +18,11 @@ requires = [
...
     "tomli; python_version < '3.11'",
-    "versioneer>=0.24",
 ]
 
 [project]
 name = "ucx-py"
-dynamic = [
-    "version",
-]
+version = "0.31.0"
```
pentschev, line 25: Sorry, I was OOO and am only seeing this now. I'm wondering here if it's possible to programmatically get the version, as it was done before, maybe by looking at the branch name? If not, who's responsible for keeping this up-to-date, is it automatically managed by CI or do we have to do it manually?

ucp/__init__.py
```diff
@@ -86,7 +85,7 @@ def _is_mig_device(handle):
     os.environ["UCX_MAX_RNDV_RAILS"] = "1"
 
 
-__version__ = _get_versions()["version"]
+__version__ = "0.31.0"
```
pentschev, line 88: Ditto.

pentschev (commented):
@vyasr sorry for the delay but had a few questions here, could you take a look?

pyproject.toml
jakirkham: Think the idea is that [the `update-version.sh` script]( https://github.com/rapidsai/ucx-py/blob/branch-0.31/ci/release/update-version.sh ) would handle this

ucp/__init__.py
jakirkham: Same as above

pyproject.toml
pentschev: Ah ok, got it now. Thanks John for clarifying!

vyasr: Yup that's right, thanks for answering here John. Long term I am hoping to do this even more programatically than we do now, but that has to do with improvements to update-version.sh and how the version there is produced.

PR #933, opened 2023-02-27
Update requirement to Python >= 3.8

pentschev:
We were maintaining support for Python 3.6 due to some users still requiring that, they are still on older UCX-Py releases so they are ok with us updating to Python 3.8+ on newer releases and they will try upgrading to a newer Python release soon.


wence- (approved):

pentschev:
Thanks both for review/approval!

jakirkham:
Thanks Peter! 🙏

PR #935, opened 2023-03-20
Update UCX pin to allow 1.14

pentschev:
https://github.com/conda-forge/ucx-split-feedstock/pull/111 was merged and UCX v1.14.0 is now available on conda-forge which we should support.


pentschev:
Thanks @raydouglass !

PR #938, opened 2023-03-30
Pin cupy and unset version

vyasr:
cupy just released version 12, which we do not yet support, so wheels CI must be pinned to require a lower version. We also decided not to overwrite the `__version__` in wheels on other RAPIDS repos, so this PR makes changes to match that.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -54,7 +54,7 @@ jobs:
       build_type: pull-request
       package-name: ucx_py
       test-docker-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g"
-      test-before-arm64: "pip install cupy-cuda11x<12.0.0 -f https://pip.cupy.dev/aarch64"
+      test-before-arm64: "pip install 'cupy-cuda11x<12.0.0' -f https://pip.cupy.dev/aarch64"
       # skipped test context: https://github.com/rapidsai/ucx-py/pull/909
       test-unittest: "pytest -k 'not test_send_recv_am' --cache-clear -vs ./ucp/_libs/tests/ && pytest --cache-clear -vs ./tests"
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
       build_type: pull-request
       package-name: ucx_py
       test-docker-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g"
-      test-before-arm64: "pip install cupy-cuda11x -f https://pip.cupy.dev/aarch64"
+      test-before-arm64: "pip install cupy-cuda11x<12.0.0 -f https://pip.cupy.dev/aarch64"
```
pentschev, line 57: I think this doesn't make sense. `cupy-cuda11x` has CUDA 11 in the name because it's supposed to be CUDA 11 only. Am I missing something?

.github/workflows/test.yaml
```diff
@@ -32,6 +32,6 @@ jobs:
       sha: ${{ inputs.sha }}
       package-name: ucx_py
       test-docker-options: "--cap-add CAP_SYS_PTRACE --shm-size=8g"
-      test-before-arm64: "pip install cupy-cuda11x -f https://pip.cupy.dev/aarch64"
+      test-before-arm64: "pip install cupy-cuda11x<12.0.0 -f https://pip.cupy.dev/aarch64"
```
pentschev, line 35: Same here.

pentschev (commented):
@vyasr could you clarify my questions here?

pentschev:
@vyasr ok, I mixed everything up very badly. `<12` is CuPy's version pin, not CUDA's, ignore my stupidity.

pentschev (approved):
Thanks @vyasr !

sevagh (approved):

PR #949, opened 2023-05-30
Fix version in release script and bump to `0.32.0`

pentschev:


ci/release/update-version.sh
```diff
@@ -39,3 +39,6 @@ sed_runner "s/cudf=.*/cudf=${NEXT_RAPIDS_VERSION}/g" dependencies.yaml
 for FILE in .github/workflows/*.yaml; do
   sed_runner "/shared-action-workflows/ s/@.*/@branch-${NEXT_RAPIDS_VERSION}/g" "${FILE}"
 done
+
+sed -i "s/^version = .*/version = \"${NEXT_FULL_TAG}\"/g" pyproject.toml
+sed -i "s/^__version__ = .*/__version__ = \"${NEXT_FULL_TAG}\"/g" ucp/__init__.py
```
charlesbluca, line 44: Should we use the `sed_runner` function here instead, or is there a reason we need to do this with `sed` directly?

```suggestion
sed_runner "s/^version = .*/version = \"${NEXT_FULL_TAG}\"/g" pyproject.toml
sed_runner "s/^__version__ = .*/__version__ = \"${NEXT_FULL_TAG}\"/g" ucp/__init__.py
```

raydouglass: Yes, it should use `sed_runner` for now.

raydouglass (approved):

charlesbluca (approved):

PR #951, opened 2023-05-30
Require Numba 0.57.0+ & NumPy 1.21.0+

jakirkham:
Align with the rest of RAPIDS on these requirements. Also needed for CUDA 12 support.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -64,7 +64,7 @@ Test Dependencies
 
     conda install -n ucx -c rapidsai -c nvidia -c conda-forge \
         pytest pytest-asyncio \
-        cupy "numba>=0.46" rmm \
+        cupy "numba>=0.57" rmm \
         distributed
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ authors = [
 license = { text = "BSD-3-Clause" }
 requires-python = ">=3.9"
 dependencies = [
-    "numpy",
+    "numpy >=1.21",
     "pynvml >=11.4.1",
 ]
 classifiers = [
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -68,7 +68,7 @@ dependencies:
     common:
       - output_types: [conda, requirements]
         packages:
-          - numpy
+          - numpy>=1.21
           - ucx
   test_python:
     common:
@@ -76,6 +76,6 @@ dependencies:
         packages:
           - cudf=23.06
           - cupy
-          - numba
+          - numba>=0.57
           - pytest
           - pytest-asyncio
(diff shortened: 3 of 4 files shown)


docs/source/install.rst
```diff
@@ -64,7 +64,7 @@ Test Dependencies
 
     conda install -n ucx -c rapidsai -c nvidia -c conda-forge \
         pytest pytest-asyncio \
-        cupy "numba>=0.46" rmm \
+        cupy "numba>=0.57" rmm \
         distributed
```
bdice, line 68: This list of testing dependencies doesn't align with the contents here: https://github.com/rapidsai/ucx-py/blob/6f7c1aec5889eb08e488a3c9e85cdf5ad25e1368/dependencies.yaml#L73-L81

Differences are in rmm/cudf and version pinnings for numba.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ authors = [
 license = { text = "BSD-3-Clause" }
 requires-python = ">=3.9"
 dependencies = [
-    "numpy",
+    "numpy >=1.21",
     "pynvml >=11.4.1",
```
bdice, line 35: `pynvml` isn't listed as a [run dependency](https://github.com/rapidsai/ucx-py/blob/6f7c1aec5889eb08e488a3c9e85cdf5ad25e1368/dependencies.yaml#L67-L72) in `dependencies.yaml`, but `ucx` is listed? We need to make sure this is using `rapids-dependency-file-generator`.

bdice (changes requested):
This needs to update `dependencies.yaml` and check on `rapids-dependency-file-generator`.

docs/source/install.rst
jakirkham: Good point. This should now be fixed. Please let me know what you think :)

pyproject.toml
jakirkham: Not sure what's going on there

The recipe [loads dependencies from `pyproject.toml`]( https://github.com/rapidsai/ucx-py/blob/6f7c1aec5889eb08e488a3c9e85cdf5ad25e1368/conda/recipes/ucx-py/meta.yaml#L36-L38 ) atm

Maybe there is a better way to make this more generally consistent

bdice: I'm fixing this and will file a PR shortly. It may supersede this PR if that's okay with you.

bdice: Opened this: https://github.com/rapidsai/ucx-py/pull/952

Still working on it but it's close to complete, I think.

jakirkham:
Superseded by PR ( https://github.com/rapidsai/ucx-py/pull/952 )

Thanks Bradley! 🙏

PR #952, opened 2023-05-30
Require Numba 0.57.0+ & NumPy 1.21.0+, use rapids-dependency-file-generator

### LINKED ISSUE #946: Enable skipped tests in CI when dependencies are available
While going through the CI logs of https://github.com/rapidsai/ucx-py/pull/945, we figured out that due to imports being handled with `pytest.importorskip`, some tests are being skipped entirely, and silently if the dependency is unavailable. This happens especially in wheels as the dependencies on `rmm` and `cudf` wheels is not explicitly expressed at the moment.

bdice:
Align requirements and dependency management with the rest of RAPIDS. Version updates are needed for CUDA 12 support.

Closes #951 (supersedes previous PR).
Closes #946 (adds missing optional dependencies to testing environment).

I also verified that `ci/release/update-version.sh` works as expected.


ci/release/apply_wheel_modifications.sh
```diff
@@ -6,8 +6,12 @@
 VERSION=${1}
 CUDA_SUFFIX=${2}
 
+# pyproject.toml versions
 sed -i "s/^version = .*/version = \"${VERSION}\"/g" pyproject.toml
+
+# pyproject.toml cuda suffixes
 sed -i "s/^name = \"ucx-py\"/name = \"ucx-py${CUDA_SUFFIX}\"/g" pyproject.toml
+sed -i "s/cudf/cudf${CUDA_SUFFIX}/g" pyproject.toml
```
bdice, line 14: I copied this approach from the way cudf installs rmm. https://github.com/rapidsai/cudf/blob/87a8ede8dcd9b6cd6e38c41f74daec316f48e7db/ci/release/apply_wheel_modifications.sh#L17

ci/release/update-version.sh
```diff
@@ -34,10 +36,17 @@ function sed_runner() {
     sed -i.bak ''"$1"'' $2 && rm -f ${2}.bak
 }
 
-sed_runner "s/cudf=.*/cudf=${NEXT_RAPIDS_VERSION}/g" dependencies.yaml
+DEPENDENCIES=(
+  cudf
+)
+for FILE in dependencies.yaml; do
```
bdice, line 42: This pair of `for` loops, while not strictly necessary, is meant to align with recent changes I made in RAFT and cuML. All of RAPIDS should have a similar structure for this replacement logic.

ci/release/apply_wheel_modifications.sh
pentschev, line 14:
```suggestion
```

pentschev: `cudf` isn't a dependency in UCX-Py, that's why this wasn't here, it's only a dependency for testing. See https://github.com/rapidsai/ucx-py/blob/branch-0.32/pyproject.toml and https://github.com/rapidsai/ucx-py/blob/branch-0.32/dependencies.yaml.

pentschev (commented):
Looks mostly good, I've left a few comments.

raydouglass (approved):

ci/release/apply_wheel_modifications.sh
bdice: This modifies the pyproject.toml specifically for the testing dependency on cudf. cudf is now listed in the optional test dependencies, but wasn’t previously. This change to dependencies.yaml/pyproject.toml/apply_wheel_modifications.sh fixed #946.

pentschev: Ah yes, it came later on in the list of changes and I didn't make the connection. Thanks for noticing my mistake.

pentschev (approved):
LGTM, thanks @bdice !

PR #954, opened 2023-05-31
Resolve forward merger conflicts.

bdice:
Closes #953.


pentschev (approved):
LGTM, thanks @bdice !

raydouglass (approved):

pentschev:
I was ready to merge and only then realized [tests failed](https://github.com/rapidsai/ucx-py/actions/runs/5135822615/jobs/9244909590#step:9:478) with error below:

```
/__w/ucx-py/ucx-py/.github/build-scripts/patch-wheel.sh: line 23: RAPIDS_PY_WHEEL_CUDA_SUFFIX: unbound variable
```

Do you know what is missing here @bdice ? We haven't seen that error in branch-0.32, thus I believe this may have been due to some recent change for 23.08?

jakirkham:
cc @vyasr

bdice:
I'm not sure what's going wrong here... I tried to trace where that environment variable is coming from but I don't have a working hypothesis for why this is broken.

vyasr:
That error is due to https://github.com/rapidsai/shared-action-workflows/pull/91. Would recommend looking at the changeset from that PR to see how to fix. The key is making a change like [this one](https://github.com/rapidsai/shared-action-workflows/pull/91/files#diff-6bf01961563a6bf4bcc304fc8bab59ea2f144c510cffb05fb83c940b3f7ec3c9R191). The relevant tool is in https://github.com/rapidsai/gha-tools/pull/55.

divyegala:
@vyasr I coincidentally ended up spotting this issue due to other reasons, and I tried to put up a PR that fixes this. https://github.com/rapidsai/ucx-py/pull/955

However, I just realized that [RAPIDS_CUDA_VERSION](https://github.com/rapidsai/ucx-py/pull/955/files#diff-c0cbf2f72a093d84c8cd993d21cdfa61769a081d4b525b0c0cf8e6b036ae1b27R23) is setup in the new wheel images I am working on. 

While the example link you shared has GHA variable `${{ matrix.ctk }}` that is not passed to downstream images. So we have no way as of now to generate the suffix.

Perhaps it will be in our best interests in moving to the new wheel images ASAP, even if it ultimately turns out that `ucx-py` does not need information about CUDA version since it is CUDA-agnostic.

pentschev:
Thanks @bdice and @divyegala for the work here!

PR #955, opened 2023-06-01
Fix usage of `RAPIDS_PY_WHEEL_CUDA_SUFFIX`

divyegala:
This env variable is not not defined anymore after https://github.com/rapidsai/gha-tools/pull/55 was merged

### DIFF
--- a/.github/build-scripts/patch-wheel.sh
+++ b/.github/build-scripts/patch-wheel.sh
@@ -20,7 +20,7 @@ LIBUCT=$(unzip -l $WHL | awk 'match($4, /libuct-[^\.]+\./) { print substr($4, RS
 LIBUCS=$(unzip -l $WHL | awk 'match($4, /libucs-[^\.]+\./) { print substr($4, RSTART) }')
 LIBNUMA=$(unzip -l $WHL | awk 'match($4, /libnuma-[^\.]+\./) { print substr($4, RSTART) }')
 
-suffix=${RAPIDS_PY_WHEEL_CUDA_SUFFIX/-/_}
+suffix="_$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 mkdir -p ucx_py${suffix}.libs/ucx
 cd ucx_py${suffix}.libs/ucx
 cp -P /usr/lib/ucx/* .
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -45,7 +45,7 @@ jobs:
       sha: ${{ inputs.sha }}
   wheel-build:
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@export-cuda-ver
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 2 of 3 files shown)


.github/build-scripts/patch-wheel.sh
```diff
@@ -20,7 +20,7 @@ LIBUCT=$(unzip -l $WHL | awk 'match($4, /libuct-[^\.]+\./) { print substr($4, RS
 LIBUCS=$(unzip -l $WHL | awk 'match($4, /libucs-[^\.]+\./) { print substr($4, RSTART) }')
 LIBNUMA=$(unzip -l $WHL | awk 'match($4, /libnuma-[^\.]+\./) { print substr($4, RSTART) }')
 
-suffix=${RAPIDS_PY_WHEEL_CUDA_SUFFIX/-/_}
+suffix="_$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
```
pentschev, line 23: UCX-Py is CUDA-agnostic, I wonder if this is the reason this variable was not set in the first place? Maybe we should not have CUDA being part of the UCX-Py version, as is done for [conda packages](https://anaconda.org/rapidsai-nightly/ucx-py/files)?

cc @vyasr in case there's some reason why wheels must pin to a specific CUDA version.

ajschmidt8 (approved):
Approving changes so they can be merged pending the outcome of Peter's question.

.github/build-scripts/patch-wheel.sh
vyasr: We can definitely consider that, but I'd say that's out of scope of this PR. I don't think it is technically feasible at the moment but could become so eventually if we change our linking strategies.

There are a few considerations why CUDA-agnostic RAPIDS packages ended up being CUDA-specific.
1. Dependencies leak their CUDA requirement. For example, `dask_cudf` is CUDA-agnostic (it's pure Python), but if I do `pip install dask_cudf` do I install `cudf-cu11` or `cudf-cu12`? There are some ways around that as we've discovered since the initial implementation, but they're not very clean or robust. So that's one technical reason why even going forward packages might end up CUDA-specific when the underlying code is CUDA-agnostic.
2. Python packages that are CUDA-agnostic may still bundle libraries that are CUDA-specific. ucx-py bundles libucx, and one of those libraries is the CUDA transport layer libuct_cuda.so. That library contains a link to the libcudart specific to that version of the library. IOW that introduces a CUDA version-specific linkage.
2. During initial testing and implementation there was concern about name collisions between pip and conda packages because of the way our workflows were being set up. IIRC Sevag liked the idea of having the suffix be present for all packages as an easy way to delineate the two.

pentschev: Thanks for the details Vyas, given the explanation above I understand why that's important at the moment, no objections from me to keep this as is for now, I was raising the question because I wasn't aware of all that.

> Dependencies leak their CUDA requirement. For example, dask_cudf is CUDA-agnostic (it's pure Python), but if I do pip install dask_cudf do I install cudf-cu11 or cudf-cu12?

Just to be clear here, we're talking exclusively about wheels, not conda packages, right? Definitely not something we have to address now, but to me this is not something `dask-cudf` or `ucx-py` have to answer, similar to what happens in conda packages, the `cudf`/`ucx` dependency is who should answer what CUDA version should be installed. I understand this currently doesn't work this way, just thinking of what a "perfect world" would look like here.

vyasr: Yeah we are only talking about wheels. And the "perfect world" is one where pip understands CUDA.

pentschev (approved):
LGTM, thanks @divyegala and @vyasr !

PR #967, opened 2023-06-23
Add CUDA 12.0 build matrix

### LINKED ISSUE #927: UCX-Py: CUDA 12 Conda Packages
Pushing for CUDA 12 Conda package builds of libraries for RAPIDS 23.04. This issue is for tracking that work.

Requirements (please recheck):

```[tasklist]
### Developer tasks
- [x] I've confirmed the list below is correct (or updated if not)
```

```[tasklist]
### Required dependencies from Conda
- [ ] https://github.com/conda-forge/pynvml-feedstock/pull/11
- [ ] https://github.com/conda-forge/ucx-split-feedstock/issues/116
```

```[tasklist]
### Optional dependencies from Conda
- [ ] https://github.com/conda-forge/cupy-feedstock/issues/139
- [ ] https://github.com/conda-forge/numba-feedstock/issues/112
- [x] `nvtx`
```

```[tasklist]
### RAPIDS Conda dependencies
- [ ] https://github.com/rapidsai/rmm/issues/1207
```

pentschev:
Closes #927

### DIFF
--- a/conda/recipes/ucx-py/meta.yaml
+++ b/conda/recipes/ucx-py/meta.yaml
@@ -38,8 +38,6 @@ requirements:
     {% endfor %}
 
 test:
-  requires:
-    - cuda-version ={{ cuda_version }}
   imports:
     - ucp
 
--- a/conda/recipes/ucx-py/meta.yaml
+++ b/conda/recipes/ucx-py/meta.yaml
@@ -39,7 +39,7 @@ requirements:
 
 test:
   requires:
-    - cuda-version {{ cuda_version }}.*
+    - cuda-version ={{ cuda_version }}
   imports:
     - ucp
 
--- a/conda/recipes/ucx-py/meta.yaml
+++ b/conda/recipes/ucx-py/meta.yaml
@@ -39,7 +39,7 @@ requirements:
 
 test:
   requires:
-    - cudatoolkit {{ cuda_version }}.*
+    - cuda-version {{ cuda_version }}.*
   imports:
     - ucp
 
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -71,6 +71,10 @@ dependencies:
               cuda: "11.8"
             packages:
               - cudatoolkit=11.8
+          - matrix:
+              cuda: "12.0"
+            packages:
+              - cuda-version=12.0
   py_version:
     specific:
       - output_types: conda
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -79,6 +79,7 @@ dependencies:
               cuda: "12.0"
             packages:
               - cuda-version=12.0
+              - cuda-cudart
   py_version:
     specific:
       - output_types: conda
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -58,24 +58,27 @@ dependencies:
           - matrix:
               cuda: "11.2"
             packages:
-              - cudatoolkit=11.2
+              - cuda-version=11.2
+              - cudatoolkit
           - matrix:
               cuda: "11.4"
             packages:
-              - cudatoolkit=11.4
+              - cuda-version=11.4
+              - cudatoolkit
           - matrix:
               cuda: "11.5"
             packages:
-              - cudatoolkit=11.5
+              - cuda-version=11.5
+              - cudatoolkit
           - matrix:
               cuda: "11.8"
             packages:
-              - cudatoolkit=11.8
+              - cuda-version=11.8
+              - cudatoolkit
           - matrix:
               cuda: "12.0"
             packages:
               - cuda-version=12.0
-              - cuda-cudart
   py_version:
     specific:
       - output_types: conda
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -75,6 +75,7 @@ dependencies:
               cuda: "12.0"
             packages:
               - cuda-version=12.0
+              - cuda-cudart
   py_version:
     specific:
       - output_types: conda
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   conda-python-build:
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@cuda-120
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -37,7 +37,7 @@ jobs:
   upload-conda:
     needs: [conda-python-build]
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/conda-upload-packages.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-upload-packages.yaml@cuda-120
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -45,7 +45,7 @@ jobs:
       sha: ${{ inputs.sha }}
   wheel-build:
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@cuda-120
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -58,7 +58,7 @@ jobs:
   wheel-publish:
     needs: wheel-build
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-publish.yml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-publish.yml@cuda-120
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -18,26 +18,26 @@ jobs:
       - wheel-build
       - wheel-tests
     secrets: inherit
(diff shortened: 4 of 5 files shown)


pentschev:
We're seeing the following error for CUDA 12:

```
  File "/opt/conda/lib/python3.9/site-packages/boa/cli/mambabuild.py", line 141, in mamba_get_install_actions
    solution = solver.solve_for_action(_specs, prefix)
  File "/opt/conda/lib/python3.9/site-packages/boa/core/solver.py", line 244, in solve_for_action
    t = self.solve(specs)
  File "/opt/conda/lib/python3.9/site-packages/boa/core/solver.py", line 234, in solve
    raise RuntimeError("Solver could not find solution." + error_string)
RuntimeError: Solver could not find solution.Mamba failed to solve:
 - cudatoolkit 12.0.*
 - libgcc-ng >=12
 - numpy >=1.21
 - pynvml >=11.4.1
 - python >=3.9,<3.10.0a0
 - python_abi 3.9.* *_cp39
 - libstdcxx-ng >=12
 - ucx >=1.14.1,<1.15.0a0

with channels:

The reported errors are:
- Encountered problems while solving:
-   - nothing provides requested cudatoolkit 12.0.*
```

I'm not actually sure what's missing or done wrong here, I've compared to cuDF's `dependencies.yaml`, and the `cudatoolkit` section looks correct. @jakirkham @bdice @vyasr would any of you be able to take a quick look and see if there's something clearly wrong with my changes here?

bdice:
If this is an error in a conda build, check on meta.yaml. This line probably needs to be conditional on CUDA 11. https://github.com/rapidsai/ucx-py/blob/869c3f6cb7797d00b3862d4790851c33daad25d2/conda/recipes/ucx-py/meta.yaml#L42

pentschev:
> If this is an error in a conda build, check on meta.yaml. This line probably needs to be conditional on CUDA 11.
> 
> https://github.com/rapidsai/ucx-py/blob/869c3f6cb7797d00b3862d4790851c33daad25d2/conda/recipes/ucx-py/meta.yaml#L42

@bdice I looked at that, but [cuDF seems to do it the same way](https://github.com/rapidsai/cudf/blob/0b4e3543fb35a6f8bbae3ad3e07b544c534901ac/conda/recipes/cudf/meta.yaml#L103-L105), no conditionals, it does have some conditionals but those are for the `requirements` subsections, which don't really apply to UCX-Py. Do we need to introduce some conditionals in UCX-Py, even though it doesn't really depend on them directly?

jakirkham:
Could you please try dropping `cudatoolkit`?

It's already pulled in by UCX when it is needed

bdice:
@pentschev That’s a bug in cudf’s recipe but we disable recipe tests so it isn’t seen. https://github.com/rapidsai/cudf/blob/0b4e3543fb35a6f8bbae3ad3e07b544c534901ac/ci/build_python.sh#L19

This package isn’t skipping the tests so you do see a failure here.

bdice:
> Could you please try dropping `cudatoolkit`?
> 
> It's already pulled in by UCX when it is needed

Be sure that all the CUDA 11 test configurations pull in their target version. That’s why this specific test pinning exists — otherwise all the CI jobs for 11.2, 11.5, 11.8 would install any compatible version like `>=11,<12` and not test all the intended toolkit versions.

jakirkham:
We could use `cuda-version` instead

pentschev:
> Could you please try dropping `cudatoolkit`?
> 
> It's already pulled in by UCX when it is needed

No, `cudatoolkit` isn't pulled by UCX. Remember in https://github.com/conda-forge/ucx-split-feedstock/pull/111 we've switch to a hybrid build mode that can work for CPU-only and CUDA as well if it's available, therefore by default no CUDA packages will be pulled.

> We could use `cuda-version` instead

Also not sure what this comment relates to, are we supposed to add it somewhere else besides https://github.com/rapidsai/ucx-py/pull/967/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6R77 ?

bdice:
Try replacing this line in meta.yaml: https://github.com/rapidsai/ucx-py/blob/869c3f6cb7797d00b3862d4790851c33daad25d2/conda/recipes/ucx-py/meta.yaml#L42

with
```yaml
 - cuda-version {{ cuda_version }}.*
```

pentschev:
Thanks @bdice and @jakirkham , with the latest suggestion tests are passing now. Do you think we're good here and should move ahead on merging this, or are there further changes required?

dependencies.yaml
```diff
@@ -71,6 +71,11 @@ dependencies:
               cuda: "11.8"
             packages:
               - cudatoolkit=11.8
```
raydouglass, line 73: I think this should be updated to use `cuda-version` as the constraint. Similar to https://github.com/rapidsai/rmm/pull/1296/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6

Not a blocker though

raydouglass (approved):

conda/recipes/ucx-py/meta.yaml
```diff
@@ -39,7 +39,7 @@ requirements:
 
 test:
   requires:
-    - cudatoolkit {{ cuda_version }}.*
+    - cuda-version {{ cuda_version }}.*
```
bdice, line 42:
```suggestion
    - cuda-version ={{ cuda_version }}
```

dependencies.yaml
bdice: Yes, please adopt changes like those in the linked RMM PR. See also this cuDF example: https://github.com/rapidsai/cudf/pull/13615

Basically there should never be any pinnings on `cudatoolkit`. Its version should be controlled by pinnings on `cuda-version`.

bdice (changes requested):

dependencies.yaml
pentschev: Done in https://github.com/rapidsai/ucx-py/pull/967/commits/0e5b0c2a88f31873c1aabcb0ec268d7ab6cff28a .

conda/recipes/ucx-py/meta.yaml
pentschev: Done in https://github.com/rapidsai/ucx-py/pull/967/commits/c04f3de0c44f6140f8cd522ea7368f6283ff908c .

dependencies.yaml
```diff
@@ -58,19 +58,27 @@ dependencies:
...
           - matrix:
               cuda: "11.8"
             packages:
-              - cudatoolkit=11.8
+              - cuda-version=11.8
+              - cudatoolkit
+          - matrix:
+              cuda: "12.0"
+            packages:
+              - cuda-version=12.0
```
jakirkham, line 81: Think we may want these here analogous to [what `ucx` does for CUDA 12]( https://github.com/conda-forge/ucx-split-feedstock/blob/78e8d08b1391a6a29d05eb22bf8a30d1878954f2/recipe/meta.yaml#L37-L39 )

```suggestion
          - matrix:
              cuda: "12.0"
            packages:
              - cuda-version=12.0
              - cuda-cudart-dev
              - cuda-nvml-dev
```

conda/recipes/ucx-py/meta.yaml
```diff
@@ -39,7 +39,7 @@ requirements:
 
 test:
   requires:
-    - cudatoolkit {{ cuda_version }}.*
+    - cuda-version ={{ cuda_version }}
```
jakirkham, line 42: Thinking about this more, maybe we can drop the test requirements altogether?

Since all we are testing here is that `import ucp` works and that doesn't require CUDA transport support from `ucx`, maybe no additional requirements are needed for tests

Any tests needing the CUDA transport happen independently of the package build (so already install the needed dependencies)

Thoughts?

```suggestion
```

jakirkham (commented):
Thanks Peter! 🙏

Had a couple comments below

conda/recipes/ucx-py/meta.yaml
pentschev: Yes, removing those make sense. UCX-Py does not have CUDA as a hard-dependency, using it for host transfers only is totally valid. Done in https://github.com/rapidsai/ucx-py/pull/967/commits/dcac26dbf52f3888b9918a186f12af23fa18bf8e .

dependencies.yaml
pentschev: Why? UCX-Py doesn't depend on any CUDA runtime dev packages to build. I think though that the intent here is to ensure they're available at runtime to properly test CUDA, is that right? If that is really the case then we may want those same packages but without `-dev`, WDYT?

jarmak-nv:
Hi @pentschev could you please link this PR to https://github.com/rapidsai/ucx-py/issues/927 either by editing in closing keywords to the description, or through the UI?

Thanks!

dependencies.yaml
bdice: This part of `dependencies.yaml` is used to generate package lists for development environments. While `ucx-py` doesn't have a developer conda environment in https://github.com/rapidsai/ucx-py/tree/branch-0.33/conda/environments, it might be nice to list any packages needed for building/testing here. I'm unfamiliar with how nvml is used in this package so I can't comment on the correctness of this suggestion.

bdice (approved):
Seems fine, one minor comment on the open conversation about `dependencies.yaml` but nothing blocking.

dependencies.yaml
jakirkham: Yeah that is fair. Was thinking this might be useful when building UCX-Py (and maybe also UCX) in an environment. That said, no strong feelings about this particular construction

If we want a strictly runtime focused solution, would suggest something like this

```suggestion
          - matrix:
              cuda: "12.0"
            packages:
              - cuda-version=12.0
              - cuda-cudart
```

jakirkham: Peter added the latter below in commit ( https://github.com/rapidsai/ucx-py/pull/967/commits/58d5ac29f3658a69f59a5784c839a540f91a248a )

Marking this resolved

jakirkham (approved):

pentschev:
Woohoo! Thanks all for the reviews here! 😄

jakirkham:
Thanks Peter and thanks everyone for reviewing! 🙏

PR #971, opened 2023-07-24
Resolve forward-merger conflicts

pentschev:
Fixes https://github.com/rapidsai/ucx-py/pull/970


raydouglass (approved):

vyasr (commented):
Noting after the fact that the changes here all look correct to me.

PR #977, opened 2023-07-26
Remove script input from publish step in build.yaml.

bdice:
Nightly tests failed because an extra `script` variable was passed to the wheel publish workflow.

Failures look like:
```
Invalid workflow file: .github/workflows/build.yaml#L65
The workflow is not valid. .github/workflows/build.yaml (Line: 65, Col: 15): Invalid input, script is not defined in the referenced workflow.
```

https://github.com/rapidsai/ucx-py/actions/runs/5664871243

Follow-up to #973.


pentschev (commented):
LGTM, thanks @bdice !

raydouglass (approved):

Issue #982, opened 2023-08-23
PyNVML returns not supported error on 550 driver


rjzamora:
Users are seeing the following error with newer (550) drivers:

```
 File "/home/lab/miniconda3/envs/rapids-23.06-3.9/lib/python3.9/site-packages/pynvml/nvml.py", line 765, in _nvmlCheckReturn
    raise NVMLError(ret)
pynvml.nvml.NVMLError_NotSupported: Not Supported

...

 File "/home/lab/miniconda3/envs/rapids-23.06-3.9/lib/python3.9/site-packages/ucp/__init__.py", line 68, in <module>
    bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(handle).bar1Total
  File "/home/lab/miniconda3/envs/rapids-23.06-3.9/lib/python3.9/site-packages/pynvml/nvml.py", line 2070, in nvmlDeviceGetBAR1MemoryInfo
    _nvmlCheckReturn(ret)
```

It has been confirmed offline that pynvml is correctly returning `NVMLError_NotSupported` for `nvmlDeviceGetBAR1MemoryInfo`, because this device does **not** support Bar1 accesss.

The `ucp/__init__.py` logic should be updated to account for a `NVMLError_NotSupported` exception.

PR #983, opened 2023-08-23
Handle ``NVMLError_NotSupported`` exception

rjzamora:
Closes https://github.com/rapidsai/ucx-py/issues/982

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -67,8 +67,9 @@ if (
             try:
                 bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(handle).bar1Total
             except pynvml.nvml.NVMLError_NotSupported:
-                # Bar1 access not supported on this device. Skip
-                continue
+                # Bar1 access not supported on this device, set it to
+                # zero (always lower than device memory).
+                bar1_total = 0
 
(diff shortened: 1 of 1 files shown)


ucp/__init__.py
```diff
@@ -64,9 +64,13 @@ def _is_mig_device(handle):
             if _is_mig_device(handle):
                 continue
 
-            total_memory = pynvml.nvmlDeviceGetMemoryInfo(handle).total
-            bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(handle).bar1Total
+            try:
+                bar1_total = pynvml.nvmlDeviceGetBAR1MemoryInfo(handle).bar1Total
+            except pynvml.nvml.NVMLError_NotSupported:
+                # Bar1 access not supported on this device. Skip
+                continue
```
rjzamora, line 71: @pentschev - Is it "correct" to leave the `large_bar1` element as `False` for devices without Bar1 access?

pentschev, line 71:
```suggestion
                # Bar1 access not supported on this device, set it to
                # zero (always lower than device memory).
                bar1_total = 0
```

pentschev (changes requested):

pentschev (approved):
Thanks @rjzamora for working on this PR! 😄

PR #986, opened 2023-09-06
Simplify wheel build scripts and allow alphas of RAPIDS dependencies

divyegala:
This PR:

1. Removes `ci/apply_wheel_modifications.sh` and uses it inline in wheel build scripts
2. Allows for specifying alpha versioned dependencies of RAPIDS projects


ci/build_wheel.sh
```diff
@@ -3,6 +3,9 @@
 
 set -euo pipefail
 
+package_name="ucx-py"
+underscore_package_name=$(echo "${package_name}" | tr "-" "_")
```
vyasr, line 7: Good idea to save this here.

vyasr (approved):

ajschmidt8 (approved):

PR #987, opened 2023-09-07
Undo manual pinning

vyasr:
This PR is a test to validate that upstream ucx builds are no longer broken, at which point we will also be able to remove this pinning.

### DIFF
--- a/conda/recipes/ucx-py/meta.yaml
+++ b/conda/recipes/ucx-py/meta.yaml
@@ -5,7 +5,6 @@
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') %}
 {% set py_version = environ['CONDA_PY'] %}
 {% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
-{% set cuda_major = cuda_version.split('.')[0] %}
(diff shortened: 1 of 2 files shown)


quasiben (approved):

raydouglass (approved):

conda/recipes/ucx-py/meta.yaml
```diff
@@ -43,14 +36,7 @@ requirements:
     {% for r in data.get("project", {}).get("dependencies", []) %}
     - {{ r }}
     {% endfor %}
-    # Something seems wrong with the run export as well now
-    {% if cuda_major == "11" %}
     - ucx
```
pentschev, line 39: Just wanted to check that this `ucx` is required in the `run` section. I noticed it was added by https://github.com/rapidsai/ucx-py/pull/986/files#diff-516b294659252e02809cd6547b97b86e8172251ad27a8cd5a725261e36e18d7c, but it was not removed here when reverting those changes. Is that intentional @vyasr ?

vyasr: Ah no good catch, that is not intentional. It was there because there were issues with the run exports on the C++ package. I'll open a PR to remove it now and we can verify that everything works correctly.

vyasr: Opened https://github.com/rapidsai/ucx-py/pull/990

PR #990, opened 2023-09-20
Remove ucx from explicit run reqs

vyasr:
The ucx run requirement should be inferred from the run exports of ucx since it is part of the host requirements. The run explicit requirement was added to fix upstream issues with ucx packaging that have since been fixed. All other patches were removed in https://github.com/rapidsai/ucx-py/pull/987, but this reversion was missed.


vyasr:
I'm not an admin on RTD so I can't rerun that build, but it's just a timeout so a rerun should fix.

pentschev (approved):
LGTM, thanks @vyasr . I triggered an RTD rebuild and everything's green now.

raydouglass (approved):

PR #996, opened 2023-10-12
Update UCX pin to allow `1.15.x` and update docs

pentschev:
Update UCX pin to allow `1.15.x`, remove `ucx-proc` dependency and update docs and warnings accordingly.

Since UCX 1.14.1, the `ucx-proc` package is not required anymore and all UCX builds provide CUDA support.


raydouglass (approved):

wence- (approved):
Doc changes look good, thanks Peter

pentschev:
Thanks @raydouglass and @wence- .

PR #1001, opened 2023-10-18
Test increased nofile ulimit

pentschev:


pentschev:
This seems to have worked, once https://github.com/rapidsai/shared-workflows/pull/145 is merged, the `ulimit` change should be in effect without any further changes required, thus we can close this PR.

PR #1003, opened 2023-10-24
Ignore UCX run exports and allow `ucx>=1.14.1,<1.16.0`

pentschev:


vyasr (approved):
Builds are now [correctly using 1.14](https://github.com/rapidsai/ucx-py/actions/runs/6629461189/job/18008877833?pr=1003#step:7:456), and the explicit run requirement should be relaxed relative to before. I'm going to merge this so we can get some new builds out and see if that fixes the solver issues down the line.

raydouglass (approved):

bdice (commented):
Sorry. I just found this browser tab with an unsubmitted review. I have a comment that was not submitted earlier. It's not essential to address but might be worthwhile to know.

PR #1019, opened 2024-01-29
Pin to `pytest==7.*`

pentschev:
Pytest 8 introduced breaking changes that `pytest-asyncio` is still in the process of resolving. See
https://github.com/pytest-dev/pytest-asyncio/issues/763. This pin should also be done directly by `pytest-asyncio>=0.23.4`, but that version is still not available in conda-forge.


wence- (approved):

pentschev:
Thanks @wence- !

PR #1020, opened 2024-01-30
Pin to `pytest==7.*`

pentschev:
Pytest 8 introduced breaking changes that `pytest-asyncio` is still in the process of resolving. See
https://github.com/pytest-dev/pytest-asyncio/issues/763. This pin should also be done directly by `pytest-asyncio>=0.23.4`, but that version is still not available in conda-forge.

Backport of https://github.com/rapidsai/ucx-py/pull/1019 .


charlesbluca (approved):
Thanks @pentschev! Looks like the issue you linked is closed but progress on the hotfix is being tracked in https://github.com/pytest-dev/pytest-asyncio/issues/737.

Also worth noting that it looks like the hotfix is actually released now, and we're just waiting on the conda-forge release PR to get merged:

https://github.com/conda-forge/pytest-asyncio-feedstock/pull/40

pentschev:
Yup, I saw the conda-forge PR but while it isn't merged we can't count on it, so best to just go ahead and pin it ourselves in the meantime. Thanks @charlesbluca !

raydouglass (approved):

PR #1025, opened 2024-02-07
Add environment-agnostic scripts for running pytests

trxcllnt:
This PR adds an environment-agnostic `run_pytests.sh` script, and updates `test_python.sh` to call it.

The `test_python.sh` script assumes it's running in our CI environment, and it does more than just run pytest.

This PR allows devs and downstream consumers to only run the tests, and skip the unrelated logic in `test_python.sh`.

### DIFF
--- a/ci/run_benchmark_pytests.sh
+++ b/ci/run_benchmark_pytests.sh
@@ -6,8 +6,6 @@ set -euo pipefail
 # Support invoking run_pytests.sh outside the script directory
 cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 
-export UCX_TLS="${UCX_TLS:-tcp,cuda_copy}"
-
 # cd to root directory to prevent repo's `ucp` directory from being used
 # in subsequent commands
 pushd /
--- a/ci/run_benchmark_pytests.sh
+++ b/ci/run_benchmark_pytests.sh
@@ -3,9 +3,6 @@
 
 set -euo pipefail
 
-# Support invoking run_pytests.sh outside the script directory
-cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
-
 # cd to root directory to prevent repo's `ucp` directory from being used
 # in subsequent commands
 pushd /
--- a/ci/run_benchmark_pytests.sh
+++ b/ci/run_benchmark_pytests.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Copyright (c) 2024, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Support invoking run_pytests.sh outside the script directory
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
+
+export UCX_TLS="${UCX_TLS:-tcp,cuda_copy}"
+
+# cd to root directory to prevent repo's `ucp` directory from being used
+# in subsequent commands
+pushd /
+timeout 1m python -m ucp.benchmarks.send_recv -o cupy --server-dev 0 --client-dev 0 --reuse-alloc --backend ucp-async
+timeout 1m python -m ucp.benchmarks.send_recv -o cupy --server-dev 0 --client-dev 0 --reuse-alloc --backend ucp-core
+timeout 1m python -m ucp.benchmarks.cudf_merge --chunks-per-dev 4 --chunk-size 10000 --rmm-init-pool-size 2097152
+popd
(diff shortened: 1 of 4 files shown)


ci/run_pytests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Copyright (c) 2024, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Support invoking run_pytests.sh outside the script directory
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
+
+export UCX_TLS="${UCX_TLS:-tcp,cuda_copy}"
```
pentschev, line 9: What are the environments where this will normally run? We originally set this value because of CI limitations on interconnects (e.g., no InfiniBand/NVLink). If this is going to run on development test environments where this may be available we might want not to set anything. In fact, we could remove this entirely today, `UCX_TLS=all` (default) should work well anywhere on recent UCX versions.

ci/test_python.sh
```diff
@@ -36,8 +40,7 @@ run_tests() {
 
   # Test with TCP/Sockets
   rapids-logger "TEST WITH TCP ONLY"
-  timeout 10m pytest --cache-clear -vs tests/
-  timeout 2m pytest --cache-clear -vs ucp/_libs/tests
+  ./ci/run_pytests.sh
 
   rapids-logger "Run local benchmark"
```
pentschev, line 45: Although they're technically "benchmarks" and not regular `pytest`s, I'd argue they're part of the test-suite and thus I would also move these into `run_pytests.sh`.

trxcllnt: That's fine with me. I wasn't sure if they were tests that should fail a build, or just verifying the benchmark code still runs. I'm happy to move them into the test script.

ci/run_pytests.sh
trxcllnt: Yeah I wasn't sure if this should be set, so I'm being conservative. A user can override it, but I didn't want the tests to fail if they don't.

My main goal of this PR (and all the other similar ones I've filed tonight in other repos) is to provide a test script we can run in the DLFW/NVAIE builds. The test recipes change so frequently, that it's difficult to keep the DLFW tests in sync for all repos.

I typically run these tests on my local workstation, or on a lab DGX if I need another machine. Eventually we want to run the tests in Blossom or Colossus, but I can't predict what will be available in those environments.

ci/test_python.sh
trxcllnt: I made a second `run_benchmark_pytests.sh` for these. That way if they cause issues (unrelated to correctness) in DLFW, I can choose not to run them.

ci/run_pytests.sh
pentschev: If you run on DGX lab machines I'd suggest we remove `UCX_TLS` completely and use UCX's default from now on, this applies to regular CI as well. Would you mind removing `UCX_TLS` both here and in `ci/test_python.sh`?

ci/test_python.sh
pentschev: Those benchmarks should never fail either, they're distributed as part of the package because they're used by users to test interconnects, and thus they're part of the test suite here as we don't want them to be broken ever, so if they ever we would want that fixed just as any other test. We can keep `run_benchmark_pytests.sh` separate from `run_pytests.sh` if you wish but I would have combined them, when they're not combined an uninformed developer may think they're purely benchmarks in the sense of providing performance metrics and not part of the distributed package which should always be tested together.

ci/run_benchmark_pytests.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Copyright (c) 2024, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Support invoking run_pytests.sh outside the script directory
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
```
msarahan, line 7: For the sake of common patterns across each of these, I see the reason why this is here. However, it is a little unintuitive to me how we immediately pushd away from this folder, and rely on PATH/PYTHONPATH for choosing code to run. Is it worth removing this first cd here, or is that just nitpicking not worth the time?

trxcllnt: The pushd isn't strictly necessary in environments that support `PYTHONSAFEPATH=1`, so I figured it'd eventually get removed.

pentschev (approved):

AyodeAwe (approved):

pentschev:
Thanks @trxcllnt !

PR #1041, opened 2024-05-06
use libucx wheels

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/57.

Similar to https://github.com/rapidsai/ucxx/pull/226, proposes using the new UCX wheels from https://github.com/rapidsai/ucx-wheels, instead of vendoring system versions of `libuc{m,p,s,t}.so`.

## Benefits of these changes

Allows users of `ucx-py` to avoid needing system installations of the UCX libraries.

Shrinks the `ucx-py` wheels by 6.7MB compressed (77%) and 19.1 MB uncompressed (73%).

<details><summary>how I calculated that (click me)</summary>

Mounting in a directory with a wheel built from this branch...

```shell
docker run \
    --rm \
    -v $(pwd)/final_dist:/opt/work \
    -it python:3.10 \
    bash

pip install pydistcheck
pydistcheck --inspect /opt/work/*.whl
```

```text
----- package inspection summary -----
file size
  * compressed size: 2.0M
  * uncompressed size: 7.0M
  * compression space saving: 71.3%
contents
  * directories: 10
  * files: 38 (2 compiled)
size by extension
  * .so - 6.9M (97.7%)
  * .py - 0.1M (2.0%)
  * .pyx - 9.3K (0.1%)
  * no-extension - 7.1K (0.1%)
  * .pyi - 3.9K (0.1%)
  * .c - 1.7K (0.0%)
  * .txt - 39.0B (0.0%)
largest files
  * (5.3M) ucp/_libs/ucx_api.cpython-310-x86_64-linux-gnu.so
  * (1.6M) ucp/_libs/arr.cpython-310-x86_64-linux-gnu.so
  * (36.3K) ucp/core.py
  * (20.3K) ucp/benchmarks/cudf_merge.py
  * (12.1K) ucp/benchmarks/send_recv.py
```

Compared to a recent nightly release.

```shell
pip download \
    -d /tmp/delete-me \
    --prefer-binary \
    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
    'ucx-py-cu12>=0.38.0a'

pydistcheck --inspect /tmp/delete-me/*.whl
```

```text
----- package inspection summary -----
file size
  * compressed size: 8.7M
  * uncompressed size: 26.1M
  * compression space saving: 66.8%
contents
  * directories: 11
  * files: 65 (21 compiled)
size by extension
  * .0 - 14.4M (55.4%)
  * .so - 8.4M (32.2%)
  * .a - 1.8M (6.7%)
  * .140 - 0.7M (2.5%)
  * .12 - 0.7M (2.5%)
  * .py - 0.1M (0.5%)
  * .pyx - 9.3K (0.0%)
  * no-extension - 7.3K (0.0%)
  * .la - 4.2K (0.0%)
  * .pyi - 3.9K (0.0%)
  * .c - 1.7K (0.0%)
  * .txt - 39.0B (0.0%)
largest files
  * (8.7M) ucx_py_cu12.libs/libucp-5720f0c9.so.0.0.0
  * (5.3M) ucp/_libs/ucx_api.cpython-310-x86_64-linux-gnu.so
  * (2.0M) ucx_py_cu12.libs/libucs-3c3009f0.so.0.0.0
  * (1.6M) ucp/_libs/arr.cpython-310-x86_64-linux-gnu.so
  * (1.5M) ucx_py_cu12.libs/libuct-2a15b69b.so.0.0.0
```

</details>

## Notes for Reviewers

Left some comments on the diff describing specific design choices.

### The libraries from the `libucx` wheel are only used if a system installation isn't available

Built a wheel in a container using the same image used here in CI.

```shell
docker run \
    --rm \
    --gpus 1 \
    --env-file "${HOME}/.aws/creds.env" \
    --env CI=true \
    -v $(pwd):/opt/work \
    -w /opt/work \
    -it rapidsai/ci-wheel:cuda12.2.2-rockylinux8-py3.10 \
    bash

ci/build_wheel.sh
```

</details>

Found that the libraries from the `libucx` wheel are correctly found at build time, and are later found at import time.

<details><summary>using 'rapidsai/citestwheel' image and LD_DEBUG (click me)</summary>

```shell
# run a RAPIDS wheel-testing container, mount in the directory with the built wheel
docker run \
    --rm \
    --gpus 1 \
    -v $(pwd)/final_dist:/opt/work \
    -w /opt/work \
    -it rapidsai/citestwheel:cuda12.2.2-ubuntu22.04-py3.10 \
    bash
```

`rapidsai/citestwheel` does NOT the UCX libraries installed at `/usr/lib*`.

```shell
find /usr -name 'libucm.so*'
# (empty)
```

Installed the `ucx-py` wheel.

```shell
# install the wheel
pip install ./*.whl

# now libuc{m,p,s,t} at found in site-packages
find /usr -name 'libucm.so*'
# (empty)

find /pyenv -name 'libucm.so*'
# /pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib/libucm.so.0.0.0
# /pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib/libucm.so.0
# /pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib/libucm.so

# try importing ucx-py and track where 'ld' finds the ucx libraries
LD_DEBUG="files,libs" LD_DEBUG_OUTPUT=out.txt \
python -c "from ucp._libs import arr"

# 'ld' creates multiple files... combine them to 1 for easier searching
cat out.txt.* > out-full.txt
```

In that output, saw that `ld` was finding `libucs.so` first.

It searched all the system paths before finally finding it in the `libucx` wheel.

```text
1037:	file=libucs.so [0];  dynamically loaded by /pyenv/versions/3.10.14/lib/python3.10/lib-dynload/_ctypes.cpython-310-x86_64-linux-gnu.so [0]
      1037:	find library=libucs.so [0]; searching
      1037:	 search path=		(LD_LIBRARY_PATH)
      1037:	 search path=/pyenv/versions/3.10.14/lib		(RUNPATH from file /pyenv/versions/3.10.14/bin/python)
      1037:	  trying file=/pyenv/versions/3.10.14/lib/libucs.so
      1037:	 search cache=/etc/ld.so.cache
      1037:	 search path=/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v3:/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v2:/lib/x86_64-linux-gnu/tls/haswell/x86_64:/lib/x86_64-linux-gnu/tls/haswell:/lib/x86_64-linux-gnu/tls/x86_64:/lib/x86_64-linux-gnu/tls:/lib/x86_64-linux-gnu/haswell/x86_64:/lib/x86_64-linux-gnu/haswell:/lib/x86_64-linux-gnu/x86_64:/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v3:/usr/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v2:/usr/lib/x86_64-linux-gnu/tls/haswell/x86_64:/usr/lib/x86_64-linux-gnu/tls/haswell:/usr/lib/x86_64-linux-gnu/tls/x86_64:/usr/lib/x86_64-linux-gnu/tls:/usr/lib/x86_64-linux-gnu/haswell/x86_64:/usr/lib/x86_64-linux-gnu/haswell:/usr/lib/x86_64-linux-gnu/x86_64:/usr/lib/x86_64-linux-gnu:/lib/glibc-hwcaps/x86-64-v3:/lib/glibc-hwcaps/x86-64-v2:/lib/tls/haswell/x86_64:/lib/tls/haswell:/lib/tls/x86_64:/lib/tls:/lib/haswell/x86_64:/lib/haswell:/lib/x86_64:/lib:/usr/lib/glibc-hwcaps/x86-64-v3:/usr/lib/glibc-hwcaps/x86-64-v2:/usr/lib/tls/haswell/x86_64:/usr/lib/tls/haswell:/usr/lib/tls/x86_64:/usr/lib/tls:/usr/lib/haswell/x86_64:/usr/lib/haswell:/usr/lib/x86_64:/usr/lib		(system search path)
      1037:	  trying file=/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v3/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v2/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/tls/haswell/x86_64/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/tls/haswell/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/tls/x86_64/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/tls/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/haswell/x86_64/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/haswell/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/x86_64/libucs.so
      1037:	  trying file=/lib/x86_64-linux-gnu/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v3/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v2/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/tls/haswell/x86_64/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/tls/haswell/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/tls/x86_64/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/tls/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/haswell/x86_64/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/haswell/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/x86_64/libucs.so
      1037:	  trying file=/usr/lib/x86_64-linux-gnu/libucs.so
      1037:	  trying file=/lib/glibc-hwcaps/x86-64-v3/libucs.so
      1037:	  trying file=/lib/glibc-hwcaps/x86-64-v2/libucs.so
      1037:	  trying file=/lib/tls/haswell/x86_64/libucs.so
      1037:	  trying file=/lib/tls/haswell/libucs.so
      1037:	  trying file=/lib/tls/x86_64/libucs.so
      1037:	  trying file=/lib/tls/libucs.so
      1037:	  trying file=/lib/haswell/x86_64/libucs.so
      1037:	  trying file=/lib/haswell/libucs.so
      1037:	  trying file=/lib/x86_64/libucs.so
      1037:	  trying file=/lib/libucs.so
      1037:	  trying file=/usr/lib/glibc-hwcaps/x86-64-v3/libucs.so
      1037:	  trying file=/usr/lib/glibc-hwcaps/x86-64-v2/libucs.so
      1037:	  trying file=/usr/lib/tls/haswell/x86_64/libucs.so
      1037:	  trying file=/usr/lib/tls/haswell/libucs.so
      1037:	  trying file=/usr/lib/tls/x86_64/libucs.so
      1037:	  trying file=/usr/lib/tls/libucs.so
      1037:	  trying file=/usr/lib/haswell/x86_64/libucs.so
      1037:	  trying file=/usr/lib/haswell/libucs.so
      1037:	  trying file=/usr/lib/x86_64/libucs.so
      1037:	  trying file=/usr/lib/libucs.so
      1037:	
      1037:	file=/pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib/libucs.so [0];  dynamically loaded by /pyenv/versions/3.10.14/lib/python3.10/lib-dynload/_ctypes.cpython-310-x86_64-linux-gnu.so [0]
      1037:	file=/pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib/libucs.so [0];  generating link map
      1037:	  dynamic: 0x00007f4ce42d7c80  base: 0x00007f4ce427e000   size: 0x000000000006fda0
      1037:	    entry: 0x00007f4ce4290ce0  phdr: 0x00007f4ce427e040  phnum:                 1
```

Then the others were found via the RPATH entries on `libucs.so`.

`libucm.so.0`:

```text
       196:	file=libucm.so.0 [0];  needed by /pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib/libucs.so [0]
       196:	find library=libucm.so.0 [0]; searching
       196:	 search path=...redacted...:/pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib		(RPATH from file /pyenv/versions/3.10.14/lib/python3.10/site-packages/libucx/lib/libucs.so)
      ...
```

</details>

However, the libraries from the `libucx` wheel appear to be **the last place `ld` searches**. That means that if you use these wheels on a system with a system installation of `libuc{m,p,s,t}`, that system installation's libraries will be loaded instead.

<details><summary>using 'rapidsai/ci-wheel' image and LD_DEBUG (click me)</summary>

```shell
docker run \
    --rm \
    --gpus 1 \
    -v $(pwd)/final_dist:/opt/work \
    -w /opt/work \
    -it rapidsai/ci-wheel:cuda12.2.2-rockylinux8-py3.10 \
    bash
```

`rapidsai/ci-wheel` has the UCX libraries installed at `/usr/lib64`.

```shell
find /usr/ -name 'libucm.so*'
# /usr/lib64/libucm.so.0.0.0
# /usr/lib64/libucm.so.0
# /usr/lib64/libucm.so
```

Installed a wheel and tried to import from it.

```shell
pip install ./*.whl

LD_DEBUG="files,libs" LD_DEBUG_OUTPUT=out.txt \
python -c "from ucp._libs import arr"

cat out.txt.* > out-full.txt
```

In that situation, I saw the system libraries found before the one from the wheel.

```text
       226:	file=libucs.so [0];  dynamically loaded by /pyenv/versions/3.10.14/lib/python3.10/lib-dynload/_ctypes.cpython-310-x86_64-linux-gnu.so [0]
       226:	find library=libucs.so [0]; searching
       226:	 search path=/pyenv/versions/3.10.14/lib		(RPATH from file /pyenv/versions/3.10.14/bin/python)
       226:	  trying file=/pyenv/versions/3.10.14/lib/libucs.so
       226:	 search path=/pyenv/versions/3.10.14/lib		(RPATH from file /pyenv/versions/3.10.14/bin/python)
       226:	  trying file=/pyenv/versions/3.10.14/lib/libucs.so
       226:	 search path=/opt/rh/gcc-toolset-11/root/usr/lib64/tls:/opt/rh/gcc-toolset-11/root/usr/lib64:/opt/rh/gcc-toolset-11/root/usr/lib		(LD_LIBRARY_PATH)
       226:	  trying file=/opt/rh/gcc-toolset-11/root/usr/lib64/tls/libucs.so
       226:	  trying file=/opt/rh/gcc-toolset-11/root/usr/lib64/libucs.so
       226:	  trying file=/opt/rh/gcc-toolset-11/root/usr/lib/libucs.so
       226:	 search cache=/etc/ld.so.cache
       226:	  trying file=/usr/lib64/libucs.so
```

In this case, when the system libraries are available, `site-packages/libucx/lib` isn't even searched.

</details>

To avoid any RAPIDS-specific stuff tricking me, I tried in a generic `python:3.10` image. Found that the library could be loaded and all the `libuc{m,p,s,t}` libraries from the `libucx` wheel are found 🎉 .

<details><summary>using 'python:3.10' wheel (click me)</summary>

```shell
docker run \
    --rm \
    --gpus 1 \
    -v $(pwd)/final_dist:/opt/work \
    -w /opt/work \
    -it python:3.10 \
    bash

pip install \
    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
    ./*.whl

LD_DEBUG="files,libs" LD_DEBUG_OUTPUT=out.txt \
python -c "from ucp._libs import arr"
```

💥 

```text
        16:	opening file=/usr/local/lib/python3.10/site-packages/libucx/lib/libucm.so.0 [0]; direct_opencount=1
        16:	
        16:	opening file=/usr/local/lib/python3.10/site-packages/libucx/lib/libucs.so [0]; direct_opencount=1
```

</details>


.readthedocs.yml
```diff
@@ -4,6 +4,9 @@ build:
   os: "ubuntu-22.04"
   tools:
     python: "mambaforge-22.9"
+  jobs:
+    pre_install:
+      - bash ci/build_docs_pre_install.sh
```
jameslamb, line 9: There is no `pip`-installable library `libucx`... it needs to be either `libucx-cu11` or `libucx-cu12`.

Readthedocs builds are installing directly from the `pyproject.toml` checked into source control, which
doesn't have those suffixes added.

https://github.com/rapidsai/ucx-py/blob/03c864b790a276e6732e00b962c6cb4d143eb786/.readthedocs.yml#L8-L11

Resulting in this:

> INFO: pip is looking at multiple versions of ucx-py to determine which version is compatible with other requirements. This could take a while.
> ERROR: Could not find a version that satisfies the requirement libucx<1.16,>=1.15.0 (from ucx-py) (from versions: none)
> ERROR: No matching distribution found for libucx<1.16,>=1.15.0

([example build link](https://readthedocs.org/projects/ucx-py/builds/24296596/))

`readthedocs` doesn't allow customizing the `pip install` with arbitrary flags, e.g. by adding `--no-deps` ([docs](https://docs.readthedocs.io/en/stable/config-file/v2.html#python)).

So I think this pre-install script to fix the suffixes is the best option to keep those builds working. For more details on how it works, see https://docs.readthedocs.io/en/stable/build-customization.html.

conda/recipes/ucx-py/meta.yaml
```diff
@@ -18,7 +18,7 @@ build:
   number: {{ GIT_DESCRIBE_NUMBER }}
   string: py{{ py_version }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
   script:
-    - {{ PYTHON }} -m pip install . -vv
+    - {{ PYTHON }} -m pip install --no-build-isolation --no-deps . -vv
```
jameslamb, line 21: Without this, conda builds tried to install `libucx` (no suffix) and failed complaining about that library not existing.

Using `--no-deps` prevents that. I think this is a better pattern *anyway*... it defers 100% of the responsibility for providing build dependencies to `conda-build`, which means we'll get conda packages of all those things and find out via build failures if some required `build` / `host` dependency was accidentally omitted from this recipe.

pyproject.toml
```diff
@@ -16,6 +16,7 @@
 build-backend = "setuptools.build_meta"
 requires = [
     "cython>=3.0.0",
+    "libucx==1.15.0",
```
jameslamb, line 19: Proposing here to build against a specific version (`==1.15.0`) but run against a range of compatible versions (`>=1.15.0,<1.16`), based on the pattern I saw in https://github.com/rapidsai/ucxx/pull/226.

setup.py
```diff
@@ -12,8 +12,24 @@
...
+#
+# NOTE: doing this '.__file__' stuff because `sysconfig.get_path("platlib")` and similar
+#       can return paths to the main Python interpreter's installation locations...
+#       not where 'libucx' is going to be installed at build time when using
+#       build isolation
+try:
+    import libucx
+
+    libucx_libdir = os.path.join(os.path.dirname(libucx.__file__), "lib")
+    library_dirs.append(libucx_libdir)
```
jameslamb, line 28: Without this, builds in an environment without a system installation of `libuc{m,p,s,t}` fail like this:

```text
[1/1] Cythonizing ucp/_libs/ucx_api.pyx
building 'ucp._libs.ucx_api' extension
creating build/temp.linux-x86_64-cpython-310
creating build/temp.linux-x86_64-cpython-310/ucp
creating build/temp.linux-x86_64-cpython-310/ucp/_libs
creating build/temp.linux-x86_64-cpython-310/ucp/_libs/src
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/pyenv/versions/3.10.14/include -I/pyenv/versions/3.10.14/include/python3.10 -c ucp/_libs/src/c_util.c -o build/temp.linux-x86_64-cpython-310/ucp/_libs/src/c_util.o -std=c99 -Werror
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/pyenv/versions/3.10.14/include -I/pyenv/versions/3.10.14/include/python3.10 -c ucp/_libs/ucx_api.c -o build/temp.linux-x86_64-cpython-310/ucp/_libs/ucx_api.o -std=c99 -Werror
gcc -pthread -shared -L/pyenv/versions/3.10.14/lib -Wl,-rpath,/pyenv/versions/3.10.14/lib -L/usr/lib -L/pyenv/versions/3.10.14/lib -Wl,-rpath,/pyenv/versions/3.10.14/lib -L/usr/lib build/temp.linux-x86_64-cpython-310/ucp/_libs/src/c_util.o build/temp.linux-x86_64-cpython-310/ucp/_libs/ucx_api.o -L/pyenv/versions/3.10.14/lib -L/pyenv/versions/3.10.14/lib -lucp -luct -lucm -lucs -o build/lib.linux-x86_64-cpython-310/ucp/_libs/ucx_api.cpython-310-x86_64-linux-gnu.so
/opt/rh/gcc-toolset-11/root/usr/bin/ld: cannot find -lucp
/opt/rh/gcc-toolset-11/root/usr/bin/ld: cannot find -luct
/opt/rh/gcc-toolset-11/root/usr/bin/ld: cannot find -lucm
/opt/rh/gcc-toolset-11/root/usr/bin/ld: cannot find -lucs
collect2: error: ld returned 1 exit status
error: command '/opt/rh/gcc-toolset-11/root/usr/bin/gcc' failed with exit code 1
```

ucp/__init__.py
```diff
@@ -16,6 +16,18 @@
...
+
+# If libucx was installed as a wheel, we must request it to load the library symbols.
+# Otherwise, we assume that the library was installed in a system path that ld can find.
+try:
+    import libucx
+except ImportError:
+    pass
+else:
+    libucx.load_library()
+    del libucx
```
jameslamb, line 28: Took this directly from here: https://github.com/rapidsai/ucxx/pull/226/files#r1583692142.

pentschev (approved):
Overall this looks fine to me, I'll anyway say @vyasr should also give this a pass as he's gonna definitely be able to comment on any details that are gonna pass unnoticed by me.

jameslamb:
Thanks for reviewing @pentschev !

conda/recipes/ucx-py/meta.yaml
vyasr: My understanding was that `conda-build` automatically configured the environment not to install pip dependencies by setting the `PIP_NO_DEPENDENCIES` environment variable, see https://github.com/conda/conda-build/blob/main/conda_build/build.py#L3008. IOW I thought these two flags would be implied and superfluous. Let's double-check that with some of our conda experts to see why they might be necessary before we commit this.

dependencies.yaml
```diff
@@ -115,9 +117,50 @@ dependencies:
         packages:
           - numpy>=1.23,<2.0a0
           - pynvml>=11.4.1
+  depends_on_ucx_build:
+    common:
+      - output_types: conda
+        packages:
+          - &ucx_conda_build ucx==1.15.0
```
vyasr, line 124: It doesn't look like this reference is ever used anywhere.
```suggestion
          - ucx==1.15.0
```

pyproject.toml
vyasr: Yup that is exactly right, it's what we want.

setup.py
vyasr: Ah yeah I completely forgot that ucx-py is not built with scikit-build-core or CMake, so we need to provide the library location another way. Rather than rewriting this logic manually, which makes ucx-py sensitive to changes in the libucx wheel, I think you could instead do something like what I've proposed as a change.

vyasr, line 28: Try this instead and see if it works
```suggestion
    import libucx
    libs = libucx.load_library()
    libucx_dir = os.path.dirname(libs[0]._name)
    library_dirs.append(libucx_libdir)
```

vyasr (changes requested):
Looks great overall!

conda/recipes/ucx-py/meta.yaml
jameslamb: hmmm this *might* have been before I added the `if not r.startswith("libucx")` a few lines down below, in the part where we read in all the host + run dependencies from the corresponding tables in `pyproject.toml`.

I can doble-check that.

setup.py
jameslamb: Happy to try this, but for my own understanding, why is it preferable?

And I guess the `try-except` will now also need to catch `OSError` and `RuntimeError` to account for these other conditions: https://github.com/rapidsai/ucx-wheels/blob/ef470edb120771255b5161b29453592073e763b4/python/libucx/libucx/load.py#L37-L52

dependencies.yaml
jameslamb: blegh yeah this is left over from an earlier version. Thank you!

conda/recipes/ucx-py/meta.yaml
jameslamb:
```suggestion
    - {{ PYTHON }} -m pip install . -vv
```

jakirkham: Agree this shouldn't be needed, but the additional flags are safe to add

jameslamb: Yep looks like that was the problem... filtering `libucx` out of every place where we read `pyproject.toml` deps into conda recipe deps with jinja was enough, and these flags were unnecessary.

successful conda builds: https://github.com/rapidsai/ucx-py/actions/runs/8991800910/job/24700209674?pr=1041

Thanks for noting it, I didn't know before this conversation that `--no-deps` was the default in `conda-build`, that's helpful!

jameslamb: Thanks @jakirkham , looks like we posted at the same time. As long as it's not needed, I'll leave this reverted to shrink the size of the changeset here.

setup.py
vyasr: Oh hmm that is a good point about those extra exceptions. A couple of notes:
- I preferred this approach because it makes ucx-py insensitive to the internal file layouts of the ucx wheels. By deferring to the libucx package's `load_library` function, ucx-py doesn't have to know the file layout in libucx.
- I hadn't considered the extra exceptions, and it's making me rethink my suggestion. I was assuming that the libraries would be safely loadable in the environment where we're building, but that may not be true. We want (and need) ucx-py to be possible to build in environments that do not actually have GPUs (like our CI runners), and I don't know if we're guaranteed that libcuda.so will actually load in such environments. It's worth testing out but if this raises exceptions then I don't think they're worth handling because I don't think there's any way to recover from that. If that happens we should probably skip my suggestion here and just stick to your original code.

jameslamb: > *because it makes ucx-py insensitive to the internal file layouts of the ucx wheels*

Ahhh yeah great point, I agree that that's desirable.

> *We want (and need) ucx-py to be possible to build in environments that do not actually have GPUs*

Ok yeah I just tried this suggested change without a GPU available, and got one of those errors from `libucx.load_library()`:

```text
RuntimeError: The CUDA driver library libcuda.so.1 was not found on your system.
This library cannot be provided by the libucx wheel and must be installed separately.
```

So I don't think we should use `libucx.load_library()`.

I just pushed a commit with a different approach: https://github.com/rapidsai/ucx-py/pull/1041/commits/0523ba1905d9607962e45a04b5cd751859b3a503

```python
import glob
import libucx

# find 'libucx'
module_dir = os.path.dirname(libucx.__file__)

# find where it stores files like 'libucm.so.0'
libs = glob.glob(f"{module_dir}/**/lib*.so*", recursive=True)

# deduplicate those library paths
lib_dirs = {os.path.dirname(f) for f in libs}
if not lib_dirs:
    raise RuntimeError(
        f"Did not find shared libraries in 'libucx' install location ({module_dir})"
    )

# add all the places 'libucx' stores libraries to that paths
# considered by the linker when compiling extensions
library_dirs.extend(lib_dirs)
```

It's a little more complicated, but not too much, and it avoids assuming anything about the file layout in the `libucx` wheel.

I tested that locally using the approach described in the PR description, everything worked well.

.readthedocs.yml
vyasr: I agree, that seems a bit cleaner to me.

Did you try `PIP_NO_DEPENDENCIES`? I don't remember which is the right name (or maybe both work), and also the behavior of true/false for pip's `*_NO_*` variables can be confusing so it's worth double-checking all the options with the setting that you proposed in the env.yaml before we set this on RTD. If all else fails, though, I'm fine with setting on RTD.

jameslamb: > Did you try `PIP_NO_DEPENDENCIES`?

I didn't, but that'd just take a minute. I'll do that.

The `pip` docs mention both of them: https://pip.pypa.io/en/latest/cli/pip_install/#cmdoption-no-deps

vyasr: OK cool. If that doesn't work, I'm fine with setting the flags appropriately in our RTD env.

jameslamb: Ok yeah, none of these variations worked:

```yaml
variables:
  - PIP_NO_DEPENDENCIES: true
  
variables:
  - PIP_NO_DEPENDENCIES: "true"

variables:
  - PIP_NO_DEPENDENCIES: "on"

variables:
  - PIP_NO_DEPENDENCIES: "off"

variables:
  - PIP_NO_DEPENDENCIES: 1
```

@pentschev could you and set environment variable `PIP_NO_DEPENDENCIES` to "True" for readthedocs builds here, following https://docs.readthedocs.io/en/stable/guides/environment-variables.html? One you do that, I'll push a commit to test.

setup.py
jameslamb: cool, thanks for talking through it with me

.readthedocs.yml
pentschev: Done.

pentschev: @jameslamb I invited you to be a RTD maintainer as well, this might make things easier for you to test.

jameslamb: Excellent, thank you!! Testing right now.

jameslamb: Alright that failed ([build link](https://readthedocs.org/projects/ucx-py/builds/24324188/)), I think because the environment variable you set up was marked "private".

From https://docs.readthedocs.io/en/stable/environment-variables.html#custom-env-var-exceptions

> *Custom environment variables that are not marked as Public will not be available in [pull request builds](https://docs.readthedocs.io/en/stable/pull-requests.html)*

I just created a new public one and kicked off a rebuild... unfortunately that failed too 😭 ([build link](https://readthedocs.org/projects/ucx-py/builds/24324249/))

Thanks for adding me as a maintainer, I'll use those permissions to keep trying stuff.

jameslamb: Ahhhhh I forgot that `PIP_NO_DEPENDENCIES` does not stop `pip install` from trying to download **build** dependencies.

```text
Processing /home/docs/checkouts/readthedocs.org/user_builds/ucx-py/checkouts/1041
  Installing build dependencies: started
  Installing build dependencies: finished with status 'error'
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [4 lines of output]
      Collecting cython>=3.0.0
        Downloading Cython-3.0.10-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (3.2 kB)
      ERROR: Could not find a version that satisfies the requirement libucx==1.15.0 (from versions: none)
      ERROR: No matching distribution found for libucx==1.15.0
      [end of output]
```

([build link](https://readthedocs.org/projects/ucx-py/builds/24324302/))

I just added `PIP_NO_BUILD_ISOLATION="True"` to the environment as well... still didn't work 😞 

([build link](https://readthedocs.org/projects/ucx-py/builds/24324476/))

I have one other idea, will post in a second.

jameslamb: To summarize.... these things did not work:

* set env variable `PIP_NO_DEPS="True"` in readthedocs builds
* set env variable `PIP_NO_BUILD_ISOLATION="True"` in readthedocs builds
* set both of the above in readthedocs builds
* set any combination of those environment variables in `variables: ` block in `builddocs.yml` conda env
* pass `pip: - --no-deps ../../` via `builddocs.yml` conda env
* use a real, suffixed dependency name like `libucx-cu11>=1.15.0` in `pyproject.toml` + add `--extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple` in `builddocs.yml` conda env

That last one failing revealed a more significant issue, not limited to just docs building...

**calling `libucx.load_library()` unconditionally means `ucx-py` will not be importable unless the process has access to a GPU** 😱 

```text
RuntimeError: The CUDA driver library libcuda.so.1 was not found on your system. This library cannot be provided by the libucx wheel and must be installed separately.
```

([build link](https://readthedocs.org/projects/ucx-py/builds/24325258/))

The docs DO BUILD successfully on the current state of this branch 🎉 ([build link](https://readthedocs.org/projects/ucx-py/builds/24325407/)), but only because I added behavior to work around not having a GPU available.

So @pentschev @vyasr that's a design decision for you.

When someone tries to `import ucp` without access to a GPU, what should happen?

* the import should fail
* the import should warn but succeed
* something else

jameslamb: Summarizing our offline conversations:

* yes `ucx-py` is expected to be usable without a GPU
* the `libucx` wheels should stop raising exceptions when `libucx.load_library()` is called in an environment without a GPU

In response, we published a new `libucx==1.15.0.post1` removing those exceptions: https://github.com/rapidsai/ucx-wheels/pull/5

Tested it with `ucxx` on systems with and without a GPU, and it seemed to work well: https://github.com/rapidsai/ucx-wheels/pull/5#issuecomment-2103619706

Those new wheels helped the docs builds here get further, but those builds exposed another issue... as of the changes in this PR, `ucx-py` was compiling with the **system-installed headers** (e.g. `/usr/include/ucm`) but the **wheel-provided shared libraries** (e.g. `site-packages/libucx/lib/libucm.so.0`).

That showed up in "symbol not found" issues like this:

```text
OSError: ${HOME}/conda/1041/lib/python3.12/site-packages/libucx/lib/libucs_signal.so: undefined symbol: ucs_debug_is_error_signal
```

([build link](https://readthedocs.org/projects/ucx-py/builds/24342353/))

I just pushed a change that fixes that: https://github.com/rapidsai/ucx-py/pull/1041/commits/eba110fe2116978e9488b46b7d884ab6b5ffb876

And looks like the docs builds are happy and docs are rendering correctly ([RTD build link](https://readthedocs.org/projects/ucx-py/builds/24343567/)).

setup.py
```diff
@@ -5,18 +5,67 @@
...
+import glob
 import os
 from distutils.sysconfig import get_config_var, get_python_inc
 
 from Cython.Distutils.build_ext import new_build_ext
 from setuptools import setup
 from setuptools.extension import Extension
 
+
+def _find_libucx_libs_and_headers():
```
jameslamb, line 17: Adding the headers too made this long enough that I decided it was time for it to go into a function.

For context, see https://github.com/rapidsai/ucx-py/pull/1041#discussion_r1596916903

ci/build_docs_pre_install.sh
```diff
@@ -0,0 +1,17 @@
...
+#   It does that by running 'pip install .' from the root of the repo. This script
+#   is used to modify readthedocs' local checkout of this project's source code prior
+#   to that 'pip install' being run.
+#
+#   For more, see https://docs.readthedocs.io/en/stable/build-customization.html
+#
+
+set -euo pipefail
+
+sed -r -i "s/libucx==(.*)\"/libucx-cu12==\1\"/g" ./pyproject.toml
```
jameslamb, line 17: I'm proposing bringing back this script for docs builds, but having it substitute in a real, suffixed name for `libucx-cu12` so that docs builds will really install that.

I think that's necessary to meet these 3 constraints:

* docs builds run `pip install .` to install `ucx-py`
* fallback matrices in `dependencies.yaml` need to be the unsuffixed names (e.g. `libucx` not `libucx-cu12`)
   - *per the offline discussion that came out of https://github.com/rapidsai/rapids-build-backend/pull/29#issuecomment-2103461693*
* `pre-commit` hook running `rapids-dependency-file-generator` is going to put into `pyproject.toml` whatever is in the fallback matrices in `dependencies.yaml`

.readthedocs.yml
vyasr: Good catch on the headers! I forgot to check on those. Glad you found this issue.

setup.py
vyasr: Works for me, thanks!

vyasr (approved):

raydouglass (approved):

.readthedocs.yml
jameslamb: Update for those finding this thread from search or other links.... thanks to @bdice 's suggestion in https://github.com/rapidsai/ucx-py/pull/1044#discussion_r1636690394, we were able to get this working without messing with environment variables or the conda environment YAML file, like this in `.readthedocs.yaml`:

```yaml
    post_create_environment:
      - |
        pip install \
          --extra-index-url "https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/" \
          -C rapidsai.disable-cuda=true \
          -C rapidsai.matrix-entry="cuda=12.2" \
          .
```

PR #1042, opened 2024-05-06
remove 'tomli' dependency

jameslamb:
Proposes removing the build-time dependency on `tomli` for wheels and conda packages.

Noticed that working on #1041. It doesn't appear to be used anywhere here.

```shell
git grep tomli
```

## Notes for Reviewers

That dependency was added back in #895. I'm not sure why, but I suspect it was related to the use of `versioneer` in this project at the time. Reference: https://github.com/python-versioneer/python-versioneer/issues/338#issuecomment-1381170813

This project doesn't use `versioneer` any more (#931). I strongly suspect that the dependency on `tomli` can be removed.


jakirkham (approved):

raydouglass (approved):

jakirkham:
Thanks all! 🙏

PR #1044, opened 2024-06-05
use rapids-build-backend

jameslamb:
Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89
Contributes to https://github.com/rapidsai/build-planning/issues/72

Proposes:

* introducing `rapids-build-backend` as this project's build backend, to reduce the complexity of various CI/build scripts
* making get-data-from-`pyproject.toml` code in conda recipe a bit stricter

## Notes for Reviewers

This reverts some of the workarounds introduced for docs builds in #1041. See this thread for context: https://github.com/rapidsai/ucx-py/pull/1041#discussion_r1592789721

~To do that, I set 2 environment variables manually in in the readthedocs console ([link](https://readthedocs.org/dashboard/ucx-py/edit/)):~

* ~`RAPIDS_DISABLE_CUDA=true` = used to prevent needing `nvcc` in the docs-building environment~
* ~`RAPIDS_MATRIX_ENTRY='cuda=12.2'` = used to ensure that `pip install .` tries to install a real package (e.g. `libucx-cu12`)~

~I think this is the first place we're using the `rapids-build-backend` environment variable support.~

see https://github.com/rapidsai/ucx-py/pull/1044#discussion_r1636690394


dependencies.yaml
```diff
@@ -136,8 +150,6 @@ dependencies:
           - matrix: {cuda: "11.*"}
             packages:
               - libucx-cu11==1.15.0
-          # NOTE: this fallback needs to be a real, suffixed version
-          #       so 'pip install .' (e.g. as used in docs builds) will work
```
jameslamb, line 140: I accidentally left behind this comment in #1041 ... it's not accurate on `branch-0.39` or on this branch.

conda/environments/builddocs.yml
```diff
@@ -16,5 +16,15 @@ dependencies:
...
+  # NOTE: this 'pip install' working correctly depends on two environment variables
+  #       used by rapids-build-backend, set in the readthedocs UI
+  #
+  #       # do not try to run 'nvcc --version' to determine CUDA version
+  #       RAPIDS_DISABLE_CUDA=true
+  #
+  #       # install CUDA-12-specific dependencies
+  #       # (e.g. libucx-cu12)
+  #       RAPIDS_MATRIX_ENTRY="cuda=12.2"
+  #
```
jameslamb, line 28: Thanks to @pentschev for giving me access to set environment variables in the readthedocs console for this project 😁

conda/environments/builddocs.yml
```diff
@@ -16,5 +16,15 @@ dependencies:
...
   - ../../
```
bdice, line 29: It looks like we use a conda env with a pip dependency on `../..`. I think it would be clearer to install a conda env with the relevant dependencies and then invoke `pip install` with the appropriate flags on the command line instead of requiring environment variables in the UI. Is that an option?

conda/recipes/ucx-py/meta.yaml
```diff
@@ -30,15 +30,18 @@ requirements:
...
     - pip
     - ucx
+    {% for b_r in data["build-system"]["requires"] %}
+    - {{ b_r }}
+    {% endfor %}
     # 'libucx' wheel dependency is unnecessary... the 'ucx' conda-forge package is used here instead
-    {% for r in data.get("build-system", {}).get("requires", []) if not r.startswith("libucx") %}
-    - {{ r }}
+    {% for rbb_r in data["tool"]["rapids-build-backend"]["requires"] if not rbb_r.startswith("libucx") %}
+    - {{ rbb_r }}
```
bdice, line 38:
```suggestion
    {% for rbb_req in data["tool"]["rapids-build-backend"]["requires"] if not rbb_req.startswith("libucx") %}
    - {{ rbb_req }}
```

conda/recipes/ucx-py/meta.yaml
```diff
@@ -30,15 +30,18 @@ requirements:
     - python
     - pip
     - ucx
+    {% for b_r in data["build-system"]["requires"] %}
+    - {{ b_r }}
```
bdice, line 34:
```suggestion
    {% for build_req in data["build-system"]["requires"] %}
    - {{ build_req }}
```

conda/recipes/ucx-py/meta.yaml
```diff
@@ -30,15 +30,18 @@ requirements:
...
+    {% for rbb_r in data["tool"]["rapids-build-backend"]["requires"] if not rbb_r.startswith("libucx") %}
+    - {{ rbb_r }}
     {% endfor %}
   run:
     - python
     - ucx >=1.15.0,<1.16.0
     # 'libucx' wheel dependency is unnecessary... the 'ucx' conda-forge package is used here instead
-    {% for r in data.get("project", {}).get("dependencies", []) if not r.startswith("libucx") %}
+    {% for r in data["project"]["dependencies"] if not r.startswith("libucx") %}
     - {{ r }}
```
bdice, line 45:
```suggestion
    {% for req in data["project"]["dependencies"] if not req.startswith("libucx") %}
    - {{ req }}
```

conda/recipes/ucx-py/meta.yaml
```diff
@@ -47,12 +50,12 @@ test:
     - ucp
 
 about:
-  home: {{ data.get("project", {}).get("urls", {}).get("Homepage", "") }}
-  license: {{ data.get("project", {}).get("license", {}).get("text", "") }}
+  home: {{ data["project"]["urls"]["Homepage"] }}
```
bdice, line 53: I thought the `.get(...)` was required because conda-build doesn't actually inject the Jinja variables on every pass and that led to problems in metadata. I'm not sure. We use the `.get(...)` logic in dask-cuda, too. https://github.com/rapidsai/dask-cuda/blob/branch-24.08/conda/recipes/dask-cuda/meta.yaml

tests/test_version.py
```diff
@@ -9,6 +9,11 @@ def test_get_ucx_version():
     assert ucp.core._ctx is None
 
 
+def test_git_commit_constant():
+    # __git_commit__ will only be non-empty in a built distribution
+    assert isinstance(ucp.__git_commit__, str)
+
+
 def test_version_constant():
```
bdice, line 17: We have a test in the other RAPIDS libraries that the version string has nonzero length. Can you take a look at one of the other recent RBB PRs and do the same here?

bdice (changes requested):
A few suggestions, but overall this is pretty close.

conda/environments/builddocs.yml
jameslamb: It's not an option. `readthedocs` doesn't allow you to customize the flags passed to `pip install`.

You can see the history that led to this pattern here: https://github.com/rapidsai/ucx-py/pull/1041#discussion_r1592789721

conda/recipes/ucx-py/meta.yaml
jameslamb: It looks to me like this is working as expected.

```shell
wget \
  https://downloads.rapids.ai/ci/ucx-py/pull-request/1046/43beaa1/ucx-py_conda_python_cuda12_py39_x86_64.tar.gz

tar -xvzf ./ucx-py_conda_python_cuda12_py39_x86_64.tar.gz
cat ./linux-64/repodata.json | jq .
```

For example, I see the right `license:` string there.

And looking in one of the packages, all this metadata looks correct too:

```shell
mkdir ./delete-me
cp ./linux-64/ucx-py-0.39.00a1-py39_240612_g43beaa1_1.tar.bz2 ./delete-me
cd ./delete-me
cph extract ./ucx-py-0.39.00a1-py39_240612_g43beaa1_1.tar.bz2
cat ucx-py-0.39.00a1-py39_240612_g43beaa1_1/info/about.json | jq .
```

Unless you can remember the specific reason to prefer `.get()`, I think we should move away from it in all conda recipes, to turn "silently ignored missing metadata" cases into "got a loud error at build time" cases. I actually just wrote that up in https://github.com/rapidsai/build-planning/issues/72.

conda/environments/builddocs.yml
bdice: Hmm. I saw there were options to run arbitrary commands like installing with `poetry` or `uv`, so I figured we could do the same with `pip` and custom flags. https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry

conda/recipes/ucx-py/meta.yaml
bdice: Ok cool! Go for it.

bdice: @jakirkham may remember the specific reason why we used `.get(...)` in these recipes, I recall discussing it with him at some point.

bdice (approved):
All seems fine once you're comfortable with the state of the conversations from my last review.

conda/environments/builddocs.yml
jameslamb: !!! in all of that conversation in #1041, I was so focused on "how do we change the `install:` step", I never considered the possibility "just ignore it completely and install the library through one of the other run custom commands" options. The tunnel vision was real 😭 

Thanks for that link, let me try this.

tests/test_version.py
pentschev: +1 on this suggestion.

pentschev (approved):
LGTM, I agree with @bdice 's suggestion on version test, other than that we're good. Thanks @jameslamb !

conda/environments/builddocs.yml
jameslamb: This worked! I can't believe I missed that going through options in #1041, thank you so much 😅 

I removed the environment variables in the readthedocs console, then pushed the changes you now see in `.readthedocs.yaml` on this PR.

* build: https://readthedocs.org/projects/ucx-py/builds/24674174/
* docs: https://ucx-py--1044.org.readthedocs.build/en/1044/

I think this is unambiguously better than having the `pip:` section in a conda environment YAML file.

tests/test_version.py
jameslamb: Yes you're right, I missed that.

Decided to just overwrite the existing tests on `__git_commit__` and `__version__` with the one we've been copying around all the other repos for the `rapids-build-backend` work.

```python
def test_version_constants_are_populated():
    # __git_commit__ will only be non-empty in a built distribution
    assert isinstance(ucp.__git_commit__, str)

    # __version__ should always be non-empty
    assert isinstance(ucp.__version__, str)
    assert len(ucp.__version__) > 0
```

conda/recipes/ucx-py/meta.yaml
jameslamb: I found a related thread: https://github.com/rapidsai/pynvjitlink/pull/33#discussion_r1449525892

I don't think it changes the fact that this is working and should be used for the specific places I've made this change in this PR, and doesn't need to block this PR. But might mean "never use `.get()`" is too strong.

I'll post over in the `build-planning` issue.

conda/environments/builddocs.yml
bdice: Aw Yeah

conda/recipes/ucx-py/meta.yaml
bdice: Maybe `conda-build` has changed. 🤷‍♂️ In any case, I am happy to drop `.get()` if this is working.

Issue #1047, opened 2024-06-13
[QST] How to install from source with system's ucx libs


cctry:
Hi,

I am trying to install uxc-py from source using`pip install .` with libucx installed in $CONDA_PREFIX/lib.
I got the error about trying to find the wheel for libucx.
> ERROR: Could not find a version that satisfies the requirement libucx-cu12==1.15.0 (from versions: 1.14.1.post1, 1.15.0.post1, 1.16.0.post1)

How should I avoid this dependency check or direct it to the system's version?

pentschev:
Thanks for your question @cctry , there have been a [few recent changes to our build system](https://github.com/rapidsai/ucx-py/pull/1044) that we also noticed yesterday were breaking changes, we're looking to [have them fixed](https://github.com/rapidsai/ucx-py/pull/1048) by today or early next week. There's also a chance that we'll need to pass more options to `pip install .` which we will document accordingly and update this issue then. If this is an immediate blocker for you, reverting https://github.com/rapidsai/ucx-py/pull/1044 should allow `pip install .` still but that's not going to be a long-term solution.

pentschev:
With https://github.com/rapidsai/ucx-py/pull/1048 now merged, installing via `pip install .` should now be supported again, apologies for the longer than expected delay in hashing out all details.

Since you're using conda, the simpler approach for you is to install the [`ucx` package from conda-forge](https://anaconda.org/conda-forge/ucx), that will most likely cover your needs and you'll still be able to install UCX-Py from source, please refer to the [conda installation docs here](https://ucx-py.readthedocs.io/en/latest/install.html#id2). If you absolutely need the `libucx-cu*` package from PyPI that should be supported as well but it is a much more involved process and much less user-friendly as it requires setting `LD_LIBRARY_PATH` both for install and runtime, refer to [the docs here](https://ucx-py.readthedocs.io/en/latest/install.html#id3). Unfortunately using a custom UCX-Py install with PyPI packages without setting `LD_LIBRARY_PATH` would be a much more challenging and time-consuming effort and thus we do not intend in supporting at the moment.

I'll tentatively close this issue as I believe your case is now resolved, but please feel free to reopen this or open a new issue should you encounter any issues.

PR #1048, opened 2024-06-13
make pip source installs a bit easier

jameslamb:
Follow-up to #1044
Related to #1047

* updates `libucx` build requirements so they can be satisfied by only packages from pypi.org
* removes `--extra-index-url https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/` in `pip install` calls, now that `rapids-build-backend` is on pypi.org (https://pypi.org/project/rapids-build-backend/)
* sets `rapids-build-backend` config `disable-cuda=true` in `pyproject.toml`
  - *this means that now `pip install .` will not require `nvcc` or produce a wheel with a suffix like `-cu12`*
  - *modified CI script used to build wheels to override this, so the published wheels will still have `-cu${ver}` suffixes
* updates docs on source installation to reflect these changes

## Notes for Reviewers

These changes came out of an offline conversation with @pentschev and @vyasr


dependencies.yaml
```diff
@@ -142,17 +142,19 @@ dependencies:
...
           - --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
     specific:
+      # very tight >=x.x.x,<x.x.(x+1) here allows for installation of
+      # post release like 1.15.0.post1
       - output_types: [requirements, pyproject]
         matrices:
           - matrix: {cuda: "12.*"}
             packages:
-              - libucx-cu12==1.15.0
+              - libucx-cu12>=1.15.0,<1.15.1
```
jameslamb, line 151: Once I removed the `--extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple`, I saw the docs build fail like this:

```text
ERROR: Could not find a version that satisfies the requirement libucx-cu12==1.15.0 (from versions: 1.14.1.post1, 1.15.0.post1, 1.16.0.post1)
ERROR: No matching distribution found for libucx-cu12==1.15.0
```

([build link](https://readthedocs.org/projects/ucx-py/builds/24689587/)

Exactly as reported in #1047.

It looks like a requirement like `==1.15.0` will not match a version like `1.15.0.post` 🤦🏻 

Looked back at a recent successful CI run from a previous job, and I see it was downloading `libucx==1.15.0`, NOT `libucx==1.15.0.post1`.

```text
Downloading https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/libucx-cu11/1.15.0/libucx_cu11-1.15.0-py3-none-manylinux_2_28_x86_64.whl (25.3 MB)
```

([build link](https://github.com/rapidsai/ucx-py/actions/runs/9486356769/job/26140585007))

I think we didn't notice this before because:

* the RAPIDS nightly index is passed as an `--extra-index-url` in CI
* the RAPIDS nightly index contains both `libucx==1.15.0` and `libucx==1.15.0.post1`

But we're now noticing it in the docs build because:

* pypi.org does not have `libucx==1.15.0`, only `libucx==1.15.0.post1`

How I tested that this new specifier would work:

```text
pip install --dry-run 'libucx-cu12==1.15.0'
# ERROR: Could not find a version that satisfies the requirement libucx-cu12==1.15.0 (from versions: 1.14.1.post1, 1.15.0.post1, 1.16.0.post1)

pip install --dry-run 'libucx-cu12>=1.15.0,<1.15.1'
# Would install libucx-cu12-1.15.0.post1
```

cc @vyasr

dependencies.yaml
```diff
@@ -142,17 +142,19 @@ dependencies:
...
           - matrix: {cuda: "11.*"}
             packages:
-              - libucx-cu11==1.15.0
+              - libucx-cu11>=1.15.0,<1.15.1
           - matrix: null
             packages:
-              - libucx==1.15.0
+              - libucx==1.15.0,<1.15.1
```
vyasr, line 157: In this case, should the null matrix entry actually be empty? Since it is valid to build without CUDA, this is a bit different from other RAPIDS libs where we'd specify this matrix entry to satisfy the use case where you're building locally (not wheels) but also building dependencies (e.g. unified devcontainer).

Not a big deal either way since ucx-py is going away soon in favor of ucxx, but it's worth thinking about since this could impact the corresponding ucxx dependencies.

docs/source/install.rst
```diff
@@ -170,6 +170,6 @@ Building and installing UCX-Py can be done via `pip install`. For example:
     conda activate ucx
     git clone https://github.com/rapidsai/ucx-py.git
     cd ucx-py
-    pip install -v .
+    pip install -v -C rapidsai.matrix-entry="cuda=12.2" .
```
vyasr, line 173: Rather than always including this, should the default development instructions not enable CUDA, and therefore not specify a matrix entry? This assumes we make the other change I suggest above.

jameslamb: I think this is a question for @pentschev ?

These docs are under a section that says it's assumed you're building on a system with CUDA.

https://github.com/rapidsai/ucx-py/blob/7b70211452d73ef8a0486b0b8ffa04777ee6fef4/docs/source/install.rst#L50-L53

I'm happy to change this PR either way, but I don't feel qualified to make the call about whether or not the CUDA-enabled build should be the default.

I DO think that we should continue specifying this flag (and therefore depending on the `libucx` package) in docs builds here. Since we've seen that those docs builds end up also carrying some test coverage to catch works-without-CUDA types of issues (https://github.com/rapidsai/ucxx/issues/231).

So @pentschev , the specific question for you is, which of these would you prefer?

* `pip install .` works with 0 other flags, but requires a system installation of UCX
* `pip install .` requires passing `-C rapidsai.matrix-entry="cuda12.2"` or similar, but in exchange the locally-built package depends on `libucx` wheels (so no system version required)

jameslamb: I just pushed https://github.com/rapidsai/ucx-py/pull/1048/commits/7b675a48f80c66d84e6230d0ca19b246e866816a implementing this, by the way, so you can look at the diff and see what it'd look like.

pentschev: The docs assume a CUDA system because of the `--with-cuda=$CUDA_HOME` build flag, which is our primary use case, if that flag is omitted then it should work on non-CUDA systems as well, but we don't document that at the moment.

> the specific question for you is, which of these would you prefer?
> 
> * `pip install .` works with 0 other flags, but requires a system installation of UCX
> * `pip install .` requires passing `-C rapidsai.matrix-entry="cuda12.2"` or similar, but in exchange the locally-built package depends on `libucx` wheels (so no system version required)

Can I have both? 🙂 

Ideally, I would like both to work, but given your phrasing it seems this is not possible. So let's assume you have UCX installed on the system and `libucx` installed, what would happen then both with and without `-C rapidsai.matrix-entry="cuda12.2"`?

jameslamb: > *assume you have `libucx` installed*

This question is about whether or not `pip install .` pulls that dependency in **for you**, not about what happens if it happens to already be installed.

If you've already done something like the following:

```shell
apt-get install openucx
pip install libucx-cu12
```

Then `import ucp` will call `libucx.load_library()`, which should find that system installation of `libucs`, `libucm`, etc., but will fall back to the `libucx`-bundled ones if finding them fails for some reason.

Ok so what's the point of `-C rapidsai.matrix-entry`? `libucx` isn't distributed as the name `libucx`.... you have to pick `libucx-cu11` (CUDA 11) or `libucx-cu12` (CUDA 12). That selection of major CUDA version can happen automatically in `rapids-build-backend`, but that detection is being turned off here via the use of `disable-cuda = true`, to support source installation on non-CUDA systems by default.

Another option we could pursue, if you want, is to choose a specific CUDA verrsion as the default one for source installations by e.g. recording `libucx-cu12` as a dependency in `pyproject.toml`. Then `pip install .` would "just work" with no additional flags on a system with or without CUDA available and with or without a system installation of UCX. But anyone using CUDA 11 would have to pass `-C rapidsai.matrix-entry="cuda11.8"` or similar, or they'd get a `libucx-cu12` (built against CUDA 12), which might lead to runtime issues.

pyproject.toml
```diff
@@ -116,10 +115,12 @@ build-backend = "setuptools.build_meta"
 commit-files = [
     "ucp/COMMIT_FILE"
 ]
+# by default, do not rename the package 'ucx-py-cu${ver}'
+# (this is overridden in wheel publishing)
+disable-cuda=true
```
jakirkham, line 120: Does `ucx-py` actually do anything different between CUDA versions? My understanding was no

Only `libucx` would potentially do different things for different CUDA versions

So am curious why we want to suffix here. Is this just to provide a way to influence `libucx` indirectly? Or is there not really a need?

jameslamb: I'll let @pentschev or @vyasr elaborate, but I can share a bit.

> *why we want to suffix here*

I can say confidently it's not ONLY to influence the `libucx` dependency, because `ucx-py` has been published with CUDA suffixes as far back as v0.32 (https://pypi.nvidia.com/ucx-py-cu12/), and `libucx` was only introduced in v0.38 (#1041).

But now that we do have `libucx`... I the suffixing does serve that purpose of influencing which version you get.

e.g. `cugraph-cu12` depends on `ucx-py-cu12` which depends on `libucx-cu12`

> *Recognize we may just be agreeing*

I don't think so.

This `disable-cuda=true` in this part of the diff here is removing that suffixing **for source builds like `pip install .`**, based on @pentschev 's feedback that he wanted that workflow to be as simple as possible, especially since in some cases `ucx-py` is used without any CUDA at all.

If this PR was accepted in its current state, we'd still be publishing `ucx-py-cu{version}` wheels, via `pip wheel --config-settings rapidsai.disable-cuda=false` (see changes to `ci/build_wheel.sh` in this PR).

pentschev: > Does `ucx-py` actually do anything different between CUDA versions? My understanding was no
> 
> Only `libucx` would potentially do different things for different CUDA versions

John is right, UCX-Py doesn't have a direct dependency on CUDA.

> So am curious why we want to suffix here. Is this just to provide a way to influence `libucx` indirectly? Or is there not really a need?

I have previously asked the same question myself, which James answered above.

jakirkham: In that case, couldn't a user just specify the `libucx-cu*` install if they want that? What does it provide to have `ucx-py` also including the same information?

Would add on the Conda side we don't provide this distinction

pentschev: > In that case, couldn't a user just specify the `libucx-cu*` install if they want that? What does it provide to have `ucx-py` also including the same information?

I think the issue is that `ucx-py` needs `libucx`, if we don't encode the CUDA version in `ucx-py` then the user is responsible to install the appropriate `libucx-cu*` which can be error prone. I.e., installing `ucx-py` without `libucx-cu*` as a dependency will leave `ucx-py` in an unusable state, whereas `ucx-py` with a `libucx-cu*` dependency would make the correct CUDA version uninstallable (would `ucx-py` depend on `libucx-cu11` or `libucx-cu12`?). Please correct me if I'm wrong @jameslamb .

jameslamb: > *...on the Conda side we don't provide this distinction*

Right, because there's `ucx` from conda-forge which builds variants for different CUDA versions.

https://github.com/conda-forge/ucx-split-feedstock/blob/b60fb426784cbb6d5cbd384a76c86920737838f0/recipe/meta.yaml#L51-L53

And can constrain everything without needing to use names to disambiguate, by relying on the `cuda-version` metapackage.

With wheels and `pip` those same mechanisms don't exist (part of what @msarahan and others have been talking about in https://discuss.python.org/t/implementation-variants-rehashing-and-refocusing/54884).

If you want to use `ucx-py` with CUDA-enabled UCX, something somewhere has to get that CUDA-enabled UCX installed on the system... and one that was built against the correct version of CUDA. My understanding of https://github.com/rapidsai/build-planning/issues/57 was that this idea of distributing `libucx` wheels was a response to difficulties users were facing when asked to install UCX themselves outside of `ucx-py`.

pentschev: > this idea of distributing `libucx` wheels was a response to difficulties users were facing when asked to install UCX themselves outside of `ucx-py`.

I wouldn't even go as far as saying "users", even for ourselves it was quite challenging to bundle UCX with the libraries that need it, so much so that we had UCX broken for many months. One step further, this also bloated package sizes.

vyasr: Prior to the libucx wheels existing, ucx-py statically linked/bundled components of the CTK in order to support running with CUDA. In that world, yes it did have a CUDA dependency because the CUDA transport layer built into the wheel was specific to a given CUDA version. That CUDA dependency has now migrated to the libucx package, so the purpose of the suffix in ucx-py is now to select the appropriate libucx dependency.

> What does it provide to have ucx-py also including the same information?

This is generally what we've done throughout RAPIDS because if a pure Python package A depends on a CUDA-dependent package B there is no way to express the which CUDA version it should select for B without having separate versions of A. For instance, we do this in `dask-cudf-cuXY`, which has a hard dependency on `cudf-cuXY`. ucx-py is a bit of a special case because it works without CUDA, so the CUDA versions are truly optional unlike with something like `dask-cudf` (which requires `cudf` and therefore must pull a specific CUDA versioned one). In the past we've discussed is having some more automated way of having a default choice of CUDA version, but that still requires a default behavior so you can't get around the different packages in any way that I can tell.

For ucx-py, CUDA support is optional , and since https://github.com/rapidsai/ucx-wheels/pull/5 we allow installing libucx on a system without a GPU. I don't see an option for the dependency specification that provides a good UX in all cases though. You either have to force users to manually install dependencies or you make the package not work by default out of the box AFAICT.

jameslamb: Thanks for that! The used-to-statically-link-bits-of-the-CTK was a part I was missing.

I think influencing the `libucx` dependency via a CUDA suffix should continue to be the approach here.

docs/source/install.rst
jameslamb: Whenever you have time @pentschev , this is ready for another review.

I've taken @vyasr 's suggestions in this thread + https://github.com/rapidsai/ucx-py/pull/1048#discussion_r1651370374 and changes this so that there is no `libucx` dependency in `pyproject.toml` by default.

In the current state of this PR, wheels published to PyPI would still have that dependency, but running `pip install .` from the root of the repo would require that you provide UCX libraries some other way and should work with any of the following:

* building UCX yourself from source
  - *(like https://github.com/rapidsai/devcontainers/blob/branch-24.08/features/src/ucx/install.sh)*
* using an installation from another package manager
  - *(e.g. `apt-get install openucx` or `conda install ucx`)*
* installing the `libucx` wheels manually
  - *(e.g. `pip install libucx-cu{11,12}`)*

If you're ok with that being the development experience, then this PR could be merged as-is. The docs can continue to specify `pip install .` with 0 flags, like this:

https://github.com/rapidsai/ucx-py/blob/7b70211452d73ef8a0486b0b8ffa04777ee6fef4/docs/source/install.rst#L171-L173

If you want source installation to work without *requiring* a separate installation of UCX, then `ucx-py` built from source needs a `libucx` dependency.

And it could get it from any of the following:

* in the docs, instructing people to pass `-C rapidsai.matrix-entry="cuda=12.2"` (and similar for CUDA 11)
* putting a default matrix entry in `pyproject.toml` (e.g. either defaulting to CUDA 11 or CUDA 12)

pentschev: This looks great, I tested with system and conda installation of UCX and both worked as expected! Thanks a lot for pushing this through @jameslamb .

The only thing I would document, not just for a public reference of how it should be done but for my reference as well, is how to install the package with UCX wheels as well. Would `pip install --extra-index-url=https://pypi.nvidia.com libucx-cu1{1,2}; pip install -v .` work or is something more elaborate necessary?

pentschev: The above alone doesn't work (it still links to the system install), either with or with `-C rapidsai.matrix-entry="cuda=12.2"`, I suspect this is because the library is under `$CONDA_PREFIX/lib/python3.XY/site-packages/libucx` vs `$CONDA_PREFIX` when using the conda package. Is there a way to get the installer to find the `libucx` package?

jameslamb: > The above alone doesn't work (it still links to the system install),

It's expected that even if `libucx` is installed, system libraries will be found first. That's intentional behavior.

See:

https://github.com/rapidsai/ucx-wheels/blob/ff1946193342bdc75c082c3d6aa153d8cdd400a9/python/libucx/libucx/load.py#L37-L39

If you want to force the use of `libucx`, even when there's another installation of the UCX libraries available, you'd have to do one of the following:

* remove those other installations (delete the files or at least move them to a path not checked by the loader/linker)
* use environment variable `LD_LIBRARY_PATH` (or maybe `LD_PRELOAD`, depending on the context) to force the ones from the `libucx` wheel to be found first

Knowing that, do you still want me to add docs explaining how to do that?

> `--extra-index-url https://pypi.nvidia.com`

By the way, `libucx` is up on public PyPI: https://pypi.org/project/libucx-cu12/

So I don't think the `--extra-index-url` should be required. The packages there on `pypi.org` are the real wheels, not stubs that redirect to the NVIDIA index. And in fact, we don't even have `libucx` published on`pypi.nvidia.com`: https://pypi.nvidia.com/

pentschev: > It's expected that even if libucx is installed, system libraries will be found first.

Ah right, I forgot about that.

> * remove those other installations (delete the files or at least move them to a path not checked by the loader/linker)

That is usually not possible, if it's a system install where you're not an admin you can't do it.

> * use environment variable `LD_LIBRARY_PATH` (or maybe `LD_PRELOAD`, depending on the context) to force the ones from the `libucx` wheel to be found first

But in that case `LD_LIBRARY_PATH` needs to always be set, not just when building. Would there be any other way of telling `pip install` to link to libraries in a specific path such as `$CONDA_PREFIX/lib/python3.XY/site-packages/libucx`? I don't think it's worth spending too much time on this, but I'm picking your brain in case there's a simple solution, if there isn't one then we can document the `LD_LIBRARY_PATH` solution and be done with it.

> By the way, `libucx` is up on public PyPI: https://pypi.org/project/libucx-cu12/
> 
> So I don't think the `--extra-index-url` should be required. The packages there on `pypi.org` are the real wheels, not stubs that redirect to the NVIDIA index. And in fact, we don't even have `libucx` published on`pypi.nvidia.com`: https://pypi.nvidia.com/

Ah right, my bad. Even simpler then. 😄

jameslamb: No problem! There are lots of interconnected bits of this.

> Would there be any other way of telling `pip install` to link to libraries in a specific path such as `$CONDA_PREFIX/lib/python3.XY/site-packages/libucx`?

For as long as we want the default behavior to be that the libraries from the `libucx` wheels are only used if there are not any other installations of UCX around, I can't think of any other, "simple" ways to override that behavior besides those I listed above.

I think we should document and test the `LD_LIBRARY_PATH` approach (I can do that today), and not go further until we observe a use-case where managing the UCX installation via `pip` (e.g. being able to use `pip install libucx` to get a different version than the system install) is preferred and where the `LD_LIBRARY_PATH` approach is too cumbersome.

pentschev: > I think we should document and test the `LD_LIBRARY_PATH` approach (I can do that today), and not go further until we observe a use-case where managing the UCX installation via `pip` (e.g. being able to use `pip install libucx` to get a different version than the system install) is preferred and where the `LD_LIBRARY_PATH` approach is too cumbersome.

I'm absolutely of the same opinion, if there was a simple path that would be good, but it's not worth spending more time here. If you can document it, that would be great, I can also work on it and push to this branch if you prefer.

jameslamb: Thanks for talking through it with!

Ok sure, you could do it and push to this branch. You know this project's documentation and the audience for it much better than I do 😊

pentschev: Done in https://github.com/rapidsai/ucx-py/pull/1048/commits/a39811cc14034fc4890086dadd8e5022dddcdd28. Would you mind having a look if everything seems correct to you @jameslamb ?

Also, I've accidentally committed [this local change](https://github.com/rapidsai/ucx-py/pull/1048/files#diff-4ce4b8bd5b44b77d2e86d1f995f3c9e47e819dcf799a12fcd1e974c494e67cc4L10-R10) that I was going to push to another PR. It's harmless and corrects the environment now, so if you don't mind let's leave it.

jameslamb: I'd written up a long comment about how the docs should indicate that CPU-only is totally supported by `pip install ucx-py-cu{11,12}` .... then re-read what you wrote and saw that you already said exactly that 🤦🏻 

Anyway, in case it's interesting, here's how I tested that to convince myself:

<details><summary>how I tested that (click me)</summary>

Ran the following on my macOS laptop (so no possibility of accidentally finding CUDA):

```shell
docker run \
  --rm \
  -it python:3.11 \
  bash

# system-install UCX so we don't link to the the wheel one
apt-get update
apt-get install -y --no-install-recommends \
  libucx-dev

# install ucx-py-cu12 (which also pulls in libucx-cu12)
pip install --extra-index-url https://pypi.nvidia.com 'ucx-py-cu12>=0.38'

# import ucx-py and check where the loader found libraries
python -c "import ucp"
ldconfig -p | grep libuc
#	libuct.so.0 (libc6,AArch64) => /lib/aarch64-linux-gnu/libuct.so.0
#	libuct.so (libc6,AArch64) => /lib/aarch64-linux-gnu/libuct.so
#	libucs_signal.so.0 (libc6,AArch64) => /lib/aarch64-linux-gnu/libucs_signal.so.0
#	libucs_signal.so (libc6,AArch64) => /lib/aarch64-linux-gnu/libucs_signal.so
# ... etc., etc. ... it found them all

# write the demo code from https://github.com/rapidsai/ucx-py/blob/branch-0.39/docs/source/quickstart.rst
cat > server.py <<EOF
import asyncio
import time
import ucp
import numpy as np

n_bytes = 2**30
host = ucp.get_address(ifname='eth0')  # ethernet device name
port = 13337

async def send(ep):
    # recv buffer
    arr = np.empty(n_bytes, dtype='u1')
    await ep.recv(arr)
    assert np.count_nonzero(arr) == np.array(0, dtype=np.int64)
    print("Received NumPy array")

    # increment array and send back
    arr += 1
    print("Sending incremented NumPy array")
    await ep.send(arr)

    await ep.close()
    lf.close()

async def main():
    global lf
    lf = ucp.create_listener(send, port)

    while not lf.closed():
        await asyncio.sleep(0.1)

if __name__ == '__main__':
    asyncio.run(main())
EOF

cat > ./client.py <<EOF
import asyncio
import ucp
import numpy as np

port = 13337
n_bytes = 2**30

async def main():
    host = ucp.get_address(ifname='eth0')  # ethernet device name
    ep = await ucp.create_endpoint(host, port)
    msg = np.zeros(n_bytes, dtype='u1') # create some data to send

    # send message
    print("Send Original NumPy array")
    await ep.send(msg)  # send the real message

    # recv response
    print("Receive Incremented NumPy arrays")
    resp = np.empty_like(msg)
    await ep.recv(resp)  # receive the echo
    await ep.close()
    np.testing.assert_array_equal(msg + 1, resp)
    print("successfully received")

if __name__ == '__main__':
    asyncio.run(main())
EOF

# start the client and server
python ./server.py &
python ./client.py &
```

Saw output like this:

```text
Send Original NumPy array
Receive Incremented NumPy arrays
Received NumPy array
Sending incremented NumPy array
successfully received
```

</summary>

Maybe that'd be a helpful smoke test to add for https://github.com/rapidsai/ucxx/issues/231.

Anyway... everything you wrote looks correct and clear to me! I think this is good to merge. Thanks as always for your patience working through it with me 😊

pentschev (approved):
LGTM. Thanks @jameslamb !

docs/source/install.rst
pentschev: Smoke tests are always useful, my concern is how would that work in CI? Not sure if it's worth the trouble if we have to add another job at this time, from existing jobs it's probably relatively complicated to do so in a "safe" manner (i.e., ensuring linkage isn't resolved via the proper PyPI/conda UCX package).

I'm also +1 on merging this, I've approved the PR. Thanks so much James for the work here! 😄

jameslamb: > *Not sure if it's worth the trouble*

For sure, this wasn't a proposal to do add a new job or anything **right now**.

Thanks! I'll merge this once CI passes.

pentschev:
@vyasr maybe you have to approve too because you've been explicitly asked to review? Not sure to be honest.

jameslamb:
@pentschev This was my mistake, we still need a `rapidsai/packaging-codeowners` review.

image

I've asked @vyasr offline to to help us with one.

pentschev:
> @pentschev This was my mistake, we still need a `rapidsai/packaging-codeowners` review.

Oh, now I see that. I always miss the grey text. 😅

docker/ucx-py-cuda11.5.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   - python=3.9
   - cudatoolkit=11.5
   - setuptools
-  - cython>=0.29.14,<3.0.0a0
+  - cython>=3.0.0
```
vyasr, line 10: How did this change here?

docs/source/install.rst
```diff
@@ -173,3 +221,40 @@ Building and installing UCX-Py can be done via `pip install`. For example:
     pip install -v .
     # or for develop build
     pip install -v -e .
+
+
+UCX-Py with custom UCX install
```
vyasr, line 226: This whole section feels very complicated to document, and probably indicates that we should improve the UX. I've been working on a more general solution, but in the interim one way to make this more configurable at runtime would be for the `libucx.load_library` function to accept some argument(s) that controls the load order of the library. You could also have that controllable by a generic environment variable, or have ucx-py change the arguments it passes to that function based on an environment variable. That would be one way to force it to prefer using the library within the wheel in a more ergonomic manner than having to specify `LD_LIBRARY_PATH`.

docker/ucx-py-cuda11.5.yml
jameslamb: @pentschev accidentally committed it (it was intentional, just not intended for this PR) and proposed just keeping it: https://github.com/rapidsai/ucx-py/pull/1048#discussion_r1662715558

docs/source/install.rst
jameslamb: I had written something really similar to this but then stopped myself... these docs as written are an accurate description of the current state, and would be regardless of the changes in this PR.

Will defer to you and @pentschev on whether you want to block this PR until that interface is changed or just document that in https://github.com/rapidsai/ucx-wheels/issues as a feature request. I personally can't commit to the design work necessary to make a change like that to `libucx.load_library()` right now, but if you two come to an agreement about how you want it to look I could help with the implementation.

pentschev: I agree it's complicated but we don't have a simpler solution _right now_. James and I discussed that before and reached an agreement in https://github.com/rapidsai/ucx-py/pull/1048#discussion_r1662568912 . While I also agree `libucx.load_library` would be a nice solution, we don't have a true use case for that right now and I don't want that any of us has to spend the time working on a niche problem. If someone really _wants_ to do so in the near term, that would be nice, otherwise keeping things as is should be sufficient for now and thus I vote to leave it as is for this PR.

vyasr: OK got it, that's fine.

vyasr (approved):

docs/source/install.rst
pentschev: Actually I'm writing a response to #1047 , and while re-reading the original request the use case is exactly this one. With that said, I think `libucx.load_library()` would be the proper solution for that user's use case and thus there is an actual use case, unlike what I wrote above. However, it's also important to note that someone is going to need to take the time to implement that which I know is not any of us have lots of these days. Therefore, I'm thinking of replying with the up-to-date instructions (using `LD_LIBRARY_PATH`) and if the user requires something more sophisticated we should try and engage efforts from his side to get that done sooner as that is the only known use case right now. WDYT @jameslamb @vyasr ?

PR #1051, opened 2024-06-28
Increase maximum UCX runtime pin

pentschev:
Increase maximum UCX runtime pin to support UCX up to `v1.17.x`, but still build against `v1.15.0` until we gain confidence with newer UCX.

Additionally disable protov2 by default as CUDA async/managed memory is not yet supported and leads to poor performance.


pentschev:
Thanks @wence- for the suggestion, could you please have a look if the changes from https://github.com/rapidsai/ucx-py/pull/1051/commits/02bc2a53c99ea6c0533e7464646a454ce179c48e are sufficient?

pentschev:
I actually realized removing checks for older versions is not a good idea, we still provide support for UCX installed in the system from source and thus we cannot guarantee the installed versions are met. Therefore, @wence- please have a look at https://github.com/rapidsai/ucx-py/pull/1051/commits/02b148f873f81e0462c4b7b03913f1493b11a3e0 on top of https://github.com/rapidsai/ucx-py/commit/02bc2a53c99ea6c0533e7464646a454ce179c48e .

pentschev:
@wence- could you have a look at my comments above whether the changes address your suggestion satisfactorily?

docs/source/configuration.rst
```diff
@@ -45,6 +46,21 @@ UCX Environment Variables in UCX-Py
...
+Configuration variables applying to the UCP context.
+
+UCX_PROTO_ENABLE
+````````````````
+
+Values: y, n
+
+Enable the new protocol selection logic, also known as "protov2". Its default has been changed to ``y`` starting with UCX 1.16.0.
+
+The new protocol solves various limitations from the original "protov1" including, for example, invalid choice of transport in systems with hybrid interconnectivity, such as a DGX-1 where only a subset of GPU pairs are interconnected via NVLink. On the other hand, it may still lack proper support on not be as well tested for lesser common use cases, such as CUDA async and managed memory.
```
wence-, line 61:
```suggestion
The new protocol solves various limitations from the original "protov1" including, for example, invalid choice of transport in systems with hybrid interconnectivity, such as a DGX-1 where only a subset of GPU pairs are interconnected via NVLink. On the other hand, it may still lack proper support or not be as well tested for less common use cases, such as CUDA async and managed memory.
```

wence- (approved):

pentschev:
Thanks @wence- and @msarahan for reviews/approvals.

PR #1052, opened 2024-06-28
Remove entries for legacy UCX

pentschev:
Remove various tests, documentation and general support for older, now unsupported UCX versions.


pentschev:
This is not a good idea, we still provide support for UCX installed in the system from source and thus we cannot guarantee the installed versions are met.

PR #1057, opened 2024-07-24
split up CUDA-suffixed dependencies in dependencies.yaml

jameslamb:
## Description

In short, RAPIDS DLFW builds want to produce wheels with unsuffixed dependencies, e.g. `cudf` depending on `rmm`, not `rmm-cu12`.

This PR is part of a series across all of RAPIDS to try to support that type of build by setting up CUDA-suffixed and CUDA-unsuffixed dependency lists in `dependencies.yaml`.

For more details, see:
* https://github.com/rapidsai/build-planning/issues/31#issuecomment-2245815818
* https://github.com/rapidsai/cudf/pull/16183

### Why target 24.08?

This is targeting 24.08 because:

1. it should be very low-risk
2. getting these changes into 24.08 prevents the need to carry around patches for every library in DLFW builds using RAPIDS 24.08


jameslamb:
I've updated this based on the suggestions from https://github.com/rapidsai/cudf/pull/16183.

Ran the following to check `update-version.sh`.

```shell
git fetch upstream --tags
ci/release/update-version.sh '24.10.00'

git grep -E '24\.8|24\.08'
```

That revealed a few other places that needed updates in `update-version.sh`. Pushed those fixes here as well.

dependencies.yaml
```diff
@@ -196,20 +204,40 @@ dependencies:
...
               - cudf-cu11==24.8.*,>=0.0.0a0
               - &cupy_cu11 cupy-cuda11x>=12.0.0
+          - matrix:
+              cuda: "11.*"
+              cuda_suffixed: "false"
+            packages:
+              - *cudf_unsuffixed
+              # NOTE: cupy still has a "-cuda11x" suffix here, because it's suffixed
+              #       in DLFW builds
+              - *cupy_cu11
```
vyasr, line 239: This matrix looks like it hasn't been updated to use fallback the way that the other repo's PR have, but is that intentional here because of the need to specify the cupy dep?

jameslamb: yes exactly. More context: https://github.com/rapidsai/cugraph/pull/4552#discussion_r1690483837

.readthedocs.yml
```diff
@@ -10,7 +10,7 @@ build:
       # and therefore tested in this no-CUDA environment
       - |
         pip install \
-          -C rapidsai.matrix-entry="cuda=12.2" \
+          -C rapidsai.matrix-entry="cuda=12.2;cuda_suffixed=true" \
```
bdice, line 13: Was this missed in an upgrade to 12.5? Feel free to slip that in here or in a follow-up PR.

dependencies.yaml
bdice: If you want, it might be nice to refactor this to a "depends on cudf" / "[depends on cupy](https://github.com/rapidsai/raft/blob/759095aca27bec5ea37e072e2b1b0428d7eb0bf8/dependencies.yaml#L377)" structure like we have for other repos.

bdice (approved):

dependencies.yaml
jameslamb: I agree, I mentioned that in that comment I linked above (https://github.com/rapidsai/cugraph/pull/4552#discussion_r1690483837).

In the interest of time and not wanting to go through another CI cycle, I'll defer that to a follow-up PR.

.readthedocs.yml
jameslamb: Probably, yes. I'll put up a follow-up PR fixing this... against 24.10. There's not a functional difference for `ucx-py`, since this is just a way to matrix matrices with wildcard selectors like `"12.*"`.

jameslamb:
I merged this in the interest of reducing the amount of in-flight stuff ahead of code freeze. @bdice I made myself a reminder to put up a follow-up PR against 24.10 implementing your suggestions.

Don't want you to think I was ignoring them... I am just being really conservative with pushing new commits given how oversubscribed our CI runners are right now.

jakirkham:
Just know that RAPIDS 24.08 already includes 12.5

So if we need this to work with RAPIDS 24.08 CUDA 12.5, we would need that fix in 24.08

jameslamb:
> *So if we need this to work with RAPIDS 24.08 CUDA 12.5, we would need that fix in 24.08*

It's not necessary to address https://github.com/rapidsai/ucx-py/pull/1057#discussion_r1690530680 in 24.08.

That line is just a selector to choose between installing `libucx-cu11` and `libucx-cu12` in docs-building (an environment that doesn't have CUDA or GPUs), and all that matters there functionally is that it match a `dependencies.yaml` matrix `"12.*"`.

Functionally, it could be `"cuda=12.9999"` and still achieve the same goal.

I agree with changing the `12.2` to `12.5` just for the general benefit of defaulting to `12.5` everywhere as a form of  subtle documentation about what's supported, but there's no need to rush that change into 24.08.

jakirkham:
Ok cool. Could it just be `12.` then?

Maybe that avoids some confusion and saves future churn

jameslamb:
@jakirkham @bdice I've put up https://github.com/rapidsai/ucx-py/pull/1059, targeting `branch-0.40`, with changes that I think address all of your comments here.

PR #1059, opened 2024-07-25
refactor dependency-file-generator handling of cupy, default to CUDA 12 in fallback dependency lists

jameslamb:
Follow-up to #1057.

Implements the following suggestions based on review comments from @jakirkham and @bdice there:

* splits `cupy` out into its own list in `dependencies.yaml`, to allow removing some `cuda_suffixed: "false"` blocks that were otherwise redundant
* passes `--matrix-entry 'cuda=12.x'` in docs builds, to make it clearer that those builds are just trying to match `cuda: "12.*"` and are not at all dependent on any particular minor version of CUDA

Also proposes defaulting to `cupy-cuda12x` in `pyproject.toml`, as part of https://github.com/rapidsai/build-planning/issues/68.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,11 +6,11 @@ build:
     python: "mambaforge-22.9"
   jobs:
     post_create_environment:
-      # explicitly passing matrix-entry so that 'libucx' is pulled in,
-      # and therefore tested in this no-CUDA environment
+      # explicitly passing matrix-entry so that 'libucx' (with appropriate CUDA suffix)
+      # is pulled in, and therefore tested in this no-CUDA environment
       - |
         pip install \
-          -C rapidsai.matrix-entry="cuda=12.2;cuda_suffixed=true" \
+          -C rapidsai.matrix-entry="cuda=12.x;cuda_suffixed=true" \
           .
 
 conda:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -150,12 +150,12 @@ dependencies:
       - output_types: [requirements, pyproject]
         matrices:
           - matrix: {cuda: "12.*"}
-            packages: &cupy_packages_cu12
-              - cupy-cuda12x>=12.0.0
+            packages:
+              - &cupy_cu12 cupy-cuda12x>=12.0.0
           - matrix: {cuda: "11.*"}
             packages:
               - cupy-cuda11x>=12.0.0
-          - {matrix: null, packages: *cupy_packages_cu12}
+          - {matrix: null, packages: [*cupy_cu12]}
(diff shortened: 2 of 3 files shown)


dependencies.yaml
```diff
@@ -135,6 +138,24 @@ dependencies:
...
           - numpy>=1.23,<2.0a0
           - pynvml>=11.4.1
+  depends_on_cupy:
+    common:
+      - output_types: conda
+        packages:
+          - cupy>=12.0.0
+    # NOTE: This is intentionally not broken into groups by a 'cuda_suffixed' selector like
+    #       other packages with -cu{nn}x suffixes in this file.
+    #       All RAPIDS wheel builds (including in devcontainers) expect cupy to be suffixed.
```
jameslamb, line 148: this is exactly the language I'm proposing over in https://github.com/rapidsai/cugraph/pull/4552#discussion_r1691437257 as well

dependencies.yaml
```diff
@@ -135,6 +138,24 @@ dependencies:
...
+    specific:
+      - output_types: [requirements, pyproject]
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - &cupy_cu12 cupy-cuda12x>=12.0.0
+          - matrix: {cuda: "11.*"}
+            packages:
+              - cupy-cuda11x>=12.0.0
+          - {matrix: null, packages: [*cupy_cu12]}
```
bdice, line 158: Are we intentionally tackling https://github.com/rapidsai/build-planning/issues/68 here, or is that coincidental?

bdice (approved):

dependencies.yaml
jameslamb: Yep it's intentional. I mentioned it in the PR description, but not in the title, sorry about that. I've edited the title to make that clearer.

.readthedocs.yml
```diff
@@ -6,11 +6,11 @@ build:
...
   jobs:
     post_create_environment:
-      # explicitly passing matrix-entry so that 'libucx' is pulled in,
-      # and therefore tested in this no-CUDA environment
+      # explicitly passing matrix-entry so that 'libucx' (with appropriate CUDA suffix)
+      # is pulled in, and therefore tested in this no-CUDA environment
       - |
         pip install \
-          -C rapidsai.matrix-entry="cuda=12.2;cuda_suffixed=true" \
+          -C rapidsai.matrix-entry="cuda=12.x;cuda_suffixed=true" \
```
jakirkham, line 13: Nice! 😄

jakirkham (commented):
Thanks James! 🙏

PR #1064, opened 2024-08-19
Remove NumPy <2 pin

seberg:
This PR removes the NumPy<2 pin.  This is desirable for RAPIDS projects that do not have a strong numpy/cupy dependency.
Not doing any math, they don't run into subtle issues, but the pin was still necessary briefly before NumPy 2 was initially released.

(Other RAPIDS projects rely heavily on CuPy/NumPy and need the soon to be released 13.3.0 to function properly, e.g. due to promotion related changes in NumPy.)


dependencies.yaml
```diff
@@ -136,7 +136,7 @@ dependencies:
     common:
       - output_types: [conda, requirements, pyproject]
         packages:
-          - numpy>=1.23,<2.0a0
+          - numpy>=1.23,<3.0a0
```
jameslamb, line 139: Putting this in a comment on the diff so we can thread the conversation.

The description of this PR says that it'll be safe to remove this pin "once CuPy 13.3.0 is released".

It hasn't been yet.

* conda-forge: **13.2.0** ([link](https://anaconda.org/conda-forge/cupy))
* PyPI: **13.2.0** ([link](https://pypi.org/project/cupy-cuda12x/#history))

Could you share the relevant context on the issues with using CuPy 13.2.0 and NumPy 2.x? I don't see it at https://github.com/rapidsai/build-planning/issues/38.

jameslamb (changes requested):
Started a thread... I'd like some more context on the issues with CuPy 13.2.0 and NumPy 2.x.

dependencies.yaml
jakirkham: To provide some clarity, [CuPy 13.2.0]( https://github.com/cupy/cupy/releases/tag/v13.2.0 ) shipped with NumPy 2 support

However there were a couple of bugs that we wanted to fix upstream. Most notably these two:

* https://github.com/cupy/cupy/issues/8391
* https://github.com/cupy/cupy/issues/8392

We switched to a project board for tracking. Unfortunately that board is private (as that is the default visibility). There is nothing private there. However you can find it under [RAPIDS Projects]( https://github.com/orgs/rapidsai/projects ) where it is called "NumPy 2 Bringup"

Sebastian, myself and few others have been working with upstream to fix these bugs and test them using packages we have been building here: https://github.com/conda-forge/cupy-feedstock/pull/272

On the testing side, the changes in CuPy's `v13` are looking good

When talking with PfN offline earlier this week, they stated that we should be good to wrap things up and release Thursday (they are based in Japan so that may very well be Wednesday for some). They have generated some artifacts already and are just finalizing some things (guessing QA) before publishing

<hr>

With the release on route, Sebastian and I decided to go ahead and get these up and start the review process

These PRs were generated with rapids-reviser so contain the same messaging in every case. That said, it is worth noting there are two kinds of CuPy consumers in RAPIDS:

1. Minimal users (simply wrap GPU memory in CuPy arrays and do little else)
2. General users (perform more operations on CuPy arrays and may write algorithms with them)

Sebastian and I decided that it was safe to open ready to review PRs for minimal users and leave PRs in draft for general users. UCX-Py is in the minimal user category

Hope that helps. Please feel free to ask more questions as needed

jameslamb: Got it! Ok thanks very much for that.

I do think as a general rule, the fact that `rapids-reviser` or any other automation was used isn't a strong justification for the PR description being inaccurate. Especially since in RAPIDS repos, the PR description is added automatically to the commit log when you ask the bot to merge. Example: https://github.com/rapidsai/ucx-py/commit/b0cbd6e2540caf620b478b6cc6844079e8bdf2a7

I think it'd be good for you or @seberg to modify this and the other "minimal users" PR descriptions before any of them are merged.

But that's enough context for me to review this and the other "minimal users". Based on what's currently open from the list at https://github.com/rapidsai/build-planning/issues/38#issuecomment-2296690876 and what I see on the NumPy 2 Bringup board, I guess that's just these:

* https://github.com/rapidsai/kvikio/pull/441
* https://github.com/rapidsai/ucxx/pull/271
* this PR

jameslamb (approved):
Makes sense to me, based on @jakirkham 's description in https://github.com/rapidsai/ucx-py/pull/1064#discussion_r1725786361.

I spot-checked a few logs and it looks like NumPy 2.1.0 was getting pulled in. I don't see any other issues.

Please do update the PR description (removing the inaccurate language about CuPy 13.3.0) before merging.

dependencies.yaml
jakirkham: Yeah think that is a good point. Thanks James! 🙏

Yep that sounds right. The IO libraries think of containers of memory on the CPU or the GPU. Not doing math or other operations, which are left to other projects

jakirkham (approved):

jakirkham:
Thanks Sebastian and James! 🙏

jameslamb:
Great, thanks very much @seberg ! I appreciate you both moving this effort forward, I know it's a lot of work.

PR #1069, opened 2024-09-11
Update version error message

### LINKED ISSUE #1068: ImportError raises 1.11 while check is for 1.15
The check is for 1.15 however the error message raised is that 1.11+ is required. 

Bonus ask, it'll be great if we can show the path of the library that's being loaded and checked so that we can debug further if it's system library or environment library. 

https://github.com/rapidsai/ucx-py/blob/2ba56e587a83e5f664dfb6b891fc0fc428d5e16b/ucp/__init__.py#L113-L117

pentschev:
Update version error message to match the minimum support UCX version and provide instructions to check where UCX is being loaded from.

Closes #1068 .

### DIFF
--- a/ucp/__init__.py
+++ b/ucp/__init__.py
@@ -111,7 +111,7 @@ if "UCX_PROTO_ENABLE" not in os.environ and get_ucx_version() >= (1, 12, 0):
 __ucx_min_version__ = "1.15.0"
 __ucx_version__ = "%d.%d.%d" % get_ucx_version()
 
-if get_ucx_version() < tuple([int(i) for i in __ucx_min_version__.split(".")]):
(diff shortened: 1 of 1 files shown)


pentschev:
@madsbk would you mind having a look/approving here as well? This is the same change as you reviewed in https://github.com/rapidsai/ucxx/pull/277 with your suggestion also applied here now via https://github.com/rapidsai/ucx-py/pull/1069/commits/11c637da182dae633f1fb47aba03bc10b209e1fb .

madsbk (approved):
Looks good, thanks @pentschev

pentschev:
Thanks @madsbk !

PR #1074, opened 2024-09-24
Update update-version.sh to use packaging lib

AyodeAwe:
This PR updates the update-version.sh script to use the packaging library, given that setuptools is no longer included by default in Python 3.12.


pentschev:
Maybe we don't need this and can close this PR, it's been already address in `branch-0.41` via https://github.com/rapidsai/ucx-py/pull/1073, plus it fixes the use of `NEXT_RAPIDS_SHORT_TAG_PEP440`.

jameslamb:
Agree with @pentschev , this one could probably be closed. Since `branch-0.41` has already been created, this update doesn't need to make it onto `branch-0.40`.

PR #1080, opened 2024-09-28
Sync some changes from KvikIO's `Array`

jakirkham:
Pulls in a few changes from KvikIO's `Array`. Namely...

* `asarray` for easy conversion to `Array` (no-op if already an `Array`)
* Include types for a few more properties


ucp/_libs/arr.pyx
```diff
@@ -297,3 +297,10 @@ cdef inline Py_ssize_t _nbytes(Py_ssize_t itemsize,
     for i in range(ndim):
         nbytes *= shape_mv[i]
     return nbytes
+
+
+cpdef Array asarray(obj):
```
pentschev, line 302: Similar to @madsbk 's [comment on UCXX](https://github.com/rapidsai/ucxx/pull/290#discussion_r1780497341), a docstring would be nice here too.

jakirkham: Copying over KvikIO's recently added docstring ( https://github.com/rapidsai/kvikio/pull/475#pullrequestreview-2338640849 )

```suggestion
cpdef Array asarray(obj):
    """Coerce other objects to ``Array``. No-op for existing ``Array``s.

    Args:
        obj: Object exposing the Python buffer protocol or ``__cuda_array_interface__``

    Returns:
        Array: An instance of the ``Array`` class
    """
```

ucp/_libs/arr.pyi
```diff
@@ -1,7 +1,12 @@
...
+# Copyright (c) 2020-2024, NVIDIA CORPORATION. All rights reserved.
+# See file LICENSE for terms.
 
-class Array:
-    def __init__(self, obj: object): ...
+from typing import Generic, Tuple, TypeVar
+
+T = TypeVar("T")
+
+class Array(Generic[T]):
```
wence-, line 8: *nit*: This a slightly weird use of a generic type. Usually one would read a type `Array[T]` as "an array containing objects of type `T`", that is for `x : Array[T]`, then `x[i] : T forall i \in [0, len(x))`. However, here you're using it to mean the wrapped type.

ucp/_libs/arr.pyx
```diff
@@ -297,3 +297,18 @@ cdef inline Py_ssize_t _nbytes(Py_ssize_t itemsize,
...
+    """Coerce other objects to ``Array``. No-op for existing ``Array``s.
+
+    Args:
+        obj: Object exposing the Python buffer protocol or ``__cuda_array_interface__``
+
+    Returns:
+        Array: An instance of the ``Array`` class
```
pentschev, line 309:
```suggestion
    Parameters
    ----------
    obj: object
        Object exposing the Python buffer protocol or ``__cuda_array_interface__``.

    Returns
    -------
    array: Array
        An instance of the ``Array`` class.
```

pentschev: UCX-Py has been less consistent in docs formatting than UCXX so I'll leave this only as a suggestion here.

ucp/_libs/arr.pyi
jakirkham: This is taken from [verbatim from KvikIO]( https://github.com/rapidsai/kvikio/blob/2053638989525e18157cb5cad0f4414a6662187a/python/kvikio/kvikio/_lib/arr.pyi#L4-L8 ). Appears this was added in PR ( https://github.com/rapidsai/kvikio/pull/214 ), which it looks like you approved 😉

That said, no objection to changing it to something else. Could you please submit a PR to KvikIO with the syntax you prefer? Can then incorporate it here 🙂

pentschev (approved):
LGTM in spite of @wence-'s comment, that can probably be addressed in KvikIO first and then again incorporated here but it's probably a good idea to let Lawrence voice his (dis-)approval of that too. Thanks @jakirkham .

jakirkham:
Any more thoughts on this one? 🙂

pentschev:
Not from me, I think we can leave it open another 24h, if there are no further comments until then we should merge it.

ucp/_libs/arr.pyi
wence-: Apologies for letting this slip. I don't have a brilliant suggestion, so this is fine and I'll have a think.

wence- (approved):

ucp/_libs/arr.pyi
jakirkham: No worries. Maybe we can raise this as an issue on KvikIO and discuss further there?

jakirkham:
Thank you both! 🙏

PR #1082, opened 2024-10-10
make conda installs in CI stricter

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/106

Proposes specifying the RAPIDS version in `conda install` calls that install CI artifacts, to reduce the risk of CI jobs picking up artifacts from other releases.

### DIFF
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -9,7 +9,7 @@ cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
-RAPIDS_VERSION="$(rapids-version)"
+UCX_PY_VERSION="$(rapids-version)"
 
 rapids-dependency-file-generator \
   --output conda \
@@ -52,7 +52,7 @@ PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
 
 rapids-mamba-retry install \
   --channel "${PYTHON_CHANNEL}" \
-  "ucx-py=${RAPIDS_VERSION}"
+  "ucx-py=${UCX_PY_VERSION}"
 
 rapids-logger "Run tests with conda package"
 run_tests
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -9,7 +9,7 @@ cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
-UCX_PY_VERSION="$(rapids-version)"
+UCX_PY_VERSION="$(head -1 ./VERSION)"
 
 rapids-dependency-file-generator \
   --output conda \
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -9,6 +9,8 @@ cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
+RAPIDS_VERSION="$(rapids-version)"
+
 rapids-dependency-file-generator \
   --output conda \
   --file-key test_python \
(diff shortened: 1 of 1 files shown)


ci/test_python.sh
```diff
@@ -9,6 +9,8 @@ cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
+UCX_PY_VERSION="$(head -1 ./VERSION)"
```
KyleFromNVIDIA, line 12: Why `head` here? `VERSION` is only a single line. I've never used this anywhere else in checking `VERSION`.

jameslamb: Using `head -1` ensures this will do what we want even in the presence of newlines in `VERSION`.

It's used all over the conda build scripts across RAPIDS (also introduced by me, to be fair), e.g.

https://github.com/rapidsai/rmm/blob/90a5631e1093ce44c4feceb88fcf557c3dfc043b/ci/build_python.sh#L23

https://github.com/rapidsai/cudf/blob/7173b52fce25937bb69e22a083a5de4655078fa1/ci/build_python.sh#L26-L53

https://github.com/rapidsai/cuml/blob/22342aad0ecd327e3140471c479d01d83b932c23/ci/build_python.sh#L24

KyleFromNVIDIA: This should really be `rapids-version` or `rapids-version-major-minor`. If you want to adjust those to use `head`, we can do that, but let's unify our reading of `VERSION` by using the tools at `gha-tools`.

jameslamb: I've only used `head` here in `ucx-py` and `ucxx` for this purpose, because their `VERSION` files have a single-digit patch component (`.0`)

https://github.com/rapidsai/ucx-py/blob/2be910850af81cb3e1109252a0537da1be7b20b7/VERSION#L1

And `rapids-version` / `rapids-version-major-minor` explicitly look for a 2-digit patch component (that final `[0-9]{2}`).

```shell
readonly regex="^([0-9]{2})\.([0-9]{2})\.([0-9]{2}).*$"
```

([code link](https://github.com/rapidsai/gha-tools/blob/8bd8fca71b5fae38b1493c547d15e73da40b32e1/tools/rapids-version#L11))

Based on your experience from doing https://github.com/rapidsai/build-planning/issues/15, would it be safe to change `rapids-version` to something like this to account for that?

```shell
readonly regex="^([0-9]{2})\.([0-9]{2})\.([0-9]{1, 2}).*$"
```

If so I'm happy to make that change. I was just nervous about touching something so central as part of this.

jakirkham: Please note that UCX versioning diverges from the rest of RAPIDS

jameslamb: *(edited after posting my original comment, realized the major component is also 1 digit, `0.`)*

KyleFromNVIDIA: Ah - I'd forgotten that this repo has different versioning from the rest of RAPIDS. In that case I'm happy to use `tail` instead.

Anything that uses RAPIDS calver versioning should get `rapids-version` and `rapids-version-major-minor`.

jameslamb: Ah great, thanks! This is making me realize that those other cases I mentioned probably should be using `rapids-version` instead of `rapids-version-major-minor`. That should be easily `rapids-reviser`-able.

> In that case I'm happy to use `tail` instead.

Any reason to prefer `tail` to `head`? I was thinking that extra trailing newlines are more likely than empty lines at the top of the file.

KyleFromNVIDIA: > Any reason to prefer `tail` to `head`?

Because I had a brain fart and mixed up my commands. Sorry, I meant `head`.

jameslamb: ha ok no prob! Thanks for talking through it with me.

And thanks for all that prior work you did on reducing the proliferation of hard-coded versions in these repos. That's been really really helpful.

jakirkham: > Ah - I'd forgotten that this repo has different versioning from the rest of RAPIDS.

No worries. It has confused others before

If there is a good way to bake this into standard tooling to avoid the special casing here (or simplify it), that would be great!

KyleFromNVIDIA (approved):

PR #1085, opened 2024-10-21
Mark all `nogil` `cdef` functions as `noexcept`

jakirkham:
These are all pure C functions implemented in Cython. They do not raise. Cython 3 adds checks for exceptions in these functions, which is unnecessary given none of these set exceptions. So add `noexcept` to turn off these Cython checks.

xref: https://github.com/rapidsai/kvikio/pull/502


pentschev (approved):
LGTM, thanks @jakirkham !

jakirkham:
Thanks Peter! 🙏

PR #1087, opened 2024-10-22
Use Cython's `array` to back `Py_ssize_t[::1]`

jakirkham:
In `Array`, `Py_ssize_t[::1]` objects are currently backed by [CPython `array`'s]( https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#cpython-array-module ) with some internal bits expressed in Cython. However these are not compatible with [Python's Limited API and Stable ABI]( https://docs.python.org/3/c-api/stable.html#c-api-stability ). To address that, switch to [Cython's own `array` type]( https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#cython-arrays ). As this is baked into Cython and doesn't use anything special, it is compatible with Python's Limited API and Stable ABI.

xref: https://github.com/rapidsai/build-planning/issues/42

### DIFF
--- a/ucp/_libs/arr.pyx
+++ b/ucp/_libs/arr.pyx
@@ -4,13 +4,12 @@
 # cython: language_level=3
 
 
-from cpython.array cimport array, newarrayobject
 from cpython.buffer cimport PyBuffer_IsContiguous
+from cpython.mem cimport PyMem_Free, PyMem_Malloc
 from cpython.memoryview cimport (
     PyMemoryView_FromObject,
     PyMemoryView_GET_BUFFER,
 )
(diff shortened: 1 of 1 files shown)


pentschev:
I tried to do some benchmark with UCX-Py but the variability prevents us from reaching any conclusive results regarding performance before and after this change, where sometimes results are better before and other times after, therefore I think it doesn't impose any negative impact that would be noticeable in practical use cases.

pentschev (approved):
LGTM, thanks @jakirkham .

jakirkham:
Thanks Peter! 🙏

PR #1089, opened 2024-10-29
remove unnecessary sccache configuration, other small CI changes

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/111

Proposes some small packaging/CI changes, matching similar changes being made across RAPIDS.

* updating to the latest `rapids-dependency-file-generator` (v1.16.0)
* removing unnecessary calls to `rapids-configure-sccache` (this project does not use `sccache`)

## Notes for Reviewers

I'd originally started this PR with the goal of printing `sccache` stats in builds here... but realized this project does not use `sccache`.

I chose not to pursue adding `sccache` here. The fact that this project doesn't use CMake means it'd take some effort to figure out how to inject `sccache` into the compilation (`CC` / `CXX` environment variables? something else?). Conda and wheel builds are only spending around 20 seconds actually compiling Cython code, so it doesn't seem worth the effort.

### DIFF
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -5,8 +5,6 @@ set -euo pipefail
 
 rapids-configure-conda-channels
 
-source rapids-configure-sccache
-
 source rapids-date-string
 
 rapids-print-env
@@ -16,11 +14,7 @@ rapids-generate-version > ./VERSION
 rapids-logger "Begin py build"
 conda config --set path_conflict prevent
 
-sccache --show-zero-stats
-
 RAPIDS_PACKAGE_VERSION=$(head -1 ./VERSION) rapids-conda-retry mambabuild \
   conda/recipes/ucx-py
-
-sccache --show-adv-stats
 
 rapids-upload-conda-to-s3 python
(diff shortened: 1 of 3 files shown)


ci/build_python.sh
```diff
@@ -16,7 +16,11 @@ rapids-generate-version > ./VERSION
 rapids-logger "Begin py build"
 conda config --set path_conflict prevent
 
+sccache --show-zero-stats
```
jameslamb, line 19:
```suggestion
sccache --zero-stats
```

🤦🏻

msarahan (approved):

ci/build_wheel.sh
```diff
@@ -25,4 +27,4 @@ python -m auditwheel repair \
     --exclude "libuct.so.0" \
     dist/*
 
-RAPIDS_PY_WHEEL_NAME="${underscore_package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 final_dist
+RAPIDS_PY_WHEEL_NAME="ucx_py_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 final_dist
```
jameslamb, line 30:
```suggestion
RAPIDS_PY_WHEEL_NAME="ucx_py_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 python final_dist
```

grrr I missed this one

PR #1093, opened 2024-11-15
Set PyNVML upper pin

pentschev:
The plan is for `pynvml` to become a metapackage in version 12, where it not anymore ship NVML bindings but instead install `nvidia-ml-py` to serve the official NVML bindings. To provide a better future support, setting an upper pin for `pynvml` is a safer choice.


jakirkham (approved):
Thanks Peter! 🙏

PR #1099, opened 2024-12-13
document changes in library-loading, update build-UCX-from-source docs

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/118

Caused by https://github.com/rapidsai/ucx-wheels/pull/13

I originally came here to document the implications of https://github.com/rapidsai/ucx-wheels/pull/13 in the docs, namely:

* if you have a `libucx-cu{11,12}` wheel installed, then by default `ucx-py` will use UCX libraries from that wheel
* environment variable `RAPIDS_LIBUCX_PREFER_SYTEM_LIBRARY=true` can be set to opt out of this and use a system installation instead

While doing that, I noticed some other opportunities for improvement in the installation docs:

* updating build-UCX-from-source instructions to UCX 1.15 ([the oldest version this project now supports](https://github.com/rapidsai/ucx-py/blob/9efacc6069226de8e207177a359189f8880203a8/dependencies.yaml#L159))
* clarifying and simplifying some language

### How I tested this

Followed these instructions in a Docker container running on an x86_64 machine with 8 V100s.

```shell
docker run \
    --rm \
    --gpus 0 \
    -v $(pwd):/opt/work \
    -w /opt/work \
    -it rapidsai/ci-conda:latest \
    bash
```

Used `conda` to set up the build environment:

```shell
conda create -n ucx -c conda-forge \
    automake make libtool pkg-config \
    "python=3.12" "setuptools>=64.0" "cython>=3.0.0" \
    cuda-nvcc \
    cuda-cudart-dev \
    cuda-nvml-dev \
    cuda-nvtx-dev \
    cuda-version=12.5
```

Ran variations of this code snippet to test my install:

```shell
python -c "import ucp; print(ucp.get_ucx_version())"
```

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -235,10 +235,11 @@ To build UCX-Py using those UCX packages (to avoid needing to build UCX from sou
 This will automatically handle installing appropriate, compatible ``libucx-cu{11,12}`` packages for build-time and runtime use.
 When you run UCX-Py code installed this way, it will load UCX libraries from the installed ``libucx-cu{11,12}`` package.
 
-UCX-Py packages are built against the oldest version of UCX the project supports, and can run against a range
+UCX-Py packages are built against the oldest version of UCX that UCX-Py supports, and can run against a range
 of ABI-compatible UCX versions.
 
-You can use packages from PyPI to customize the UCX version used at runtime. For example, to use UCX 1.16 at runtime, run the following.
+You can use packages from PyPI to customize the UCX version used at runtime.
+For example, to switch to using UCX 1.16 at runtime, run the following.
 
 ::
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -96,9 +96,8 @@ Build Dependencies
         "python=3.12" "setuptools>=64.0" "cython>=3.0.0"
 
 .. note::
-    The Python version must be explicitly specified here.
-    See https://docs.rapids.ai/install/ for the list of supported Python versions
-    for recent RAPIDS releases.
+    The Python version must be explicitly specified here, UCX-Py currently supports
+    Python versions 3.10, 3.11, and 3.12.
 
 Test Dependencies
 ^^^^^^^^^^^^^^^^^
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -126,7 +126,7 @@ Instructions for building UCX >= 1.15.0 (minimum version supported by UCX-Py), m
     mkdir build
     cd build
     # Performance build
-    ../contrib/configure-release --prefix=$CONDA_PREFIX --with-cuda=/opt/conda/pkgs/cuda-nvml-dev-12.5.82-he02047a_0 --enable-mt
+    ../contrib/configure-release --prefix=$CONDA_PREFIX --with-cuda=$CUDA_HOME --enable-mt
     # Debug build
     ../contrib/configure-devel --prefix=$CONDA_PREFIX --with-cuda=$CUDA_HOME --enable-mt
     make -j install
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -34,17 +34,24 @@ support is active by checking for the presence of ``/dev/infiniband/rdma_cm`` an
 Conda
 -----
 
-Conda packages can be installed as so. Replace ``<CUDA version>`` with the
-desired version (minimum ``11.2``). These are available both on ``rapidsai``
-and ``rapidsai-nightly``. Starting with the UCX 1.14.1 conda-forge package,
-InfiniBand support is available again via rdma-core, thus building UCX
-from source is not required solely for that purpose anymore but may still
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -85,11 +93,12 @@ Build Dependencies
...
         automake make libtool pkg-config \
-        "python=3.12" setuptools "cython>=3.0.0"
+        "python=3.12" "setuptools>=64.0" "cython>=3.0.0"
 
 .. note::
-    The Python version must be explicitly specified here, UCX-Py currently supports
-    Python versions 3.10, 3.11, and 3.12.
+    The Python version must be explicitly specified here.
+    See https://docs.rapids.ai/install/ for the list of supported Python versions
+    for recent RAPIDS releases.
```
pentschev, line 101: I think this is not the most user-friendly way to phrase this here, unfortunately we do not have UCX-Py explicitly documented in https://docs.rapids.ai/install/ which has a different versioning scheme than the rest of RAPIDS, and thus it would be difficult for someone to understand what Python version is supported for UCX-Py based solely on the RAPIDS version. I agree the way it was documented isn't optimal and I can't think of a better way to do this differently than what it was either.

docs/source/install.rst
```diff
@@ -102,17 +111,17 @@ Test Dependencies
         dask distributed cloudpickle
 
 
-UCX >= 1.11.1
+UCX >= 1.15.0
```
pentschev, line 114: You are right we don't officially support UCX < 1.15.0, nevertheless UCX-Py should still work with UCX >= 1.11.1, so the versions were left for the benefit of users who may still need older versions. In any case, I think it's fine to keep this bump in the docs too.

docs/source/install.rst
```diff
@@ -205,7 +214,40 @@ PyPI
 
 The following instructions assume you'll be installing UCX-Py on a CUDA-enabled system, in a pip-only environment.
 
-Installing UCX-Py from source in a pip-only environment has additional limitations when compared to conda environments. Unlike conda packages, where the ``ucx`` package is installed under the ``CONDA_PREFIX``, ``libucx`` is installed under ``site-packages`` which is normally not looked for system libraries. Therefore, you will either need UCX to be installed in the system path, or include the UCX install path in ``LD_LIBRARY_PATH``.
+Installing UCX-Py from source in a pip-only environment has additional limitations when compared to conda environments.
+
+UCX-Py with UCX from PyPI
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+CUDA-enabled builds of the UCX libraries are available from PyPI, under the names ``libucx-cu{11,12}``.
```
pentschev, line 222:
```suggestion
CUDA-enabled builds of the UCX libraries are available from PyPI, under the names ``libucx-cu{11,12}``. 
Notice that those builds do not currently include InfiniBand support, if InfiniBand is required you will 
need to provide a custom UCX install as described in the "UCX-Py with custom UCX install" section.
```

pentschev (commented):
Thanks @jameslamb I've left a few comments, overall looks good.

docs/source/install.rst
jameslamb: Sure, that makes sense. My main motivation here was to avoid a hard-coded list of versions, since duplicating that kind of information in multiple places makes it more likely that it'll become out of date.

That said.... we do have a `rapids-reviser` template set up to replace this line when we roll out add-a-new-Python-version updates to all of RAPIDS.

I'll revert this change based on your comments here.

jameslamb, line 101:
```suggestion
    The Python version must be explicitly specified here, UCX-Py currently supports
    Python versions 3.10, 3.11, and 3.12.
```

docs/source/install.rst
jameslamb: Oh I didn't realize that was intentional. Happy to revert this if you want.

docs/source/install.rst
pentschev: I agree that would be ideal, unfortunately I think it's not helpful for the users given the different branching.

How does the `rapids-reviser` work? Can we easily make it replace arbitrary code/docs lines?

docs/source/install.rst
pentschev: No, it's ok to leave it like that. UCX < 1.15.0 is ancient anyway.

docs/source/install.rst
jameslamb: Alright, no prob! I've reverted that change here.

> Can we easily make it replace arbitrary code/docs lines?

Yes totally! You give it a shell script and say "run this on these repos". This type of case is easy to automatically update there with `sed`, not a problem.

docs/source/install.rst
jameslamb: Ha ok, thank you, sounds good.

pentschev (approved):
LGTM, thanks @jameslamb .

jameslamb:
Thanks @pentschev for continuing to work with me and @vyasr on these patterns for how these libraries find UCX 😁

PR #1101, opened 2025-01-07
Create Conda test environment in one go

jakirkham:
Partially addresses issue: https://github.com/rapidsai/build-planning/issues/22

Instead of creating the Conda test environment and then installing `ucx-py` into it, just create the Conda environment with the `ucx-py` package to start with. Thus the environment is constructed only once with everything it needs before testing begins.


dependencies.yaml
```diff
@@ -212,12 +213,17 @@ dependencies:
...
+      - output_types: conda
+        packages:
+          - ucx-py==0.42.*,>=0.0.0a0
   test_python:
     common:
       - output_types: [conda, requirements, pyproject]
         packages:
           - cloudpickle
-          - dask
+          - rapids-dask-dependency==25.2.*,>=0.0.0a0
```
jakirkham, line 226: Switch from `dask` to `rapids-dask-dependency` for Dask test dependency. That way it matches however we have Dask constrained elsewhere in RAPIDS

jakirkham (commented):
Added a few notes below to help contextualize changes

msarahan (approved):

pentschev (approved):
Makes sense and LGTM. Thanks @jakirkham !

PR #1103, opened 2025-01-21
Expose configuration of exchange peer info timeout

pentschev:
Add support to configure the timeout of messages during exchange peer info via constructor argument and environment variable.

This feature is important in Distributed (see https://github.com/rapidsai/ucx-py/pull/994 for details), more so on larger scale clusters. A new configuration variable will be added to allow controlling this feature via Dask config and environment variables.

### DIFF
--- a/ucp/core.py
+++ b/ucp/core.py
@@ -35,7 +35,9 @@ def _get_ctx():
     return _ctx
 
 
-async def exchange_peer_info(endpoint, msg_tag, ctrl_tag, listener, stream_timeout=5.0):
+async def exchange_peer_info(
+    endpoint, msg_tag, ctrl_tag, listener, connect_timeout=5.0
+):
     """Help function that exchange endpoint information"""
 
     # Pack peer information incl. a checksum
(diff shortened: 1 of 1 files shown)


ucp/core.py
```diff
@@ -230,6 +235,13 @@ def __init__(self, config_dict={}, blocking_progress_mode=None):
         else:
             self.blocking_progress_mode = True
 
+        if "UCXPY_CONNECT_TIMEOUT" in os.environ:
+            self.connect_timeout = float(os.environ["UCXPY_CONNECT_TIMEOUT"])
+        elif connect_timeout is not None:
+            self.connect_timeout = connect_timeout
+        else:
+            self.connect_timeout = 5.0
+
```
madsbk, line 244: `connect_timeout` takes precedence right?
```suggestion
        if connect_timeout is None:
            self.connect_timeout = float(os.get("UCXPY_CONNECT_TIMEOUT", 5))           
        else:
            self.connect_timeout = connect_timeout
```

madsbk (approved):
LGTM, I only have a minor suggestion

ucp/core.py
madsbk: I am not sure, I think a non-default `connect_timeout` argument should take precedence. Then environment variable and finally the default value.

At least, I would find it surprising if the environment variable takes precedence over an explicit argument.

madsbk: Alternatively, I have seen projects using an extra _forcing_ environment variable like: `UCXPY_CONNECT_TIMEOUT_FORCE`, which would override everything else.

pentschev: Ok, I think the force/override may be a good alternative. That is not super critical right now though and I can't spend too much time on it at the moment, so I'll just apply your suggestion and revisit this idea later.

pentschev:
Thanks @madsbk !

PR #1105, opened 2025-01-22
Increase upper UCX pin to 1.19

pentschev:
Increase upper UCX pin to 1.19 to allow us supporting UCX 1.18 that is needed for RAPIDS 25.02.

### DIFF
--- a/conda/recipes/ucx-py/meta.yaml
+++ b/conda/recipes/ucx-py/meta.yaml
@@ -39,7 +39,7 @@ requirements:
     {% endfor %}
   run:
     - python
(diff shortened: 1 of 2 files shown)


jameslamb (changes requested):
Blocking this for the reasons described in https://github.com/rapidsai/ucxx/pull/352#pullrequestreview-2567984747 ... there aren't yet `libucx` 1.18 wheels, and we should wait to see CI pass using UCX 1.18 before merging.

jameslamb:
I just cancelled CI here to save CI resources... will restart it once we have `libucx` 1.18 wheels published.

PR #1145, opened 2025-06-13
add shellcheck to pre-commit

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/135

Adds `shellcheck` to `pre-commit`, to catch issues like unsafe access patterns, unused variables, etc. in shell scripts.

Fixes these:

```text
SC2034 (warning): UCX_PY_VERSION appears unused. Verify use (or export if used externally).
SC2034 (warning): CURRENT_SHORT_TAG appears unused. Verify use (or export if used externally).
SC2046 (warning): Quote this to prevent word splitting.
SC2068 (error): Double quote array expansions to avoid re-splitting elements.
SC2086 (info): Double quote to prevent globbing and word splitting.
SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.
SC2145 (error): Argument mixes string and array. Use * or separate argument
```

Also updates other RAPIDS-specific pre-commit hooks to their latest versions.

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+# Copyright (c) 2022-2025, NVIDIA CORPORATION.
+
 ########################
 # ucx-py Version Updater #
 ########################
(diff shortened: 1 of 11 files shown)


docker/bench-all.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION.
 
 set -e
 
 function logger {
-    echo -e "\n$@\n"
+    echo -e "\n${1}\n"
```
jameslamb, line 7: Here's what `shellcheck` was unhappy about:

```text
In docker/run.sh line 6:
    echo -e "\n$@\n"
               ^-- SC2145 (error): Argument mixes string and array. Use * or separate argument
```

There aren't any uses of this function in this script where more than 1 positional argument is passed, so I think just changing it to `${1}` should be fine.

docker/bench-all.sh
```diff
@@ -13,8 +13,15 @@ source /opt/conda/etc/profile.d/conda.sh
...
+
 # Benchmark using command-line provided transports or else default
-for tls in ${@:-"tcp" "all"}; do
+if [[ ${#@} -eq 0 ]]; then
+    transport_types=(tcp all)
+else
+    transport_types=("${@}")
+fi
+
+for tls in "${transport_types[@]}"; do
```
jameslamb, line 24: Here's what `shellcheck` was unhappy about:

```text
In docker/bench-all.sh line 17:
for tls in ${@:-"tcp" "all"}; do
           ^---------------^ SC2068 (error): Double quote array expansions to avoid re-splitting elements.
```

Now I think it's highly unlikely that you'd intentionally pass any value with spaces here, so if you'd prefer I'd be happy to revert this change and just stick a `# shellcheck disable` comment on it.

gforsyth: This seems like a reasonable change to me

ci/release/update-version.sh
```diff
@@ -13,18 +15,14 @@ NEXT_FULL_TAG=$1
 
 # Get current version
 CURRENT_TAG=$(git tag | grep -xE 'v[0-9\.]+' | sort --version-sort | tail -n 1 | tr -d 'v')
-CURRENT_MAJOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[1]}')
-CURRENT_MINOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[2]}')
-CURRENT_PATCH=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[3]}')
-CURRENT_SHORT_TAG=${CURRENT_MAJOR}.${CURRENT_MINOR}
```
gforsyth, line 19: :tada:

gforsyth (approved):
This looks good!  :shipit:

docker/bench-all.sh
jameslamb: Thanks! I'd like to give @pentschev a chance to review this particular change too... want to be sure he's comfortable with it as well.

pentschev: Yes, they are intended to be separate arguments, not arguments with spaces in it.

pentschev (approved):
The scripts under `docker/` are not regularly tested and probably haven't been used in a couple years, so it's possible they're just entirely broken. I need to revive them but have no bandwidth now, with that said, the changes are ok even if they end up breaking something.

Thanks @jameslamb !

jameslamb:
Alright sounds good, thanks @pentschev !

PR #1149, opened 2025-06-27
Use CUDA 12.9 in Conda, Devcontainers, Spark, GHA, etc.

jakirkham:
Addressing issues:
* https://github.com/rapidsai/build-planning/issues/195
* https://github.com/rapidsai/build-planning/issues/173

## Description

Use CUDA 12.9 throughout different build and test environments.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -46,7 +46,7 @@ Change `cuda-version` to pin to a different CUDA minor version if you'd like.
 
     # CUDA 12
     conda create -n ucx -c conda-forge -c rapidsai \
-      cuda-version=12.8 ucx-py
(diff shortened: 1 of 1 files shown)


jakirkham:
Looks like this is unneeded. Closing...

jakirkham:
Only the docs needed an update. So have pushed that change here

pentschev (approved):
For reference, no other changes are needed because both CUDA 12.8 and 12.9 are [tested in CI](https://github.com/rapidsai/ucx-py/blob/b9858695a86dc3eaf259019a45b53eae27b5933a/dependencies.yaml#L86-L93).

LGTM. Thanks John!

PR #1156, opened 2025-07-18
chore: bump to 0.46

gforsyth:


pentschev (approved):
Thanks @gforsyth .

bdice (approved):
The branch was created already -- did we run the burndown scripts correctly? Normally `update-version.sh` is run as a part of branch creation.

cc: @AyodeAwe

gforsyth:
> The branch was created already -- did we run the burndown scripts correctly? Normally `update-version.sh` is run as a part of branch creation.

That's on me, I think `ucx-py` wasn't included in burndown because of the pending removal, so I created the branch off of 0.45 then PR'd this in with the changes.

gforsyth:
Crap, muscle memory and I hit "merge" instead of typing `/merge`.  Should be ok, just a FYSA

Issue #1160, opened 2025-08-12
UCX-Py archival


pentschev:
With RAPIDS 25.08 release last week, that is expected to have marked the final release of UCX-Py, v0.45. The 25.08 release has transparently switched to UCXX in Dask-CUDA, and while we don't expect any major issues we're still building UCX-Py should something critical show up. Provided no critical reports show up in the next two weeks, the plan is to begin with the complete removal UCX-Py as a dependency to all RAPIDS projects, followed by disabling CI, publishing packages and ultimately archiving this repository. The plan is to start on August 25th and it is expected to take up to one or two weeks to complete everything.

Tasks to complete UCX-Py archival are listed below, in order which they should be completed to avoid breaking other projects:

- [x] Remove from [cuml](https://github.com/rapidsai/cuml): https://github.com/rapidsai/cuml/pull/7130
- [x] Remove from [cugraph](https://github.com/rapidsai/cugraph): https://github.com/rapidsai/cugraph/pull/5238
- [x] Remove from [raft](https://github.com/rapidsai/raft): https://github.com/rapidsai/raft/pull/2791
- [x] Remove from [dask-cuda](https://github.com/rapidsai/dask-cuda): https://github.com/rapidsai/dask-cuda/issues/1517, https://github.com/rapidsai/dask-cuda/pull/1546, https://github.com/rapidsai/dask-cuda/pull/1548
- [x] Remove from [distributed](https://github.com/dask/distributed): https://github.com/dask/distributed/pull/9105
- [x] Add configuration schema to [distributed-ucxx](https://github.com/rapidsai/ucxx/tree/branch-0.46/python/distributed-ucxx): https://github.com/rapidsai/ucxx/pull/497
- [x] Register `protocol="ucx"` in [distributed-ucxx](https://github.com/rapidsai/ucxx/tree/branch-0.46/python/distributed-ucxx): https://github.com/rapidsai/ucxx/pull/504
- [x] Move and update existing UCX-Py docs into UCXX: https://github.com/rapidsai/ucxx/pull/510
  - [x] Update Dask-CUDA docs to mention/point to UCXX docs: https://github.com/rapidsai/dask-cuda/pull/1554
- [x] Remove from [rapids-dask-dependency](https://github.com/rapidsai/rapids-dask-dependency): https://github.com/rapidsai/rapids-dask-dependency/pull/120
- [x] UCX-Py archival
  - [x] Disable CI
  - [x] Disable publishing packages
  - [x] Archive repository

PR #1162, opened 2025-08-20
Build and test with CUDA 13.0.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test
* moves some dependency pins:
  - `cupy`: `>=13.6.0`

Contributes to https://github.com/rapidsai/build-planning/issues/68

* updates to CUDA 13 dependencies in fallback entries in `dependencies.yaml` matrices (i.e., the ones that get written to `pyproject.toml` in source control)

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.


jameslamb:
Ok, 2 things we need to resolve.

## Problem 1: building wheels against `libucx-cu13==1.19.0` headers fails

```text
  Compiling ucp/_libs/ucx_api.pyx because it changed.
  [1/1] Cythonizing ucp/_libs/ucx_api.pyx
  building 'ucp._libs.ucx_api' extension
  creating build/temp.linux-aarch64-cpython-313/ucp/_libs/src
  gcc -pthread -fno-strict-overflow -Wsign-compare -DNDEBUG -g -O3 -Wall -fPIC -I/pyenv/versions/3.13.6/include -I/tmp/pip-build-env-jzi5ae7w/normal/lib/python3.13/site-packages/libucx/include -I/pyenv/versions/3.13.6/include/python3.13 -c ucp/_libs/src/c_util.c -o build/temp.linux-aarch64-cpython-313/ucp/_libs/src/c_util.o -std=c99 -Werror
  In file included from /tmp/pip-build-env-jzi5ae7w/normal/lib/python3.13/site-packages/libucx/include/ucs/memory/memory_type.h:12,
                   from /tmp/pip-build-env-jzi5ae7w/normal/lib/python3.13/site-packages/libucx/include/ucp/api/ucp_def.h:14,
                   from /tmp/pip-build-env-jzi5ae7w/normal/lib/python3.13/site-packages/libucx/include/ucp/api/ucp.h:12,
                   from ucp/_libs/src/c_util.h:8,
                   from ucp/_libs/src/c_util.c:6:
  /tmp/pip-build-env-jzi5ae7w/normal/lib/python3.13/site-packages/libucx/include/ucs/sys/compiler_def.h:81:50: error: operand of ‘?:’ changes signedness from ‘int’ to ‘long unsigned int’ due to unsignedness of other operand [-Werror=sign-compare]
     81 | #define UCS_MASK(_i)             (((_i) >= 64) ? ~0 : (UCS_BIT(_i) - 1))
        |                                                  ^~
  /tmp/pip-build-env-jzi5ae7w/normal/lib/python3.13/site-packages/libucx/include/ucp/api/ucp.h:511:31: note: in expansion of macro ‘UCS_MASK’
    511 |     UCP_DATATYPE_CLASS_MASK = UCS_MASK(UCP_DATATYPE_SHIFT) /**< Data-type class
        |                               ^~~~~~~~
  cc1: all warnings being treated as errors
  error: command '/opt/rh/gcc-toolset-14/root/usr/bin/gcc' failed with exit code 1
  error: subprocess-exited-with-error
  
  × Building wheel for ucx-py-cu13 (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
```

([wheel-build link](https://github.com/rapidsai/ucx-py/actions/runs/17107231950/job/48519361574?pr=1162#step:12:332))

## Problem 2: tests require CUDA 13 `cudf` packages, which don't exist yet

```text
error    libmamba Could not solve for environment specs
    The following packages are incompatible
    ├─ cuda-version =13.0 * is requested and can be installed;
    └─ cudf =25.10,>=0.0.0a0 * is not installable because it requires
       └─ cuda-version >=12,<13.0a0 *, which conflicts with any installable versions previously reported.
critical libmamba Could not solve for environment specs
```

([conda-python-tests link](https://github.com/rapidsai/ucx-py/actions/runs/17107231950/job/48519692132?pr=1162#step:10:67))

pentschev:
I think this comes up from a place of need, but could we skip CUDA 13 for UCX-Py? The plan is to start archival of it next Monday, see https://github.com/rapidsai/ucx-py/issues/1160, which I presume will take approximately a week or two at the most, so the timing is a little bit off here. Anyway, I want us to explore if there's a way we can move forward with CUDA 13 support without UCX-Py support, since this work here is likely gonna be removed in about a week.

jameslamb:
> _explore if there's a way we can move forward with CUDA 13 support without UCX-Py support_

If we're not going to release `ucx-py` 0.46 packages, we don't technically "need" to support CUDA 13 here.

However we DO need to get CUDA 13 support shipped in RAPIDS 25.10 (cc @robertmaynard). There are 18 (U.S.) working days left until burndown begins for the 25.10 release: https://docs.rapids.ai/maintainers/

If we wait for `ucx-py` to be factored out of RAPIDS and if it takes 1-2 weeks and doesn't start until Monday, that wouldn't leave much time to work through these CUDA 13 updates across the rest of RAPIDS and discover other issues like dependencies that haven't updated yet.

I tried to roughly map out the dependencies in https://github.com/rapidsai/build-planning/issues/208

With that plus the list you provided in https://github.com/rapidsai/ucx-py/issues/1160, I think we're going to get stuck pretty soon without `ucx-py` CUDA 13 packages... because `ucx-py` is a dependency (directly or indirectly) of most of RAPIDS. 

For example.... `raft-dask` has a hard runtime dependency on `ucx-py` and not having CUDA 13 packages for all of the packages from the `raft` repo blocks `cuvs`, `cuml`, `cugraph`, and `cugraph-gnn`.

Here are some options I see:

* get `ucx-py` CUDA 13 support working as soon as possible
* try to update the paths through the dependency graph that don't need `ucx-py`
  - e.g., only C++ builds / tests?
  - e.g., could we update all of RAFT **except** `raft-dask`?

Trying to work around `ucx-py` like that would be a lot more effort than otherwise going through normal build/packaging updates... but I don't know how much, and I don't know how to compare it to the amount of effort it'd take to get `ucx-py` building with CUDA 13.

jameslamb:
I've also started an internal chat thread about this, if you'd rather talk through the options there.

bdice:
@jameslamb I recommend skipping/disabling the tests that require cuDF for now.

```
error    libmamba Could not solve for environment specs
    The following packages are incompatible
    ├─ cuda-version =13.0 * is requested and can be installed;
    └─ cudf =25.10,>=0.0.0a0 * is not installable because it requires
       └─ cuda-version >=12,<13.0a0 *, which conflicts with any installable versions previously reported.
```

jameslamb:
Just pushed some changes I think will help. Just ignoring that one compiler warning from https://github.com/rapidsai/ucx-py/pull/1162#issuecomment-3207678724, I was able to build and run most tests locally like this:

```shell
docker run \
    --rm \
    -v $(pwd):/opt/work \
    -w /opt/work \
    -it rapidsai/ci-wheel:25.10-cuda13.0.0-rockylinux8-py3.13 \
    ./ci/build_wheel.sh

# NOTE: with test_wheel.sh changed to install locally-build wheel in dist/ instead of downloading it
docker run \
    --rm \
    --gpus all \
    -v $(pwd):/opt/work \
    -w /opt/work \
    -it rapidsai/ci-wheel:25.10-cuda13.0.0-rockylinux8-py3.13 \
    ./ci/test_wheel.sh
```

Just a few `cupy` errors because the driver on my local workstation is too old for CUDA 13. So hopefully those changes will be all we need, let's see 😅

jameslamb:
At this point everyone involved is subscribed to notifications anyway via comments, so moved this out of draft.

jameslamb:
Some CUDA 13 tests are segfaulting like this:

```text
Current thread 0x000073d0c2768b80 (most recent call first):
  File "/pyenv/versions/3.10.18/lib/python3.10/site-packages/numba/cuda/cudadrv/driver.py", line 326 in safe_cuda_api_call
  File "/pyenv/versions/3.10.18/lib/python3.10/site-packages/numba/cuda/cudadrv/driver.py", line 505 in __enter__
  File "/pyenv/versions/3.10.18/lib/python3.10/site-packages/numba/cuda/cudadrv/devices.py", line 121 in ensure_context
  File "/pyenv/versions/3.10.18/lib/python3.10/contextlib.py", line 135 in __enter__
  File "/pyenv/versions/3.10.18/lib/python3.10/site-packages/numba/cuda/cudadrv/devices.py", line 231 in _require_cuda_context
...
==== backtrace (tid:    727) ====
 0  /pyenv/versions/3.10.18/lib/python3.10/site-packages/libucx/lib/libucs.so(ucs_handle_error+0x294) [0x73d0c0fb7b14]
 1  /pyenv/versions/3.10.18/lib/python3.10/site-packages/libucx/lib/libucs.so(+0x34cca) [0x73d0c0fb7cca]
 2  /pyenv/versions/3.10.18/lib/python3.10/site-packages/libucx/lib/libucs.so(+0x34f7e) [0x73d0c0fb7f7e]
 3  /usr/lib/x86_64-linux-gnu/libc.so.6(+0x45330) [0x73d0c2898330]
 4  /usr/lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0x11c) [0x73d0c28f1b2c]
 5  /usr/lib/x86_64-linux-gnu/libc.so.6(gsignal+0x1e) [0x73d0c289827e]
```

([build link](https://github.com/rapidsai/ucx-py/actions/runs/17131005285/job/48595574082?pr=1162#step:11:357))

Seeing `numba.code` in the stack trace makes me think... maybe we need to add an explicit `numba-cuda` dependency here, similar to https://github.com/rapidsai/dask-cuda/pull/1531

docs/source/install.rst
```diff
@@ -40,6 +40,10 @@ Change `cuda-version` to pin to a different CUDA minor version if you'd like.
 
 ::
 
+    # CUDA 13
+    conda create -n ucx -c conda-forge -c rapidsai \
+      cuda-version=13.0 ucx-py
+
     # CUDA 12
     conda create -n ucx -c conda-forge -c rapidsai \
       cuda-version=12.9 ucx-py
```
pentschev, line 49:
```suggestion
    # CUDA 12
    conda create -n ucx -c conda-forge -c rapidsai \
      cuda-version=12.9 ucx-py

    # CUDA 13
    conda create -n ucx -c conda-forge -c rapidsai \
      cuda-version=13.0 ucx-py
```

pentschev: Just to keep consistent ordering with text and keep chronological ordering.

docs/source/install.rst
```diff
@@ -52,15 +56,18 @@ be done if desired (e.g., to test for new capabilities or bug fixes).
...
 installed instead. CUDA versions are differentiated by the suffix
 ``-cuXY``, where ``XY`` must be replaced with the desired CUDA version.
 
 ::
 
+    # CUDA 13
+    pip install ucx-py-cu13
+
     # CUDA 12
     pip install ucx-py-cu12
```
pentschev, line 72:
```suggestion
    # CUDA 12
    pip install ucx-py-cu12

    # CUDA 13
    pip install ucx-py-cu13
```

docs/source/install.rst
```diff
@@ -225,19 +232,22 @@ To build UCX-Py using those UCX packages (to avoid needing to build UCX from sou
...
 You can use packages from PyPI to customize the UCX version used at runtime.
-For example, to switch to using UCX 1.16 at runtime, run the following.
+For example, to switch to using UCX 1.19 at runtime, run the following.
 
 ::
 
+    # CUDA 13
+    pip install 'libucx-cu13>=1.19.0,<1.20'
+
     # CUDA 12
```
pentschev, line 249:
```suggestion
    # CUDA 12
    pip install 'libucx-cu12>=1.19.0,<1.20'

    # CUDA 13
```

docs/source/install.rst
```diff
@@ -225,19 +232,22 @@ To build UCX-Py using those UCX packages (to avoid needing to build UCX from sou
...
-    pip install 'libucx-cu12>=1.16.0,<1.17'
+    pip install 'libucx-cu12>=1.19.0,<1.20'
```
pentschev, line 250:
```suggestion
    pip install 'libucx-cu13>=1.19.0,<1.20'
```

VERSION
```diff
@@ -1 +1 @@
-0.46.0
+0.45.00a30
```
pentschev: Is this intentional?

docs/source/install.rst
jameslamb: @bdice asked me in https://github.com/rapidsai/kvikio/pull/803#discussion_r2287989908 to order these types of things with newer CUDA first.

I'm indifferent about it, will let him comment here and will do whatever you two agree on.

VERSION
jameslamb: This was a mistake from local testing, will revert it, thanks for catching it!

docs/source/install.rst
pentschev: I'd prefer we keep it consistent with the text, it looks very strange to me having 13 come before 12, this is how it always looked in the past (11 then 12).

VERSION
jameslamb: reverted in https://github.com/rapidsai/ucx-py/pull/1162/commits/32806dbcbbb0ff5c55b8f59d1472f1d4e8645e83

docs/source/install.rst
bdice: We want to shift the default (first command offered) to be the newest. The newest CUDA version will be supported for longer and we want to encourage users to adopt new versions.

bdice: We can rearrange the text as needed but should prefer 13 over 12 in ordering and in any situations where we only give one example.

pentschev: I'd prefer if we then rephrase (some of) the text, if we want to encourage that we could start referring to CUDA 12 as "legacy" or something like that to provide some strong encouragement.

pentschev (approved):
LGTM, thanks @jameslamb !

VERSION
```diff
@@ -1 +1 @@
-0.46.0
+0.46.0
```
jameslamb:
```suggestion
0.46.0

```

ugh, we lost the newline in my attempt to revert something from local testing. This fixes that, so this file should completely drop out of the diff.

Since we're planning to archive this repo soon (#1160 ) I won't add a PR with the `pre-commit` hook to fix trailing newlines.

docs/source/install.rst
```diff
@@ -225,19 +232,22 @@ To build UCX-Py using those UCX packages (to avoid needing to build UCX from sou
     # or for develop build
     pip install -v -e .
 
-This will automatically handle installing appropriate, compatible ``libucx-cu12`` packages for build-time and runtime use.
-When you run UCX-Py code installed this way, it will load UCX libraries from the installed ``libucx-cu12`` package.
+This will automatically handle installing appropriate, compatible ``libucx-cu{12,13}`` packages for build-time and runtime use.
+When you run UCX-Py code installed this way, it will load UCX libraries from the installed ``libucx-{cu12,13}`` package.
```
bdice:
```suggestion
When you run UCX-Py code installed this way, it will load UCX libraries from the installed ``libucx-cu{12,13}`` package.
```

bdice (approved):
One tiny suggestion, otherwise LGTM

docs/source/install.rst
jameslamb: Oh totally agree! This was my intent, but clearly not what my fingers typed 🙃 

Thank you

docs/source/install.rst
jameslamb: I don't think we should hold up this PR over this phrasing in documentation. Since you've both approved the PR (so approved all the other changes), I'm planning to merge this as soon as CI finishes, to keep moving forward with supporting CUDA 13 in RAPIDS.

I'd be happy to review a follow-up PR changing the text around these install instructions.

pentschev: Yes, please move forward @jameslamb .

jameslamb: Thank you!

PR #1164, opened 2025-08-29
restore cudf dependency

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

#1162 temporarily removed the `cudf` test-time dependency here, because there weren't yet CUDA 13 `cudf` packages.

Those now exist (https://github.com/rapidsai/cudf/pull/19768), so this restores that dependency.


pentschev (approved):
LGTM, thanks James.

vyasr (approved):

PR #1165, opened 2025-09-02
use 'nvidia-ml-py' package for 'pynvml' module

jameslamb:
Contributes to https://github.com/rapidsai/build-infra/issues/293

The `pynvml` *package* (https://github.com/gpuopenanalytics/pynvml) provides 2 things:

* the `pynvml` **Python module**, transitively via a dependency on the `nvidia-ml-py` package
* the `pynvml_utils` Python module

This project's doesn't need the `pynvml_utils` Python module, so this PR proposes dropping the dependency on the `pynvml` package in favor of `nvidia-ml-py`.

## Notes for Reviewers

I know this project's about to be archived (#1160), but that is still in progress and not guaranteed to be done in time for the 25.10 release.

Proposing making this change to help completely remove the `pynvml` package from RAPIDS environments, so we can test that it's truly safe to do that.


vyasr (approved):

pentschev (approved):
LGTM, thanks James!

PR #1167, opened 2025-09-15
Add deprecation warning to docs

pentschev:
Add a deprecation warning to docs to advise users to migrate to UCXX.

PR #1169, opened 2025-09-17
Make final discontinuation changes

pentschev:
This is expected to be the last commit to UCX-Py. It updates the README and docs warnings to inform users that UCX-Py has been permanently discontinued. Along with it, this change removes CUDA 13 mentions from docs, since there will be no CUDA 13 packages released for UCX-Py, however, the functional changes to allow building with CUDA 13 are kept should those be needed for any reason in the near-term.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 > [!CAUTION]
-> UCX-Py has been discontinued, version 0.45 (RAPIDS 25.08) was its last release and it will receive no further updates. Projects depending on UCX-Py are advised to migrate to [UCXX](https://github.com/rapidsai/ucxx) immediately. [UCXX documentation](https://docs.rapids.ai/api/ucxx/nightly/).
+> UCX-Py has been discontinued, version 0.45 (RAPIDS 25.08) was its last release and it will receive no further updates, see the [RAPIDS Support Notice 50](https://docs.rapids.ai/notices/rsn0050/) for details. Projects depending on UCX-Py are advised to migrate to [UCXX](https://github.com/rapidsai/ucxx) immediately. [UCXX documentation](https://docs.rapids.ai/api/ucxx/nightly/).
 
 [![https://ucx-py.readthedocs.io/en/latest/](https://readthedocs.org/projects/ucx-py/badge/ "ReadTheDocs")]( https://ucx-py.readthedocs.io/en/latest/ )
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-> [!WARNING]
-> UCX-Py is deprecated and will be phased out soon, version 0.45 (RAPIDS 25.08) is expected to be the last release. Projects that rely on UCX-Py are advised to migrate to [UCXX](https://github.com/rapidsai/ucxx) immediately.
+> [!CAUTION]
+> UCX-Py has been discontinued, version 0.45 (RAPIDS 25.08) was its last release and it will receive no further updates. Projects depending on UCX-Py are advised to migrate to [UCXX](https://github.com/rapidsai/ucxx) immediately. [UCXX documentation](https://docs.rapids.ai/api/ucxx/nightly/).
 
 [![https://ucx-py.readthedocs.io/en/latest/](https://readthedocs.org/projects/ucx-py/badge/ "ReadTheDocs")]( https://ucx-py.readthedocs.io/en/latest/ )
 
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -40,10 +40,6 @@ Change `cuda-version` to pin to a different CUDA minor version if you'd like.
 
 ::
 
-    # CUDA 13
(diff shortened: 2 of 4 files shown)


docs/source/install.rst
```diff
@@ -243,9 +235,6 @@ For example, to switch to using UCX 1.19 at runtime, run the following.
 
 ::
 
-    # CUDA 13
-    pip install 'libucx-cu13>=1.19.0,<1.20'
-
     # CUDA 12
     pip install 'libucx-cu12>=1.19.0,<1.20'
```
jameslamb: Since we're not planning to release `ucx-py-cu12==0.46`, anyone running these instructions will get `ucx-py-cu12==0.45`.

`ucx-py-cu12==0.45` packages are pinned to not allow UCX 1.19:

https://github.com/rapidsai/ucx-py/blob/d0a58d340f07d109a6339a75b0dc50e29213abfd/dependencies.yaml#L197

So I think something needs to be done here.

If the plan is to merge this into `branch-0.46` and then `branch-0.46` into `main` and then archive the repo, and for these docs to be the ones users find at https://ucx-py.readthedocs.io/en/latest/, then I think we need to also revert any docs changes about UCX 1.19, back to the range support by 0.45.

I think you may just want to do this on your PR branch:

```shell
git checkout branch-0.45 -- docs/
```

And then just manually re-add the warning added in https://github.com/rapidsai/ucx-py/pull/1167

README.md
```diff
@@ -1,5 +1,5 @@
-> [!WARNING]
-> UCX-Py is deprecated and will be phased out soon, version 0.45 (RAPIDS 25.08) is expected to be the last release. Projects that rely on UCX-Py are advised to migrate to [UCXX](https://github.com/rapidsai/ucxx) immediately.
+> [!CAUTION]
+> UCX-Py has been discontinued, version 0.45 (RAPIDS 25.08) was its last release and it will receive no further updates. Projects depending on UCX-Py are advised to migrate to [UCXX](https://github.com/rapidsai/ucxx) immediately. [UCXX documentation](https://docs.rapids.ai/api/ucxx/nightly/).
```
jameslamb:
```suggestion
> UCX-Py has been discontinued, version 0.45 (RAPIDS 25.08) was its last release and it will receive no further updates. Projects depending on UCX-Py are advised to migrate to [UCXX](https://github.com/rapidsai/ucxx) immediately. [UCXX documentation](https://docs.rapids.ai/api/ucxx/nightly/). For more details, see https://docs.rapids.ai/notices/rsn0050/.
```

I think we should add a link to the RSN. That's helpful in teaching RAPIDS users that that's a section of the docs they should watch for breaking changes like this.

I've been doing that elsewhere, for that reason:

* https://github.com/rapidsai/pynvjitlink/pull/155
* https://github.com/rapidsai/cuspatial/pull/1574

jameslamb (changes requested):
If I've understood the plan correctly, I think we also need to update the version here:

https://github.com/rapidsai/ucx-py/blob/51f87cd97adc9584b8441402ea797e029bfc6496/docs/source/conf.py#L22-L26

To a hard-coded:

```python
version = "0.45.0"
```

Otherwise, when we merge this branch to `main` and build docs from there, the docs will say `0.46.0` for the version wherever Sphinx puts it, like in the tab titles:

image

pentschev:
> If I've understood the plan correctly, I think we also need to update the version here:
> 
> https://github.com/rapidsai/ucx-py/blob/51f87cd97adc9584b8441402ea797e029bfc6496/docs/source/conf.py#L22-L26
> 
> To a hard-coded:
> 
> ```python
> version = "0.45.0"
> ```

You're right, sorry I forgot about this!

docs/source/install.rst
pentschev: Now I understand your point, previously I thought about this but thought it wasn't necessary because we can support 1.19.0, but I didn't think about pins in the 0.45 packages.

README.md
pentschev: Done with a slightly reorder in phrasing, see https://github.com/rapidsai/ucx-py/pull/1169/commits/64f73f8469bb233134cd20f0265656c07b8c6812, let me know if you prefer your phrasing/ordering instead.

pentschev:
Thanks @jameslamb , I think I've addressed everything now, please have another look when you have a chance.

README.md
jameslamb: Looks totally fine to me, all I cared about is getting the RSN link in there somewhere. Thanks!

jameslamb (approved):
Changes are looking good to me, thank you!

pentschev:
Just realized one more thing that we want to change, which is the minimum UCX version, which is 1.15.0 in 0.45. Applied that change now in https://github.com/rapidsai/ucx-py/pull/1169/commits/4e88e63ce4651639c79110df892195590e2499c2 .

madsbk (approved):

